PR #25, opened 2018-04-30
Start Figwheel REPL

pedrorgirardi:
Changed default command to start Figwheel REPL.


pedrorgirardi:
Yeah, I suspected that I was missing something. 😅 I will check your commit later to learn 😀

PR #27, opened 2018-05-01
Add new configuration to set if Calva should show greetings

### LINKED ISSUE #14: Do we need to show the output panel on initialization?
What concerns do you have about changing this behavior to not showing the output panel? What I like about Calva is that it 'just works'. Once the programmer has Calva installed one doesn't need to think much about it.

pedrorgirardi:
What about giving the user the option whether to show or not the greetings? This would fix #14 .

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,6 +98,11 @@
                     "type": "string",
                     "default": ".",
                     "description": "The root directory of your Clojure project, relative to the Workspace root"
+                },
+                "calva.showGreetings": {
+                    "type": "boolean",
+                    "default": true,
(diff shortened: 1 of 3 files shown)


PEZ:
I am a bit hesitant to make this a configuration option.

What about this idea: We have a small status bar indicator for Calva that tells if Calva is activated. The tooltip for it says to click it to get any messages we would like them to read. When clicked it opens up an information pane with our messages.

pedrorgirardi:
Sounds good to me. 👍

PR #37, opened 2018-05-12
Fix invalid imports and exports

### LINKED ISSUE #36: Linting failing
Something has broken linting. Joker recently updated to 0.9.2 so it could be the culprit, but the update didn't include anything major.

Trace:
`TypeError: Cannot read property 'getDocument' of undefined
	at Object.lintDocument$$module$calva$repl$middleware$lint [as lintDocument]`

WhoNeedszZz:
**Disclaimer**: While this should normally be separated out into smaller commits for each corresponding import/export pair, due to the nature of the issue it was impossible to test each individually as each invalid import/export pair affected multiple areas of the extension.

Multiple issues were happening here. Some modules were exporting a default object of function names rather than the intended exporting of multiple functions. Some modules were exporting a default object of a single function rather than the intended exporting of a default function. The respective imports for these cases were then also invalid and have now been adjusted accordingly.

Fixes #36 and other behind the scenes functionality that was silently failing.

### DIFF
--- a/src/main/calva/connector.js
+++ b/src/main/calva/connector.js
@@ -1,42 +1,42 @@
 import vscode from 'vscode';
 import _ from 'lodash';
 import fs from 'fs';
-import state from './state';
-import repl from './repl/client';
+import { cursor, deref, reset } from './state';
+import createReplClient from './repl/client';
 import message from 'goog:calva.repl.message';
-import * as util from './utilities';
-import shadow from './shadow';
-import status from './status';
-import terminal from './terminal';
+import { getProjectDir, getSession } from './utilities';
+import { isShadowCljs, shadowBuilds, shadowNReplPortFile } from './shadow';
+import updateStatus from './status';
+import { createREPLTerminal } from './terminal';
 //const evaluate = require('./repl/middleware/evaluate');
 
 function nreplPortFile() {
-    if (fs.existsSync(shadow.shadowNReplPortFile())) {
-        return shadow.shadowNReplPortFile();
+    if (fs.existsSync(shadowNReplPortFile())) {
+        return shadowNReplPortFile();
     }
     else {
-        return util.getProjectDir() + '/.nrepl-port'
+        return getProjectDir() + '/.nrepl-port'
     }
 }
 
 function disconnect(options = null, callback = () => { }) {
-    let chan = state.deref().get('outputChannel'),
+    let chan = deref().get('outputChannel'),
         connections = [];
 
 
     ['clj', 'cljs'].forEach(sessionType => {
-        if (util.getSession(sessionType)) {
-            connections.push([sessionType, util.getSession(sessionType)]);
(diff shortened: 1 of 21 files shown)


PEZ:
Hello, many thanks for this PR! I'd like to pull only the refactor lint commit, and leave the other one for now. Not sure how I would make such a cherry pick. Can you recook the PR or maybe make a new one?

The reasons to not pull in the import fixes is that I have about the same fixes in another branch. One that doesn't change the style of using code from other modules. We have been going in the direction of changing the code base from having code like `getDocument()` to `util.getDocument()`, something that we will continue to do in the `cljs` porting.

WhoNeedszZz:
The refactoring was purely for readability. The fixing of the imports and exports is what actually fixed the problems.  The direction you are talking about is an anti-pattern. To do something like `util.getDocument()` you would have to `import * as util from './util'`, which would import everything in that scope. This is the reason imports and exports are handled the way they are and what is industry standard best practice.

PEZ:
I like readability, and that's why I also prefer `(util/getDocument)` over just `(getDocument)`. It is painful to see those `import *`things, but at least in Clojure(Script) it is Kosher to `require` using `:as`, and that is where we are heading.

WhoNeedszZz:
I don't know what you think this is. You asked for help contributing and here is a professional Software Engineer with over 10 years of experience trying to help you not just with the code itself, but with programming best practices in general. I've explained multiple times and quite clearly that what you are referring to is an anti-pattern both in the JavaScript world and the Clojure/ClojureScript world. It is not a stylistic choice at all, but has specific performance issues because of importing code that you aren't using. Using `:as` in a `require` expression is not kosher for the exact same reason `import * as foo from './bar'`; it includes all code from that namespace. The proper and recommended way is to use `:refer` to only refer the functions and macros you need. Something very important in programming that many developers overlook is that readability should never come at the cost of an anti-pattern.

I took my free time and the better part of a day to go through your code and fix these problems so for you to reject the PR and fight me over this over a misguided personal preference is quite frustrating. I didn't realize that this a pet project for you. I don't think you realize how many people were excited to see a functioning Clojure/ClojureScript extension for VSCode that was working before you decided to make these poor decisions, which have left the extension in a broken state. I will be going back to Atom with proto-repl as it works as intended and the developer doesn't release broken builds to the public. Good luck...

PEZ:
I'm sorry you give up on this extension over this. The changes you are recommending for fixing the issue at hand are implemented and the linting now works again. I wouldn't call that ”rejecting the PR”. I appreciate that you took the time to look over things.

We are of course trying not to leave the extension in a broken state. While it does and will happen that we break things, we will keep being alert and fix those things as fast as we can.

Regarding import strategy. The use of `:as` is recommended in this [Clojure Style Guide](https://github.com/bbatsov/clojure-style-guide#prefer-require-over-use), that has so far served me and many other Clojurians well.

PR #52, opened 2018-06-26
feature: command to evaluate the current forms parent

julienvincent:
Given the code block with cursor at position of `|`

```clojure
(defn x []
  (a (b (+| 1 2))))
```

Calling `calva.evaluateSelection` will give me `=> 3`. This is nice but I often want to be able to evaluate the root form `(defn x)` without having to precisely position my cursor.

This PR adds an additional command `calva.evaluateSelectionParent` which does this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -152,8 +152,8 @@
                 "category": "Calva"
             },
             {
-                "command": "calva.evaluateSelectionParent",
-                "title": "Evaluate the parent form of the current cursor position",
+                "command": "calva.evaluateCurrentTopLevelForm",
+                "title": "Evaluate the current top level form at selection or cursor",
                 "category": "Calva"
             },
             {
@@ -246,6 +246,10 @@
             {
                 "command": "calva.evaluateSelection",
                 "key": "ctrl+alt+v e"
+            },
+            {
+                "command": "calva.evaluateCurrentTopLevelForm",
+                "key": "ctrl+alt+v shift+e"
             },
             {
                 "command": "calva.evaluateSelectionPrettyPrint",
--- a/package.json
+++ b/package.json
@@ -152,6 +152,11 @@
                 "category": "Calva"
             },
             {
+                "command": "calva.evaluateSelectionParent",
+                "title": "Evaluate the parent form of the current cursor position",
+                "category": "Calva"
+            },
+            {
                 "command": "calva.evaluateSelectionPrettyPrint",
                 "title": "Evaluate selection or current form, and pretty print",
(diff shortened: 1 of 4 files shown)


PEZ:
Thanks! A question: the range you want to evaluate here, is that the same as Paredit will select when doing *Range For Defn*? If so we should probably name this command to match, like *Evaluate Range For Defn*. Also if so, I think the code for selecting it can be simplified using `paredit.navigator.rangeForDefun(ast, idx)`.

julienvincent:
Regarding the naming, I don't think `Evaluate Range For Defn` is very clear to the reader as to what the command does. Especially as it isn't only evaluating for `defn` macros but also for arbitrary nested forms such as

```clojure
(+ 1 (+ 1 ( + 1)))
```

PEZ:
Tested this now. Works great! This will cut hours per week of placing the cursor for me.

About the naming. My suggestion was to make it correspond to the Paredit command. Which does not have a very good name… But *parent form* is also confusing, at least to me. What is actually evaluated is the current top level form, right? So we could name it **Evaluate current top level form**, maybe?

I will rename the Paredit selection command when we have decided what this one should be called.

julienvincent:
Agreed, parent isn't great either. I like Evaluate current top level form - thats a lot more clear.

PEZ:
I think it should have a default key binding too. Maybe `ctrl+alt+v shift+e` makes sense?

PEZ:
I choose to use `ctrl+alt+w space` as default for making the top-level form selection in Paredit so we could use `ctrl+alt+v space` to evaluate the same form.

julienvincent:
I added a command for `ctrl+alt+v shift+e` and left it up to the user to rebind

PEZ:
Hmmm, i pulled your changes earlier, made some changes and just merged into master and pushed without noticing the latest commit, @julienvincent. I'm not sure what to do now. But I'll figure it out. :smile:

julienvincent:
Oops :) I'll resolve the conflicts a little later 👍

PEZ:
No, I can't figure it out. But we made about the same changes. I choose `ctrl+alt+v space` for shortcut and also added a command so that the top level form can be sent to the REPL terminal and evaluated there (shortcut `ctrl+alt+v alt+space`).

I'm closing this PR now, since it is pulled. But be aware that your fork has diverged because of my slip-up.

PEZ:
I'll update the README some and release this nice feature later today.

julienvincent:
Alright no problem 👍

PR #61, opened 2018-08-15
prevent terminal from taking focus, cleaned up unnecessary params

LukeDefeo:
When you pass keep focus true to terminal.show() it makes the editor keep the focus. There were many code paths where false was being passed

I removed the params from the outer methods as i couldn't think of a reason that we would want the terminal to take focus and it simplifies things a bit. If theres something im not considering (highly likely) let me know and ill re-add them


PEZ:
This is the right call. Thanks!

PR #62, opened 2018-08-16
added top level pretty print command and shortcut

LukeDefeo:
this feature was missing seemed like a no brainer


PEZ:
No brainer, indeed. Thanks!

PR #65, opened 2018-08-20
added option to sync repl namespace to current file

LukeDefeo:
added a simple implementation that mimics intellijs behavior, i think its convenient and beginner friendly so defaulted to on. let me know what you think

### DIFF
--- a/calva/extension.ts
+++ b/calva/extension.ts
@@ -118,6 +118,9 @@ function activate(context) {
     }));
     context.subscriptions.push(vscode.window.onDidChangeActiveTextEditor((editor) => {
         status.update();
+        if (state.config().syncReplNamespaceToCurrentFile) {
+            terminal.setREPLNamespace()
+        }
     }));
     context.subscriptions.push(new vscode.Disposable(() => {
         connector.disconnect();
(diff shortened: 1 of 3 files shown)


PEZ:
Thanks! I wonder, does it make sense to do something similar with the REPL session backing the file?

LukeDefeo:
I’m not sure what you mean ?

PEZ:
I might be confusing also myself, but anyway.

When Calva connects to the Clojure nREPL session it creates two REPL sessions. One is used ”internally” to back the files being edited (peek at definitions, evaluate code, etcetera). The other is created by first creating a Terminal session and from there issuing a CLI command to connect to a REPL and showing a prompt.

So, when switching files automatically loads the namespace in the terminal REPL, maybe it should also do so for the ”internal” REPL? I think Calva once had this behaviour, but that it confused some experienced users so I removed it. But if we put it behind a setting, like you have done here, maybe that makes most sense? I guess this depends a bit how Calva is used bu a particular user. For instance, I very seldom use the Terminal REPLs myself, I use `comment` forms in the files and evaluate stuff from there.

PEZ:
Sorry, I didn't want to stall this PR. Going to pull it now. Hopefully I can release it this weekend.

But, anyway. Did anything of what I wrote make sense? 😀

LukeDefeo:
I’ve not had the chance to go through the repl parts in details. The n repl
communication area is definitely an area I’m a bit confused by, let’s talk
 bout it on slack over the weekend

PR #67, opened 2018-08-30
Bug in selectCurrentForm.  Type mismatch in editor.selection assignment.

martinbertolino:
I noticed a problem with selectCurrentForm.  It always failed on my setup, both Windows 10 and macOS, code version 1.26.1 and Calva version 1.3.36.

I tracked the problem to a type mismatch in select.ts:selectCurrentForm when assigning to editor.selection.  getFormSelection returns a vscode.Range type but editor.selection is vscode.Selection type.

I fixed by type annotating getFormSelection and then using the returned vscode.Range to construct a vscode.Selection and then assign to editor.selection.

I posted the differences in the Slack channel as well.

Let me know what you think.

Thank you,
Martin

### DIFF
--- a/calva/repl/middleware/select.ts
+++ b/calva/repl/middleware/select.ts
@@ -33,7 +33,7 @@ function getFormSelection(doc, pos, topLevel) : vscode.Selection {
         ast = paredit.parse(allText),
         idx = doc.offsetAt(pos),
         range = topLevel ? paredit.navigator.rangeForDefun(ast, idx) : paredit.navigator.sexpRange(ast, idx),
-        vsSelection = range? new vscode.Selection(doc.positionAt(range[0]), doc.positionAt(range[1])) : new vscode.Selection(pos, pos);
+        vsSelection = range ? new vscode.Selection(doc.positionAt(range[0]), doc.positionAt(range[1])) : new vscode.Selection(pos, pos);
 
     return vsSelection;
 }
--- a/calva/repl/middleware/select.ts
+++ b/calva/repl/middleware/select.ts
@@ -28,14 +28,14 @@ function adjustRangeIgnoringComment(doc, range) {
 }
 
 
-function getFormSelection(doc, pos, topLevel) : vscode.Range {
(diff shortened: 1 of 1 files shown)


PEZ:
Hello! Many thanks for this help!

It was a while since I visited this part of the Calva code, so before I pull this I'd like to ask a thing. Is there a reason why `getFormSelection()` should return a Range? Not saying that there isn't, just wondering and this if maybe we should just make `getFormSelection()` return a selection as its name implies. (Thinking this might be a reason for the bug to enter the code.)

Let me know. I'll pull this if you tell me there are reasons to return a Range there. Otherwise I'd rather go for fixing the return type.

martinbertolino:
@PEZ , you are absolutely right.  I wasn't sure how much I should change the code vs. just fix the bug.  I agree that getFormSelection should return vscode.Selection and then avoid having to convert types later.  I will take care of this soon and create a new pull request.

martinbertolino:
I was thinking about this some more and I think I need to make sure that changing the return type of getFormSelection does not break other parts of the code.  A quick search shows that select.getFormSelection is used in some other parts of the code as well.  I need to ensure I execute and test those other sections as well.

martinbertolino:
In addition to testing select.selectCurrentForm, I also did some testing of the following commands:

* terminal.evalCurrentFormInREPLTerminal, this needs to be executed with no form selected in order to trigger the call of select.getFormSelection.  This seems to be working as expected.
* evaluate.evaluateSelection, again this needs to be executed with no form selected in order to trigger the call of select.getFormSelection.  This seems to be working as expected as well.

PEZ:
Good call. Let me know when you think it is ready for pulling. And of course let me know if I can help with something, testing or whatever.

martinbertolino:
I think it is ready to be pulled.  I have done some more testing of the commands impacted by the change and could not find any problems.  Also, then I switched to working on an actual clojure project I was working and I did not noticed any effects or changes to my workflow.  Feel free to review and then pull if you are are OK with the changes.

PR #82, opened 2018-10-18
Fix spelling in package.json

Ibexian:
Just a small typo fix


PEZ:
Thanks! Sorry for letting it hang there. I got derailed.

PR #88, opened 2018-11-25
Add a test results panel to the UI

marcomorain:
Hi @PEZ  - I'm opening this PR to start a discussion on how to add the test results view into the UI.

If this the test results panel is something that you want to add to Calva:

- This needs to be cleaned up a lot before it's ready to merge.
- We need to decide what the action should be when a user clicks on a test in the panel - should we jump to definition? Re-run the test?

### DIFF
--- a/calva/repl/middleware/testRunner.ts
+++ b/calva/repl/middleware/testRunner.ts
@@ -4,10 +4,9 @@ import * as state from '../../state';
 import repl from '../client';
 import evaluate from './evaluate';
 import * as util from '../../utilities';
+import {Outcome, testDataProvider} from '../../providers/testData'
 
 import * as calvaLib from '../../../lib/calva';
-
-
 
 let diagnosticCollection = vscode.languages.createDiagnosticCollection('calva');
 
@@ -23,7 +22,10 @@ function markTestResults(responsesArray, log = true) {
             if (results !== null) {
                 _.each(results, (tests, ns) => {
                     _.each(tests, (asserts, test) => {
+                        testDataProvider.onTestResult(asserts);
+
                         _.each(asserts, a => {
+
                             if (a.type == "error") {
                                 if (log) {
                                     chan.appendLine("ERROR in: " + ns + ": " + a.file + ", line " + a.line +
--- a/calva/extension.ts
+++ b/calva/extension.ts
@@ -7,6 +7,7 @@ import terminal from './terminal';
 import CalvaCompletionItemProvider from './providers/completion';
 import TextDocumentContentProvider from './providers/content';
 import HoverProvider from './providers/hover';
+import {testDataProvider, goToFile} from './providers/testData'
 import DefinitionProvider from './providers/definition';
 import EvaluateMiddleWare from './repl/middleware/evaluate';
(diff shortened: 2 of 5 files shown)


calva/repl/middleware/testRunner.ts
```diff
@@ -195,12 +251,120 @@ function rerunTestsCommand() {
...
+export interface TestListener {
+    onTestResult(ns: string, varName: string, success: boolean): void;
+}
+
+// The test result tree-view has 2 type of node:
+// Namespace nodes and var nodes.
+// The (root) contains NamespaceNodes, which have VarNodes as children.
+type TestNode = NamespaceNode | VarNode
+
+class ClojureTestDataProvider implements TreeDataProvider<TestNode>, TestListener {
```
marcomorain: I should pull this out into it's own file.

calva/repl/middleware/testRunner.ts
```diff
@@ -7,85 +7,135 @@ import * as util from '../../utilities';
...
-                    " errors: " + total_summary.error + ", failures: " + total_summary.fail) :
-                "No tests found. 😱") +
-                ", ns: " + total_summary.ns + ", vars: " + total_summary.var);
+interface Response {
+    session: String
+    id? : String
+    status? : String[]
+}
+
+interface TestResponse extends Response {
```
marcomorain: I defined some types to try to make it easier to deal with nREPL messages:

https://clojurians.slack.com/archives/C17JYSA3H/p1542971155229800

PEZ:
Thanks for spearheading this, @marcomorain! I have never participated in a Github code review before so bare with me while I am learning.

About the code, really, the Calva code base is rather messy and you seem to be improving on it. In Calva Formatter I have also started to go the way where I try take advantage of TypeScript so I generally agree with the direction of that.

However, my longer term plan is to get much of the crunching over to ClojureScript. My general idea is to keep mostly the VS Code integration parts in TypeScript and build libraries in CLJS that provide the computations and integration towards nREPL and the Clojure world. So I'd like to ask you if you see any such ways to factor the code here? I think maybe most of the task at hand here is really VS Code integration so It might not be all that relevant for this view.

As for the view. I have tried it out some and gotten even more convinced that this would be a really good thing to add. I wonder about the problem markers that are now gone. Is that on purpose or a mistake? Also there is no indication in the view about failing tests, maybe that is just not implemented yet?

I think clicking on the items in the view should navigate to that part of the code. As I haven't look at the API for the views I don't know much about what you can do. Is it possible to add small ”rerun” buttons, for instance? Or is all such things added via the context menu? At least In the header it would be good with a ”rerun failing tests”. button.

As you have the test runner code and test ops more freshly in your head than I have, maybe you can also enlighten me if it will be possible to fold out individual asserts in the view?

marcomorain:
I spent a bit more time on this this week, I might have another play with it later on tonight. I've reverted all of the changes to the test runner.

stefan-toubia:
I know this PR is a bit old, but I also had a similar interest in a test UI.  I came across [Test Explorer UI](https://marketplace.visualstudio.com/items?itemName=hbenl.vscode-test-explorer), this could make for a good integration with Calva.

PR #92, opened 2018-12-12
Update namespace regex check to support comments

arjun27:
Hi team, thank you for this! I tried Cursive/Light Table/Emacs for Clojure dev, but settled on Calva + vscode.

I ran into an issue in using the `"calva.syncReplNamespaceToCurrentFile": true` configuration: the extension was not detecting the namespace in my `.clj` files correctly. This was because my namespace declarations were prefixed with some comments. For example:

This worked: if a file starts with any whitespace, followed by namespace declaration

```

(ns advent.2017.day20
  (:require [clojure.string :as str]))
```

This did not work: file starts with a comment, followed by namespace declaration.

```
; http://adventofcode.com/2018/day/20

(ns advent.2017.day20
  (:require [clojure.string :as str]))
```

This PR updates the namespace check to ignore comments and fix this issue. To summarize the change:

Earlier regex check: (0 or more whitespace) + (namespace check)
New regex check: (0 or more whitespace) + (0 or more comment lines) + (0 or more whitespace) + (existing namespace check)

### DIFF
--- a/calva/utilities.ts
+++ b/calva/utilities.ts
@@ -25,7 +25,7 @@ function getShadowCljsReplStartCode(build) {
 }
 
 function getNamespace(text) {
-    let match = text.match(/^[\s\t]*\((?:[\s\t\n]*(?:in-){0,1}ns)[\s\t\n]+'?([\w.\-\/]+)[\s\S]*\)[\s\S]*/);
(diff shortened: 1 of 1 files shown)


PEZ:
Thanks!

PEZ:
I have noticed that it also fails when the `ns` form has metadata. We should be having a more structural approach than relying on regexes, but this PR sure is an improvement in waiting for that.

adasomg:
Doesn't seem to work with >1 comment lines

PR #101, opened 2019-01-24
Fix evaluateCurrentFormInREPLTerminal to not use toplevel form

mseddon:
A bug introduced by https://github.com/BetterThanTomorrow/calva/commit/6b587bb8992b14c55b66e6e9344cdb33eedcf812#diff-8df9e3e120042e77ce02d69fc87fbdecL127 means that `evaluateCurrentFormInREPLTerminal` evaluates the toplevel form, not the current one.


PEZ:
Thanks!

PR #105, opened 2019-01-27
Make figwheel a bit more optional for people not using it

PEZ:
Instead of assuming that everybody wants Figwheel and showing
error messages if figwheel is not started, we inform them about how
to get Figwheel working with Calva.


mseddon:
LGTM!

PR #107, opened 2019-01-28
Stop always adding -test to the ns when running tests

PEZ:
This was breaking the case when running tests in a test namespace. And also made the test runner miss when the namespace contained tests w/o being named something ending in `-test`.


mseddon:
:D That's the one. LGTM!

Issue #132, opened 2019-02-21
Go to Definition and Peek Definition not working on Windows 10 when using WSL 


alzadude:
Hi,

I'm on Windows 10, and I have VSCode configured to use WSL instead of CMD, and so my nrepl gets started with WSL.

However it seems like this breaks `Go to Definition` (F12) and `Peek Definition` (Alt+F12).. 

For example, when I press F12 on a locally defined function in `xxx.clj`, I get a message similar to:

```
Unable to open 'xxx.clj': File not found (file:///mnt/d/Projects/my-project/src/xxx.clj).
```

Please could this be made to work for WSL? Recent Windows 10 builds have a `wslpath` command that can convert the paths, I was thinking this might help?

Thanks!

PEZ:
Hello,

I am not using Windows so this one quite tricky for me to address. But some time ago @andresrgz fixed another WSL issue... Do you know how this should be fixed, @andresrgz?

andresrgz:
> Hello,
> 
> I am not using Windows so this one quite tricky for me to address. But some time ago @andresrgz fixed another WSL issue... Do you know how this should be fixed, @andresrgz?

I do have an idea on how to fix this. In the [definition.ts](https://github.com/BetterThanTomorrow/calva/blob/99efabe109fe8ff0ce0e903a3d7cff1ed6624dd6/calva/providers/definition.ts#L15) file, the `info.file` variable contains a WSL path and should be converted into a valid Windows path before using it.

Since the repl is running within WSL, that's why the paths are being returned that way.

I may have a fix for this some time next week. 🙂 I found this [library](https://github.com/mojadev/wsl-path) that will be useful for fixing this problem.

PEZ:
Cool! Please have a go at it. However, if you can find a solution that does not require a binary installed, that would be extra good.

andresrgz:
@alzadude The PR to fix this has been merged. Maybe you could also help testing it before it gets published? 😄

[clojure4vscode-1.3.64-WSL.zip](https://github.com/BetterThanTomorrow/calva/files/2923281/clojure4vscode-1.3.64-WSL.zip)

Just rename the `.zip` extension to `.vsix`.

alzadude:
Hi @andresrgz,

I went to start testing, and I noticed that 1.3.64 must have already been "published" since I see that version number in my VSCode installation.

So I tried again, and `Go to Definition` and `Peek Definition` are now both working with WSL! 🎉 

Some notes:

[...]

PR #138, opened 2019-03-03
Enhancements for WSL support

andresrgz:
This PR includes a fix for issue #132 and also installs the [wsl-path](https://github.com/mojadev/wsl-path) library, this does depend on the `wslpath` binary, however that's already included with WSL.

Also, since this library was installed, it's now being used to convert the file path that is being sent to Joker. Here's the [line](https://github.com/andresrgz/calva/blob/d9d61578f4efa45a5d11bf5098a71e90c7e17435/calva/repl/middleware/lint.ts#L48) of that change.

Finally, the configuration parameter `useJokerOnWSL` was renamed simply to `useWSL`, that way the user specifies if the linter and nREPL are both running through WSL.

A better solution could've been achieved by instead modifying `cider-nrepl` so that it can return valid Windows paths within WSL. But this is the best I could come up with without modifying the repl dependency. 😄 

@PEZ You may also want to add the following to the wiki in order to describe how to use this extension with WSL:

# Instructions for WSL
If you want to use this extension with WSL, there are a few things that need to be configured first.

1. The `useWSL` option must be set to `true` and you must add a valid WSL path to the `jokerPath` option.
2. Make sure that you have Windows 10 version 1803 since that's the version that includes the `wslpath` tool which is used to convert between WSL and Windows paths.
3. If you're using Leiningen and if you want to view definitions of source files (e.g. `println`, `defn`, `def`) you will need to change your .m2 directory to a location directly accessible by Windows. For example, you could edit your `~/.lein/profiles.clj` like so:
```
{:user {:local-repo "/mnt/c/Users/<Your User>/.m2"}}
```


PEZ:
Many thanks! Looking at it now and will get back to you.

PEZ:
Good call to hook the Joker setting on this as well.

The code changes look right, and well applied, to me. I am not on Windows so can't really test this, but I have built it and see that the setting is there. Can you test this VSIX package for me @andresrgz ? If it does the right thing for you, I'll release it.

[clojure4vscode-1.3.64-WSL.zip](https://github.com/BetterThanTomorrow/calva/files/2923281/clojure4vscode-1.3.64-WSL.zip)

(You need to change the extension from `.zip` to `.vsix` in order to install it in VS Code.)

Issue #93, opened 2018-12-18
Calva does not show stack trace when loading a file fails


pikeview:
I have core.clj file
I imported another another ui.clj namespace in core.clj via (:require)

when I changed the ui and saved and loaded the file (ctrl, enter or ctrl+alt+v enter), calva did not show any errors in compilation..

I realized that there were errors only after restarting the repl

bpringe:
I've noticed this as well. I only know there must be errors in a file by looking at the Calva Says window. If I evaluate the file and see `nil` or `no results` in this window, something is probably wrong. It would be nice to see the error to know what went wrong.

PR #143, opened 2019-03-19
Show stdout and stderr output when loading a file

alek-sys:
See  #93


PEZ:
Thanks!

PEZ:
This is now released in v1.3.65. Sorry for it taking so long!

alek-sys:
@PEZ no worries, thanks for fixing new VS Code so quickly, found it yesterday but then noticed it is a reported already. Great job!

PR #150, opened 2019-04-07
Find parent .nrepl-port file

mrkam2:
I work in an environment, where I have multiple clojure projects opened in the same workspace. This PR allows the extension to figure out correct nrepl port based on the active file.

### DIFF
--- a/calva/connector.ts
+++ b/calva/connector.ts
@@ -1,6 +1,7 @@
 import * as vscode from 'vscode';
 import * as _ from 'lodash';
 import * as fs from 'fs';
+import * as path from 'path';
 import * as state from './state';
 import * as util from './utilities';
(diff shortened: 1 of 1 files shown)


PEZ:
Excellent. Thanks!

It calls for also solving this for shadow-cljs projects. Additionally, in the next ”major” release of Calva, we will support jack-in, which will need a similar approach. You have showed the path, so I can take it from here, but if you do have some time, I'd appreciate your help. Please join the #calva-dev channel on the Clojurian Slack to sync the efforts, if you decide you do have the time.

mrkam2:
Thank you! I need an invite to slack.

mrkam2:
Nvm, I figured it.

PR #155, opened 2019-04-08
Show full error line in output

mrkam2:
This PR improves "evaluate current file" productivity in two ways:
1. It shows the output.
2. It expands a short location of error into a long one that VSCode can convert to a link, so it is now:

```
Evaluating file: /Users/user/src/project/file.clj
Syntax error compiling at (/Users/user/src/project/file.clj:49:1).
```


PEZ:
:heart:

PR #163, opened 2019-04-13
Implement basic symbol provider

mrkam2:
This PR implements a basic and not very efficient symbol provider.

### DIFF
--- a/calva/extension.ts
+++ b/calva/extension.ts
@@ -19,6 +19,7 @@ import evaluate from "./repl/middleware/evaluate"
 import { nClient } from "./connector"
 
 import { readFileSync } from 'fs';
+import { DocumentSymbolProvider } from './providers/symbol';
 const greetings = require('@cospaia/calva-lib/lib/calva.greet');
 
 function onDidSave(document) {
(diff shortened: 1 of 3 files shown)


mrkam2:
@PEZ WDYT?

PEZ:
I like it! When it works, it improves navigation in the file a lot. Nice that it works in ClojureScript as well! You are probably aware of these but here's what I found:
* It finds some crazy symbols in the `user` namespace.
  * Navigating to some of those causes exceptions.
* Some seem to ”be” in the user namespace, which causes the ”real” symbols in my `user.clj` to show up as well (together with those crazy ones mentioned above). Since the symbols UI seems to be made only for in-file navigation we need to only list symbols we find in the same file.
* Evaluating some `def*` form in the file makes it appear in the symbols listing, but you can't navigate to it.
  * Evaluating a form several times makes it appear several times, collapsed into one but vscode, but you can't navigate to it any longer.
* There are lots of Errors in the Development Console after using the feature a few times.

I'd love it if you took this feature to a shippable state. It surprised me how quickly I started to get used to it. But please do so on the `dev` branch as I don't have bandwidth for keeping track on development on master. I have updated the [How to Contribute wiki page](https://github.com/BetterThanTomorrow/calva/wiki/How-to-Contribute) to instruct on how to get that to build. And you know where to reach me if you need assistance. :smile:

PEZ:
Yes, I also noticed that the line numbers get out of whack when lines are added and removed in the file. Reloading the file helps. Not sure what can be done about this, though.

PEZ:
I have now pushed a branch with current `dev` plus this PR: `mrkam2/symbols`. So that you don't need to resolve the (few) conflicts that I already have solved.

PR #164, opened 2019-04-14
Implement basic symbol provider v.2

mrkam2:
This PR implements second version of basic symbol provider. It is now using `apropos` to retrieve name type. It also avoids adding symbols with incorrect file information.


mrkam2:
@PEZ WDYT? If you reproduce some of the issues mentioned before, could you please provide more specific logs and steps to reproduce?

PEZ:
This time around I do not get any symbols at all. Could you have something you have not pushed yet left on your machine?

calva/providers/symbol.ts
```diff
@@ -12,15 +19,17 @@ export class DocumentSymbolProvider implements vscode.DocumentSymbolProvider {
     if (this.state.deref().get('connected')) {
       let client = util.getSession(util.getFileType(document));
       const ns = util.getNamespace(document.getText());
-      const vars = await client.nsVars(ns);
-      const items = await Promise.all(vars["ns-vars"].map((v: any) => {
-        return client.info(ns, v);
-      }));
-      return items.map((info: any) => new vscode.SymbolInformation(
+      const aproposResponse = await client.apropos({
```
mrkam2: @PEZ I switched to use [apropos](https://cljdoc.org/d/cider/orchard/0.4.0/api/orchard.apropos) and [its middleware](https://github.com/clojure-emacs/cider-nrepl/commit/cf9f3a0c98e239a6fb9ca2740ad665d129ecc17b).
Mind checking the value of `aproposResponse` trying this branch your system?

PEZ: There are no `apropos-matches` in the response when I run this. It is a `done` response... Let's chat on Slack, maybe we can use Hangout or something to figure it out together.

PEZ: Actually, I now see that I have an `appropos-matches`, an empty array, in the response. I'm sorry for misleading you with my previous comment.

PEZ:
OK. So now I've fiddled around with the dependencies and gone full circle with them, ending up with the same that I started with, and now the symbols appear for some reason.

Findings after a quick test drive:
* The problem with strange symbols appearing is gone.
* There are now no symbols for the `user` namespace in my test project. (Not a stopper, IMO, but that's how it is anyway.)
* The Outline pane sometimes populates with the symbols now. But sometimes not. I guess that needs some attention. Would be a super extra benefit to get that working consistently.

Are you planning on presenting functions and regular `def`s differently? I don't know what you get from `apropos` that could help with that, but if there is such info, it would make the symbols least easier to parse, I think.

Let me know if you need assistance merging in the updates from `dev`.

mrkam2:
> Are you planning on presenting functions and regular defs differently?

I though I already did that in "Go To Symbol..." list. 

> The Outline pane sometimes populates with the symbols now. But sometimes not.

I'll look into Outline separately. My goal was to fix "Go To Symbol..." list, this seems like a byproduct.

> There are now no symbols for the user namespace in my test project. 

It hides any symbols that have no file information. Maybe that's the case with user ones. What would be a desired workflow for those?

> Let me know if you need assistance merging in the updates from dev.

I probably don't need. Why can't this be merged into master first? What's the plan with dev?

PEZ:
When I tested `def` and `defn` symbols looked the same in the list.

About the Outline pane, I agree it is a side product, but right now this feature leaks out there, so either we need to stop that, or fix it so that we populate it consistently. IMO.

The user namespace is special. I think that no, symbols works for now. But desired would be to see all the symbols that can be accessed from there.

About `master` and `dev`. Right now `master` is really just a maintenance branch for Calva v1 and all new development should go into Calva 2, which happens in `dev`. When Calva 2 is released, we'll be back with maintaining that in master. But right now they differ too much for me to have the time to maintain both with new features.

PEZ:
Bumping this!

mrkam2:
I did some investigation. Seems like Outline is directly based on SymbolProvider output that is a requirement for Go To Symbol feature. However, they seem to have different lifecycle: outline only queries symbols once when document is first opened. Go to symbol does the same query every time "Go to Symbol" dialog is shown. When REPL is not connected or namespace is not yet evaluated, `provideDocumentSymbols` function would return nothing. Therefore, any documents opened during this time would have empty Outline. So far I wasn't able to find a way to refresh the Outline later.

mrkam2:
Since I'm new to VSCode programming I might be missing something. But so far it seems that if we want to fully support the Outline view, we should be evaluating and loading symbols for each document upon opening. Does it sounds reasonable?

PEZ:
Yes, it sounds reasonable. But maybe there is also a way to explicitly tell vscode that we do not support the outline view yet? It'll be a matter about how much work there is with any path we choose.

mrkam2:
I'm not an expert in VSCode but from what I've learned so far it seems like it's not customizable that way.

PEZ:
I have dropped this ball a bit. What is the status as far as you know, @mrkam2 ?

mrkam2:
Hey Peter,

Me too. I was distracted with other stuff.

At the moment I think the best approach would be either to ignore Navigator
or force evaluation of any new file when it is opened. I don't see how we
can control Navigator otherwise. If you know somebody more familiar with
vscode internals/APIs who can help that would be great.

PR #174, opened 2019-04-30
utilities.getNamespace() needs to match .clj as well as [sc]

Gnurdle:


PEZ:
Super! This probably explains some errors I have been seeing with namespace mismatch.

PR #196, opened 2019-05-31
Added some color and fixed jumping brackets 

hjers:
When matched the brackets and parentheses where jumping, now the don't. Also added some styling, so it is clear when standing on matching brackets, parentheses etc.


PEZ:
Super duper awesome. Much slicker now!

PR #197, opened 2019-05-31
calmer caret in repl

hjers:
less blinking and skinnier caret in repl window


PEZ:
Thanks! Calmer and easier to see what is going to be edited.

PR #200, opened 2019-06-05
Update Readme

AlexVPopov:
Add an instruction for avoiding situations, when Calva does not get activated on startup.

Sorry, I'm not much of a writer 😞, but I think it's clear what I want to say. I'm open to suggestions for better formulating this.


PEZ:
I am pretty sure I managed to fix the activation issue.

PR #201, opened 2019-06-05
Update CIDER link

bbatsov:
I guess that's self-explanatory. :-)


PEZ:
I like that you are now a direct contributor to the Calva repo! :smile:

PR #206, opened 2019-06-06
Added a note for windows users to change shell

pikeview:
Added a note for windows users to change shell from Powershell to Cmd


PEZ:
Thanks!

PR #224, opened 2019-06-13
Workaround for powershell jack-in

xanatos1992:
This is just a current workaround fix for jack-in in powershell.

It changes the setting of the integrated shell to cmd.exe and after starting the task back to the user setting.


PEZ:
Wonderful. Thanks! And I agree the workaround is needed. We just need to remember to get back to this.

PR #247, opened 2019-07-08
Fixes ignored strings and regex

svdo:
Explanation: the inner `string` and `regex` already consume the close
pattern, so it has to be matched with look behind on the `ignore-string`
and `ignore-regex`.


PEZ:
Looks good to me. ❤️

svdo:
I'm submitting this PR because the string and regex cases are fixed by this. There are a few other tests (specifically `it "tokenizes quoted expressions"` and `it "tokenizes sets"`) that fail, but those are left out of scope for this PR.

PEZ:
Thanks for helping out with that house of cards!

svdo:
It was my pleasure!

PR #252, opened 2019-07-10
[Fix #251] Add support for forwardUpSexp

### LINKED ISSUE #251: Add forwardUpSexp
As Calva supports paredits `backwardUpSexp`, it should also support `forwardUpSexp`

slipset:


PEZ:
Thanks!

PR #258, opened 2019-07-19
Update README.md

matthewmcgarvey:
Just some typos and consistency changes I made as I was reading through it.  About to start going through `Clojure for the Brave and True` and I think this extension is going to make getting back into Clojure really easy!  Feel free to point out any changes you'd like reverted or worded differently 👍


PEZ:
Thanks for that help! The changes looked right and relevant to me, though, I am not am native speaker, so in some cases I just couldn't tell, and choose to trust your call.

PR #262, opened 2019-07-22
configurable shadowcljs edn file path

punit-naik:


PEZ:
Hello,

Just checking a thing here. Looking at the change I wonder if it does not work to just configure `projectRootDir`? And if not, maybe that mechanism is broken and should be fixed. It would be a more general fix for the problem, that would work for Leiningen and clojure-cli as well.

punit-naik:
Simply configuring `projectRootDirectory` won't work I guess.

Issue #96, opened 2019-01-07
Better colors for light themes


ghost:
<img width="106" alt="janusdeployment yml charli-account-provisioning-service 2019-01-07 14-52-25" src="https://user-images.githubusercontent.com/33130874/50798075-da13de00-128b-11e9-82af-adf44853dfbd.png">

This is pretty hard on the eyes. Quite frankly, I'd prefer to have a default of the main statusbar foreground color for the enabled repl and a slightly transparent version of that for the disconnected repl. Barring that, we should probably let the user override the colors somehow.

The "bracket-pair-colorizer" extension lets you specify colors as so:

``` json
    "bracket-pair-colorizer-2.colors": [
        "#AC362D",
        "#3BA3F8",
        "#E826AE",
    ]
```

while VSCode lets you override syntax/workbench colors on a per-theme basis like so:

``` json
"workbench.colorCustomizations": {
        "[Alabaster]": {
            "statusBar.background": "#F0F0F0",
            "statusBar.foreground": "#000028",
            "statusBar.noFolderBackground": "#F7F7F7",
            "statusBar.noFolderForeground": "#000028",
        },
        "[Night Owl]": {
            "activityBar.background": "#000C1D",
            "activityBar.border": "#102a44",
            "sideBar.background": "#001122",
            "sideBar.border": "#102a44",
            "sideBar.foreground": "#8BADC1"
        },
        "[Quiet Light]": {
            "statusBar.background": "#AC3543"
        }
    },
```

for customization purposes, I'd prefer the latter, but understand if we can't do this in a plugin.

cfehse:
This is fixed with version 2.0.49.

PR #270, opened 2019-08-03
Made statusbar color configurable

xanatos1992:
Add option to customize the statusbar color item.
This soves the issue #96 .


xanatos1992:
Okay first to your second point. 
just pushed something for this :) 

To your first point i dont really understand what color to use from the themes for example disconnected/connected status.

PEZ:
I agree about the theme colors. People will have to live with that they might need to change the colors when they switch theme.

This looks good, I think.

PR #275, opened 2019-08-07
Rework how the project directory is determined for jack-in and connect

PEZ:
First we try find the workspace root, _then_ we add custom sub dir

Hello, can you review this fix of `getProjectDir()` for me? It is not supposed to add any fancy new possibilities, just fix so that a custom project root (sub directory) should work in more cases. It is a very central function so I'd appreciate an extra brain on it. Cheers!


PEZ:
Please merge it if you think it is good enough. :smile:

kstehn (commented):
Looks good.
But i think i found a problem , and it might be the reason why i thought this is buggy.

When you Jack-In the Terminal says "Executing task in folder <firstFolder> ...."
After that i always canceled the process. This time i just let it run and it is the right Project.
But i have no idea where that might happen

PEZ:
I think I might have been using the wrong constructor overload for the Task all this time. Fixed that now. Can you try and see if it fixes the problem? If so we should have a much more reliable jack-in for projects not in the workspace root now.

calva/connector.ts
```diff
@@ -400,22 +400,13 @@ export let cljSession: NReplSession;
...
-                return p;
-            }
-            prev = d;
-            d = path.resolve(d, "..");
-        }
-        return path.join(util.getProjectDir(), subPath);
-    } else {
-        return path.join(util.getProjectDir(), subPath);
+    try {
+        const projectRoot = util.getProjectDir(),
```
kstehn: Small error here  it should be a` ;` not` ,`

kstehn:
Also like we discussed to remind you, add some documentation to the function that you are currently working on :).

PEZ:
Now I have made the project root something we need to initialize as the start of the jack-in and (standalone) connect processes. When the path is actually needed, this initialized value will be used. Should be both more efficient and less error prone.

I also added that doc string, finally. :smile: Which was a very good thing to add, because the way I had phrased the strategy used made me realize that this new way of finding the project files isn't implementing that strategy! This will now fail for the ”plain” folder, multi-project case. (I haven't tested that yet, but I am pretty certain.) Committing this anyway, so that I can get your help with sanity checking the general change for init-then-use.

Will restore the old new way of finding the project files now. Then, maybe, this is closing in on being ready to merge.

PEZ:
This now works for the plain folder multi project case. I've tested it with Leiningen, Clojure CLI and shadow-cljs. An inconvenience is that Calva remembers the jack-in and connect picker selections per workspace which can be confusing and trick people to accept the wrong defaults. I think we should fix that before shipping. What say you?

I'll make some tests with a real workspace now. Please put it to a test on your side as well.

PEZ:
VSIX package built and sent to `calva-dev` in Clojurians Slack. Let's give it a day or two. If no one reports a problem, then merge! : smile:

mrkam2:
Hi @PEZ I found that broke my Jack-In presets. I need "projectRootDir" setting so that I can choose _another_ project having the same source file. Is it possible to restore it?

mrkam2:
Sorry, I got it wrong. I was actually using a custom Calva build from https://github.com/BetterThanTomorrow/calva/pull/580.

PEZ:
So you're good now?

mrkam2:
> So you're good now?

Thanks for checking.

I need that PR to be merged (as I need projectRootDir setting). I have a temporary workaround for now.

Issue #277, opened 2019-08-12
Honor main arguments for clj and Leiningen aliases


PEZ:
As is evident from #276 and its [example project](https://github.com/mkarp/cljs-nrepl-exercise) Calva still has troubles jacking in to some not-all-that-crazy projects.

I think that what is lacking is that Calva insists on it's predefined main arguments, disregarding any such arguments specified in the alias (in the `clj` case), and not even offering to start with aliases (in the Leiningen case).

- [ ] `clj` (`deps.edn`)
- [ ] Leiningen

kstehn:
Just a side note here. 
For clj it is already possible to select a alias, just checked it myself. :)

PEZ:
We should improve on referencing the issues when committing fixes. :smile:

PR #279, opened 2019-08-12
Possible to use lein aliases

kstehn:
let the user choose what alias to run when there are any available related to #277


PEZ:
Nice progress here!

After discussions offline, let me try to summarize what we've concluded:

What to do with `main` args? (Am I right to assume that `main` is equivalent to `run` in Leiningen?).
* We could try see if the alias has main args/run and if so not provide our own, and if not add our own.
* We could just never provide our own in the presence of an alias.

**Leiningen specific**:
* We need to figure out how we want profiles to work with this. Either aliases+profiles is not supported (as it is now in this branch). Or we try to merge profiles from the alias with whatever the user selects.
* There need to be an option for ”no alias, please”.
  * Here always let the user select profiles.
* Calva should remember the aliases selection between jack-ins in the same project.

kstehn:
So with the last two commits.

All the Leiningen Specifics should be addressed.
I made it so now that you can choose a alias and profiles if you want too.

For the clj main-args i implemented the second point. Where we dont supply any main if a alias was choosen.

PEZ:
The Leiningen tests I've done so far seem to work. I'll try to be more creative. :smile:

For clj, after testing I see that it can't be unconditional to skip our main. Only if the alias has main args. Jack-in will always fail if there is no main at all.

kstehn:
Okay just did it. 
I check for the selected aliases if main-opts are available. 
If calva does not supply its own main.

PEZ:
Yeah, so I think this works now.

Nice with those docs telling about what steps are involved in cooking the command lines!

Issue #282, opened 2019-08-14
Add configuration for custom repl connect sequences


PEZ:
I want to be able to have a section about customizing the repl connection sequence to go with [About Calva Jack-in](https://github.com/BetterThanTomorrow/calva/wiki/About-Calva-Jack-in) that reads something like this:

## Customizing the REPL Connection Sequence

Many projects grow out of the template phase and call for custom developer workflows involving application start commands, customized REPLs, and what have you. Even some templates add this kind of complexity. To make Jack-in usable for a broader set of projects, Calva has a setting keyed `calva.replConnectSequences` which lets you configure one ore more connect sequences.

NB: _Connect sequence configuration affect Calva's Jack-in menu in the following ways:_

1. With _no sequence_ configured, Calva will prompt for the built in sequences it has that seems to match your project.
1. With _only one_ custom connect sequence specified, Calva will skip prompting for sequences and use your custom sequence.
1. With _more than one_  sequence specified, Calva will prompt with a menu consisting of these custom sequences.

A connect sequence configures the following:

* `name`: (required) This will show up in the Jack-in quick-pick menu when you start Jack-in (see above).
* `projectType`: (required) This is either "Leiningen”, ”Clojure-CLI”, or ”shadow-cljs”.
* `afterCLJReplJackInCode`: Here you can give Calva some Clojure code to evaluate in the CLJ REPL, once it has been created.
* `cljsType`: This can be either "Figwheel Main", "lein-figwheel", "shadow-cljs", "Nashorn", or a dictionary configuring a custom type. If omitted, Calva will skip connecting a ClojureScript repl. A custom type has the following fields:
  * `dependsOn`: (required) Calva will use this to determine which dependencies it will add when starting the project (Jacking in). This can be either "Figwheel Main", "lein-figwheel", "shadow-cljs", "Nashorn", or ”User provided”. If it is "User provided", then you need to provide the dependencies in the project, or launch with an alias (deps.edn), profile (Leiningen), or build (shadow-cljs) that privides the dependencies needed.
  * `isStarted`: Boolean. For cljs repls that Calva does not need to start, set this to true. (If you base your custom cljs repl on a shadow-cljs workflow, for instance.)
  * `startCode`: Clojure code to be evaluated to create and/or start your custom CLJS REPL.
[...]

PR #284, opened 2019-08-15
Custom repl connect sequences

kstehn:
This pull Request is still work in progress.
The target is to implement the Feature described in the issue #282 

First need to discuss here the default sequence description.

### DIFF
--- a/calva/nrepl/connectSequence.ts
+++ b/calva/nrepl/connectSequence.ts
@@ -1,0 +1,75 @@
+enum ProjectTypes {
+    "Leiningen",
+    "Clojure-CLI",
+    "shadow-cljs"
+}
+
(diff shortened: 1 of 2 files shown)


PEZ:
Nice with the intellisense help in adding the settings. Three things I found:
1. The schema seems to lack descriptions for some of the fields. So the user doesn't get help in understanding what each setting does.
1. I don't get any feedback on when I have missed adding required fields. (Not marked with `optional` in the issue descriptions means `required`. :smile:)
1. When there is a sensible default, I think that should be inserted automaically. I'm not sure if it is only `cljsType/isConnectedRegExp` where we have such a default, but there it should be `"To quit, type: :cljs/quit"`.

PR #285, opened 2019-08-15
[WIP] custom repl connect sequence

kstehn:
This pull Request is still work in progress.
The target is to implement the Feature described in the issue #282

First need to discuss here the default sequence description.


PEZ:
`cljsType/connectCode` should also be required.

Issue #288, opened 2019-08-17
Enable aliases from personal deps.edn profiles


gerred:
I have a series of global aliases with my personal configuration for REBL, etc., in the vein of https://github.com/seancorfield/dot-clojure. It would be handy to be able to use these aliases for jacking in!

PR #289, opened 2019-08-19
[WIP] Add myLeinProfiles settings for jack-in

PEZ:
Addressing #288 (which is about clj-deps aliases, but I started with Leiningen profiles, just because).

This is one way to do it. It adds a setting `myLeinProfiles`, which is an array of strings. These strings are then added to the Jack-in quick-pick menu and if selected to the jack-in command at `--with-profile`. Thus the user needs to specify this ”an extra time”, and risk getting it wrong. 

Another way to do it would be to read `~/.lein/profiles.clj`, `~/.clojure/deps.edn` and populate from there. But it is more work to get it right on all various operating systems and setups, and also would maybe pick up a lot of aliases/profiles, that don't make sense...

This is more KISS, I would say. What say you?

### DIFF
--- a/calva/nrepl/jack-in.ts
+++ b/calva/nrepl/jack-in.ts
@@ -95,11 +95,12 @@ const projectTypes: { [id: string]: connector.ProjectType } = {
             }
 
             if (defproject != undefined) {
-                let profilesIndex = defproject.indexOf("profiles");
+                const profilesIndex = defproject.indexOf("profiles"),
+                    myProfiles = state.config().myLeinProfiles;
                 if (profilesIndex > -1) {
                     try {
-                        const profilesMap = defproject[profilesIndex + 1];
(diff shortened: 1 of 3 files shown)


kstehn:
I think it is fine to do this for now.
I didnt found any API from vscode to get the home folder.
Otherwise i would just use the home folder and path.join to get the correct path to both files.
This should also be then os save :)

PEZ:
I'll go ahead with using settings then. I suspect there is a reason `clojure` has a command line option to report which config file it is using...

There is also that all aliases and profiles might not make sense for jack-in. I have Sean Corfield's `:new` alias in my `~/.clojure/deps.edn` for instance. Not material for jack-in and would only clutter the menu.

PEZ:
I merged this into #285, closing.

Issue #229, opened 2019-06-17
evalCurrentFormInREPLWindow without cursor moving to REPL


grierson:
When I run the 'evalCurrentFormInREPLWindow' command my cursor moves to the REPL buffer.
Is there a way for the cursor to remain in the text buffer after running the command?

```
(+ 1 1)|
calva.evalCurrentFormInREPLWindow
(+ 1 1)|  ;; Cursor should remain here, not move into REPL buffer.
```

PEZ:
Hello, thanks for suggesting how Calva can improve!

Can you elaborate some on why you want this behaviour?

grierson:
> Can you elaborate some on why you want this behaviour?

* REPL window is prettier than the output window. (Colours and Pretty Print support)
* I like to evaluate each expression individually as described in this post [REPL DEBUGGING: NO STACKTRACE REQUIRED](http://blog.cognitect.com/blog/2017/6/5/repl-debugging-no-stacktrace-required). With the cursor jumping between the text buffer and REPL buffer after running the 'calva.evalCurrentFormInREPLWindow' I have to keep switching back to the text buffer.

alzadude:
> Can you elaborate some on why you want this behaviour?
> 
> 
> The output windows works like this however the REPL window has prettier output. (Colours and Pretty Print support)
> I like to evaluate each expression individually as described in this post REPL DEBUGGING: NO STACKTRACE REQUIRED. With the cursor jumping from the text buffer to the REPL buffer every time I run 'calva.evalCurrentFormInREPLWindow' I have to keep switching back to the text buffer.

@grierson @PEZ I've also hit this issue and I would also prefer the cursor to stay in the editor.. Glad I checked the existing issues before creating a new one! 🙂

Lakret:
I also hit this problem. It's also very different from how all other REPL-enabled languages work in VS Code, for example in Elixir LS and in Julia LS cursor stays in the editor. The typical workflow is to evaluate stuff from the editor, change it in the editor, and re-evaluate it, so jumping to REPL seems like more of an outlier than normal workflow. Having an ability to disable this behaviour would be highly appreciated :)

PEZ:
Point noted, @Lakret . However, for Calva the typical workflow is rather to evaluate it **in** the editor. Of course, @grierson's point that the REPL window has prettier output applies still, but I wonder if those `LS` extensions support inline evaluation?

PEZ:
Released!

PR #294, opened 2019-08-28
Keep focus in editor window when eval current form in REPL window

bpringe:
https://github.com/BetterThanTomorrow/calva/issues/229


PEZ:
Thanks. I'll let this action of yours convince me that we should make this change. 😄 ❤️

PR #298, opened 2019-09-04
Tokenize newlines separately

### LINKED ISSUE #292: Pretty printing in REPL window is broken
Here is a demo using an example from https://clojuredocs.org/clojure.pprint/pprint

To evaluate `big-map` I used `ctrl+enter`, which should pretty print in the REPL window.

image

This is what the output looks like if I use `clojure.pprint/pprint`.

image

PEZ:
The lexer now tokenizes newlines into separate tokens. So that the result renderer can create divs for each newline, honoring pretty print.
Fixes #292

Will need help looking at this change as well as testing that it doesn't break stuff.


bpringe:
I've found an issue that I think it actually on master now (I found it on this branch and verified that the same thing happens on master). I can make a separate issue if you want.

Here are the repro steps:

1) Run calva jack-in
2) Switch to a project file's namespace - (open file then `ctrl+alt+c, ctrl+alt+n`)
3) Send require statement from code file (in a comment block) to the REPL window for eval - `ctrl+alt+c, e` on something like `(require '[hello.world.test-req :as tr)`
4) Run *ns* in the REPL window where the prompt looks like the previously switched to namespace, but notice that it says you're in the `user` namespace. If you try to eval something from the namespace of the prompt, it will say undefined because you're REPL is actually in the `user` namespace.

This issue does not happen on my custom build from a previous calva version. I think some changes merged to master since 2.0.33 may have introduced this issue. If I run the `require` statement directly in the repl, the namespace does not change to `user`.

PEZ:
Yes, that would be really good to track in an issue of its own.

PR #299, opened 2019-09-05
Add circleci

kstehn:
Add circleci build.
Automates the test running and also creates the vsix artifact.


PEZ:
I'd like the artifact's version to contain the PR number and commit ID. Can I haz?

PR #316, opened 2019-09-13
Wip/fix issue 305

cfehse:
The current task API for vscode provides no callback to allow some house keeping before the task execution is terminated. The vscode.tasks.onDidEndTask handler is called after execution is terminated. This is too late to terminate the repl processes from the repl client. I implemented a solution which handles all situations where the task can be terminated from the extension itself:

- the extension is deactivated
- the extension created a new repl task
- the repl window is closes

If the user chooses to terminate the task from the task management UI (e.g main menu) there still will be a leftover java process. Due to this I removed the message to do so from the output channel.

Take a close look on this and perhaps you can accept the request. 

Thank You and have a nice weekend!

### DIFF
--- a/calva/extension.ts
+++ b/calva/extension.ts
@@ -228,6 +228,7 @@ function activate(context: vscode.ExtensionContext) {
 
 function deactivate() {
     state.analytics().logEvent("LifeCycle", "Dectivated").send();
+    jackIn.calvaJackout();
     paredit.deactivate()
 }
 
--- a/calva/nrepl/jack-in.ts
+++ b/calva/nrepl/jack-in.ts
@@ -4,14 +4,30 @@ import * as fs from "fs";
 import * as path from "path";
 import * as state from "../state"
 import * as connector from "../connector";
+import {nClient} from "../connector";
 import statusbar from "../statusbar";
 import { parseEdn, parseForms } from "../../cljs-out/cljs-lib";
 import { askForConnectSequence, ReplConnectSequence, CljsTypes } from "./connectSequence";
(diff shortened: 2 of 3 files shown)


PEZ:
Thanks!!! We will have a look.

@kstehn, can you test this on Windows? I'll test it on Mac.

kstehn:
sadly one java process remains when i terminate the task
the repl process is terminated but the one started by the lein-repl isnt

cfehse:
@kstehn what exactly did you test? If you terminate the task through the vscode ui one process is left. Closing vscode with a running repl, open a new repl and closing the repl window should close all java processes.

kstehn:
ah okay i closed the window repl now and all java processes are gone, good job :) 

is there a way to know before hand when the terminate task gets executed so the jackOut can be called before it gets killed?

cfehse:
@kstehn I found no way to intercept the task termination initiated by the vscode ui itself. The notification about task termination comes to late. At that point the TaskExecution.terminate() was already called which killed the cmd.exe and java.exe process that holds the repl. The leftover process holds the leiningen project. There is no way to kill this process in a save way. One can kill all Java processes but there may be Java processes not related to the task at all.

The only way to safely remove the task related processes is to kill them via the repl client before vscode terminates the task. So I implemented the mechanism at all points where the extension itself is in control (deactivation of the extension, creating a new repl connection while a repl is running und closing the repl window). I think there is not much more one can do. The vscode task API is not meant to be used to run highly configurable task which rely on user interaction to build the correct task configuration (in our case the commandline) .

PR #322, opened 2019-09-16
Added task "create-vsix" to build local visx.

cfehse:
I think it would be nice to have a UI callable task to build a release vsix file locally. The "vsce package" command runs the "vscode:prepublish" task implicitly to compile everything as release. Perhaps you do not mind adding this small change.

Thanks!

### DIFF
--- a/package.json
+++ b/package.json
@@ -1164,7 +1164,7 @@
         "disabled-release": "npm i && npm run update-grammar && npm run release-cljs && tsc -p ./",
         "vscode:prepublish": "rimraf ./out && npm run release",
         "disabled-vscode:prepublish": "npm run release",
-        "create-vsix": "rimraf ./out && vsce package",
+        "create-vsix": "vsce package",
         "postinstall": "node ./node_modules/vscode/bin/install"
     },
     "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1164,6 +1164,7 @@
         "disabled-release": "npm i && npm run update-grammar && npm run release-cljs && tsc -p ./",
         "vscode:prepublish": "rimraf ./out && npm run release",
         "disabled-vscode:prepublish": "npm run release",
+        "create-vsix": "rimraf ./out && vsce package",
         "postinstall": "node ./node_modules/vscode/bin/install"
     },
     "dependencies": {


PEZ:
Hello. Yes, maybe it makes sense to surface this a bit.

The `rimraf` seems unnecessary since that is done in prebublish. (Where we probably should rimraf `cljs-out` as well, but that's another matter.)

Personally I always terminate the watching tasks before building a package. Don't know how necessary this is, but just feels wrong to have tasks running that are potentially writing to the same directory. Not sure if this is possible to build into the script or task for building the vsix.

cfehse:
@PEZ:
- Yes the rimraf is unnecessary.
- Yes it would be nice to terminate the watching tasks. I take a look what can be done there.

cfehse:
@PEZ: there seem to be no way to stop running task from outside the vscode UI. One good solution would be to have the debug files live in out\debug and the release ones in out\release or at least have the packaging script compile the sources into an independent directory. That way neither action would interfere with each other. I will close this pull request and maybe come up with a better solution later. 

Thanks for your feedback!

PEZ:
If you come up with a solution for this, that would be great. I haven't figured out how to get around that there is only one `main` entry in the manifest. But your reasoning makes me think... maybe we can just copy the needed source files down in a git-ignored release directory and have the package built from there? Now I see that maybe that is what you are suggesting... Anyway, please ponder this if you have the time and energy. Calva would benefit from this being solved.

cfehse:
We are getting something done there. *gg*

cfehse:
@PEZ: Perhaps you can give me some advice on the build scripts:

;; The watch related scripts for development (debug mode)
;; in order of dependencies. 
"compile-cljs"
"watch-webpack"
"watch-ts"

;; These are the release related scripts (release/production mode)
;; in order of dependencies.
"update-grammar"
"release-cljs"
"webpack-release"
"release"
"vscode:prepublish"

;; These are some release process dependent scripts
;; or something
"bump-version"
"postinstall"

;; These seem to be unused or test entries
"compile"
"NOT-USED-start"
"NOT-USED-repl-window-dev-server"
"disabled-release"
"disabled-vscode:prepublish"

Is that right? Just to get a basic understanding for the configuration. To use JSON for configuration files was one of the worst decision the software engineering community made. *gg*

PEZ:
Hi! Yes, that is pretty spot on. I am sorry for the mess, not entirely JSON's fault. :smile:

I think that the ”update-grammar” script probably should be part of the watch/development as well. But, maybe that is rather a process of itself that we should document, and then we will see what kind of script support it should have. (We do not run the grammar tests as part of the CI today, even. That we really should do.)

Issue #329, opened 2019-09-20
Fixup file name and line numbers when evalulating a form


marcomorain:
When I eval a chunk of code using the "Calva: evaluate current top level form..." commands,  the file name is auto-generated as `form-init2144587467489926643.clj`, and the line numbers start from 1 for the chunk that I have evaluated.

This leads to confusing data in a stack trace. Let's say I eval a chunk of code in a file named `storage_test.clj`, and run a test that fails. The output is as follows:

```
FAIL in (storage-stores-stuff) (form-init2144587467489926643.clj:12)
```

If instead, I evaluate the entire file, and then run the test, the output changes to:

```
FAIL in (storage-stores-stuff) (/Users/foo/bar/baz/storage_test.clj:43)
```

I can click on the string `/Users/foo/bar/baz/storage_test.clj:43` in VS Code and be taken directly to that line of code. I cannot click on `form-init2144587467489926643.clj:12`, which slows me down, and makes it harder to debug problems.

Suggestion:

When Calva evaluates a form, it should specify the filename, column and initial line number when calling nrepl:

https://github.com/nrepl/nrepl/blob/cdd688f22e65580e9b52d9a8def9ad42a530914a/src/clojure/nrepl/middleware/interruptible_eval.clj#L157-L159

This will make everything line up much better.

PEZ:
Thanks. Looking at this a bit. I am not sure I can fix the actual case you mention with re-rerunning tests (haven't even looked at that yet, so it could be it is entirely fixable), but for regular evaluations it certainly works providing that info.

Will look at it a bit more and see what I can do.

PEZ:
How do you run those tests, @marcomorain ? My output does not match what you show there...

PEZ:
Can you test this VSIX package out, @marcomorain ? And let me know if it makes the difference you wanted, or not.

marcomorain:
How do I try the package out @PEZ ? I don't see a file anywhere?

PEZ:
Sorry, missed attaching the vsix. :smile:

PEZ:
@marcomorain , if you want to dig in to this some, please do. Or, if not, let me know and I'll have another go at it.

bpringe:
The test runner (at least now, not sure about back when this issue was created) shows the file and line number of the failing test when using the test commands as well as when running the test directly from the repl window using `clojure.test/test-var`.

[...]

PR #330, opened 2019-09-21
Wip/inline eval file context

PEZ:
First stab at addressing #329. Only makes a difference for inline evaluations right now. Not sure if that is enough... But I guess it is an improvement anyway. Can you have a sanity check look @kstehn ?


calva/evaluate.ts
```diff
@@ -42,9 +42,18 @@ async function evaluateSelection(document = {}, options = {}) {
...
+                const line = codeSelection.start.line,
+                    column = codeSelection.start.character,
+                    filePath = doc.fileName,
+                    context = client.eval(code, {
+                        file: filePath,
+                        line: line,
+                        column: column,
+                        stdout: m => out.push(m),
+                        stderr: m => err.push(m),
+                        pprint: !!pprint
```
kstehn: Here i dont understand the `!!` isnt it the same as not having those?

PEZ: It's a leftover from when we didn't use types. Basically it just forces it to be either `true` or `false` from merely being just truthy or falsy. It should be fixed with better typing control, I think.

PR #333, opened 2019-09-21
Enhancement REPL window handling / nREPL menu button

cfehse:
Fixes issue #333 .

**Motivation**
Inspired from the conversation about the usage of the REPL window for experienced users the intention is to enhance the REPL window handling and the "nREPL" menu. There are commands to open the different REPL windows but for inexperienced users like me it may be difficult to find - in fact at the beginning I closed the REPl accidentally and did not find a quick way do get it back. There are in fact 2 commands and you have to now the right type of REPL window to open. Choosing the wrong type leads to a strange error message telling of some undefined object. 

**Implementation**
Bundling all nREPL related commands into the nREPL command selection (pressing the "nREPL" button in the status bar). Depending on the state of the extension different commands will be included in the selection. A new disconnect command was added to provide a quick way to terminate the current REPL session. Meaningful error messages were added in case a command could not be executed. A rudimentary positioning management for the REPL windows was added. During runtime the windows keep there last position between closing and reopening (the start default was not changed). 

**Platforms tested**
Windows and linux.

**Details**
_package.json_
- added `calva.disconnect` command.

_calva/extension.ts_
- removed unused import.
- replaced the implementation of the `calva.jackInOrConnect` command with a function in `calva/nrepl/jack-in.ts`.
- added register for `calva.disconnect` command.

_calva/nrepl/jack-in.ts_
- added implementation for `calva.disconnect` command.
- added implementation for `calva.jackInOrConnect` command which includes state depended commands into the selection.

_calva/repl-window.ts_
- removed unused import.
- replaced  implementation for `calva.openCljReplWindow` command to include error messages.
- replaced  implementation for `calva.openCljsReplWindow` command to include error messages.
- added support function to see if a specific type of REPL windows is visible.
- changed the `panel.onDidChangeViewState()` handler to save the `viewColumn` for the specific REPL window. 
- changed calls to `vscode.window.createWebviewPanel()` and `panel.reveal()`to use the saved values for `viewColumn` for the specific REPL window. 

Please take a close look and perhaps you can merge these changes. 

Thank you!


PEZ:
Cool. I will test this out. But from just reading your description I wonder if we shouldn't change the column behaviour of the initial open as well, to use any saved position (otherwise default to column two).

PEZ:
First some bureaucracy. Can you open an issue for this? I like to have the problem separated from the solution, and it will also be easier to find for someone filing a similar issue.

PEZ:
Tried it now. A few initial thoughts:

1. I like the idea of using the REPL button like this
2. I think that when you are connected the jack-in and connect commands should not be included in the REPL-button menu.
3. I think I see the option for opening the REPL window sometimes failing to follow when the active editor type changes. (So showing ”Open CLJS REPL” when it is a CLJ file, and such.)
4. The command for opening the REPL window is stil there even when the window has focus.

About 3 and 4. This makes me wonder... Should it instead be a command that both opens the right REPL window and also changes the namespace to the ns of the file? And also, when the REPL window is having focus, the command should maybe instead open the corresponding file. (In both cases changing focus to the opened window/file). If such a command had a nice keyboard shortcut, it would make it swift to go between files and repl window and back at the press of a key (so to speak).

PEZ:
Also, I have in parallell with this change of yours added a command to that same menu: https://github.com/BetterThanTomorrow/calva/pull/331

Jotting this down here so that we will remember to merge carefully and also to give more of the full picture to us, regarding this button.

cfehse:
@PEZ I expected that you worked on issue #331. *gg* So if you merge it into dev I will merge it into the PR.
Regarding your other thoughts I will reply later.

cfehse:
>I think that when you are connected the jack-in and connect commands should not be included in the REPL-button menu.

From a consistency point of view I would agree but experienced user may expect the connection option to be selectable because this was always the fasted method to change / reconnect to the REPL.  

>I think I see the option for opening the REPL window sometimes failing to follow when the active editor type changes. (So showing ”Open CLJS REPL” when it is a CLJ file, and such.)

The selection which REPL window type open command is shown is not dependend on the file type. It only depends on the session type (is there a session for Clojure or/and ClojureScript present or not and is the corresponding REPL window visible). 

> Should it instead be a command that both opens the right REPL window and also changes the namespace to the ns of the file?

One command is not sufficient. There may be projects  and REPL connections which have both session type connected the same time. I tried this with the shadow-cljs sample (https://github.com/shadow-cljs/examples) cljs-react-app and with this project you can open a Clojure and a ClojureScript REPL window at the same time:

Project-with-2-repl-windows

>And also, when the REPL window is having focus, the command should maybe instead open the corresponding file. (In both cases changing focus to the opened window/file). If such a command had a nice keyboard shortcut, it would make it swift to go between files and repl window and back at the press of a key (so to speak).

There is not direct connection between the REPL windows and the file types. The command implementations for sending data to the REPL window in fact select the correct REPL type and send the data to the corresponding REPL window - and opens that window, if it does not exists.

PEZ:
Not sure I am following any of that. Or that you are following what I wrote. :smile:

The fastest way to reconnect/re-jackin is to use the keyboard shortcuts for it. The menu for the nREPL button is more of convenience/exposure. I am not so worried about changing it for experienced users, as I am of confusing new users. Calva constantly changes.

As for the other things. Can we maybe find each other on Slack? It will be easier to figure it out in a chat or a voice call.

cfehse:
@PEZ I think I must be invited or ask somebody to get into the Clojurians Slack? I don't now. 

> Not sure I am following any of that. Or that you are following what I wrote. 😄

I thing I followed you quite well. *gg*

PEZ:
I moved my request for the switching command to an issue of its own: #339

You join the Clojurians Slack here: https://clojurians.herokuapp.com

Looking forward to see you there!

cfehse:
@PEZ I merged the current changes from the upstream/dev branch.

PEZ:
When I test this and have both CLJ and CLJS REPLs connected, I don' see an option to open the CLJS REPL Window. Only the CLJ one.

cfehse:
I did see both - perhaps you have a sample project to test. Thanks

PEZ:
I tried with this project: https://github.com/PEZ/idg-code-night
It is a legacy Figwheel project.

PEZ:
Having a `clic` file open. It shows only the option to open the Clojure REPL window. I can get it to instead show the option to open the ClojureScript REPL (then the Clojure option disappears). Not sure how I can make this happen, but clicking around some between the REPL window and the editor. After that I cannot get the option to open the Clojure REPL back.

And I've never seen both at the same time.

Sorry for a messy bug report, but I don't know how to describe it better...

cfehse:
@PEZ No problem. I take a look at this. *gg*

cfehse:
I changed the version of webpack-cli to version 3.3.5 to be able to use it mondays on windows. No joke see the issue #926 in the webpack-cli project. This should be done in the dev branch regardless of this PG.  The code now used `utilities.getSession()` as you advised for the checks if a specific session exists. 

>When I test this and have both CLJ and CLJS REPLs connected, I don' see an option to open the CLJS REPL Window. Only the CLJ one.

This happens if both windows exist but one of the windows is obscured in the editor bar (so not visible). I fixed that.

> I think that when you are connected the jack-in and connect commands should not be included in the REPL-button menu.

I changed that accordingly.

> And also, when the REPL window is having focus, the command should maybe instead open the corresponding file. (In both cases changing focus to the opened window/file). If such a command had a nice keyboard shortcut, it would make it swift to go between files and repl window and back at the press of a key (so to speak).

I am not sure if I understand 100% what you mean but perhaps we should delay this to another PR to make the changes here not too large. 

> But from just reading your description I wonder if we shouldn't change the column behaviour of the initial open as well, to use any saved position (otherwise default to column two).

I think this should be done separately a well. 

What do you think? Thanks for the feedback!

PEZ:
I'm all for separating #339 to another PR and we can do the same with initial column thing.

I will test this and see how it feels now.

@kstehn : can you help give the code a read-through? I will too, but always good with your eyes on it.

PEZ:
Also, yesterday I created a [Pull Request Template](https://github.com/BetterThanTomorrow/calva/blob/dev/docs/pull_request_template.md), with a checklist of things to remember to do. You can probably tick most of the boxes, but anyway, check it out.

cfehse:
@PEZ I did. Next time with the template.

PEZ:
Now I get only the ”Disconnect from nREPL server” option.

cfehse:
> Now I get only the ”Disconnect from nREPL server” option.

When you doing what?

PEZ:
Clicking the `nREPL` button.

cfehse:
You are not connected, there are no REPL windows open and the command shows only the disconnect option?

PEZ:
No. I am sorry, should have said that I am connected. :smile:

When I am not connected the menu contains the three items I expect.

cfehse:
@PEZ Sorry a can't reproduct that. Not on windows neither on linux.  I used your test project (https://github.com/PEZ/idg-code-night) and all other projects I tested on. If there is an issue I need more information. Thanks.

cfehse:
@PEZ If the REPL window(s) are open they are not include in the selection. If not the specifc command to open the window is include. It works. *gg*

calva/repl-window.ts
```diff
@@ -26,14 +26,9 @@ export function isReplWindowVisible(mode: "clj" | "cljs" = "clj") {
...
-        // the webview may already be disposed.
-        if(!replWindows[mode].panel.visible) {
-            return(false);  
-        }
-    } catch(e) {
-        return(false);
-    }
+    // Do not check for visability because the REPL window can 
+    // exist and be obscured in the editor bar. See if it exists 
+    // but not if it is visible.
```
PEZ: This function needs another name after the change. :smile:

cfehse: @PEZ `existsReplWindow() `?

cfehse:
@PEZ 

- Renamed `isReplWindowVisible()` to `existsReplWindow()`.
- Change PR title and comment to reflect issue #333. 
- Added reference to the issue in `CHANGELOG.md`.

Issue #218, opened 2019-06-10
[2.0] Printing from go-loop not working in REPL window


bpringe:
I realized this issue when running a larger system with multiple go-loops. I saw printing from maybe one of multiple go-loops, so I don't know that it _never_ works, but I have found how to reproduce it with a small example from clojuredocs.org. Link for reference: https://clojuredocs.org/clojure.core.async/go-loop#example-542c8833e4b05f4d257a297a

![image](https://user-images.githubusercontent.com/12722744/59231574-cb1cb900-8ba6-11e9-8d8e-c171ac907d5c.png)

As you can see, nothing is printing, as it shows it should be in the clojuredocs example. I checked the terminal that the REPL is running from as well, and nothing was there either.

FYI, I'm using Windows 10 and cmd as the shell. I don't know if that matters.

Also, this is what happens if I run the code in a repl straight in cmd (also same in powershell). When I hit enter/return I can see the print statements bundled together.
![cmd-go-loop-printing](https://user-images.githubusercontent.com/12722744/59232293-5eef8480-8ba9-11e9-8ff0-a930eaf95752.gif)

My coworker confirmed that this printing works correctly in emacs/cider:
![image](https://user-images.githubusercontent.com/12722744/59232401-cd344700-8ba9-11e9-9c48-1ba5bf3f9102.png)

PEZ:
OK. So I know what is going on. It's here:

https://github.com/BetterThanTomorrow/calva/blob/master/calva/nrepl/index.ts#L224

When the repl client gets a `state: ["done"]` message, it resolves the promise and returns `true` and subsequent `out` messages no longer have anyone listening to them.

Not sure how that should be handled... If I return false instead I get the prints, but I will have a lot of promises hanging around if I do that.

@bbatsov , do you have any advice for me here?

bbatsov:
That's a tricky one and in CIDER I solve it by just mapping a callback to the id of the request, so responses are properly handled even after "done" is received. Basically there's a mapping between each request and some lambda that gets called to process responses with that request id. Probably not the nicest design, but it gets the job done.

Alternatively you can just have a generic callback for out/err messages that is used when some message arrives either without an id or it has been fully processed already (you've received "done").

PEZ:
Thanks. I guess the promise is such a mapping. Hmmm...

I was thinking about doing the generic callback thing, but got stuck on how to allow the request to be interrupted.

bbatsov:
[...]

Issue #327, opened 2019-09-18
Make pretty print a mode, removing the need for doubling the evaluation commands


bpringe:
I believe pretty printing should happen by default when things are evaluated in the REPL window, including when forms are sent from the editor (send top-level, send current).

This should probably be included as a setting that is turned on by default, so users can turn off pretty printing if they would like to.

PEZ:
From offline talk about this: Let's make it an easy to use toggle. Probably even a status bar button.

bpringe:
Had this on my todo list 😉 - should have mentioned before.

PEZ:
Not sure why this did not autoclose? Maybe you didn't include the ”fixes” message in a commit message?

Anyway. Now released!

bpringe:
Weird! The docs say "resolves" works as well, and the PR shows it should have closed it.
![image](https://user-images.githubusercontent.com/12722744/65819386-b1909300-e1d0-11e9-9706-567502fca2cd.png) 
Anyway, thanks for the release!

PR #334, opened 2019-09-21
First try for a dafault msg handler for the nrepl

kstehn:
This should address the issue that they are no output outside a evaluation.

Adresses the issue #218


PEZ:
Cool!

Checking here. Whenever there is output ”out of band” we will print it in the REPL window. Regardless of if it is a result of an inline eval or not? We do not know here if it is an inline eval or something initiated in the REPL window, so we can't really route it ”right”, but someone like me would question using the REPL window for the output. I seldom use that window, and instead work from the editor files. Maybe it should be a setting where this kind of ”async” output should go?

kstehn:
That is correct every time we dont find a handler for it, it gets printed.

Well why not also output it to calva says?

PEZ:
Or there could be a separate output channel where we send all output. If it is ”within band” it also gets printed with the results.

It could even be the ”Calva Connection Log” (renamed, but just somewhere where the user knows to check for all stdout and stderr stuff ever resulting from an nREPL message...

kstehn:
I could get behind the extra output channel.
Not sure how you feel about this, but i think it would be also nice to have it pprinted.

PEZ:
> i think it would be also nice to have it pprinted.

Here I think that #335, that @bpringe is working on, might come in handy.

(But also, maybe stdout stuff is seldom subject to pprinting?)

kstehn:
@bpringe 
could you test this if it fix your use case :)

bpringe:
I'm not sure which issue of mine this would fix, but for these issues I have created it does not affect them - #218 and #327. I'm working on 327 currently.

kstehn:
@bpringe 
it should fix #218 like i wrote in the PR message
and when i tested this with @PEZ yesterday it worked like for emacs

bpringe:
@kstehn Sorry, I don't know what happened last time but I thought it was not working as expected, and I also missed your note about #218 up above. I now see the output in the Calva says window when I run the go-loop with println statement as in #218. I like the "Send Async Output To" setting as well! Thanks.

PR #345, opened 2019-09-25
[Fix] Simplify build system

### LINKED ISSUE #346: [Buildsystem] Simplify build system configuration
I think the current build system configuration could be a bit simplified do make the usage easier for developers and to improve the automation in the CI/CD process. I would suggest the following changes:
- Only one watch task for development which starts both watchers concurrently. The current configuration has two tasks which must be started in a specific order. Another disadvantage of this configuration is that die output of the first task is obscured by the second task.
- Replace all relative path values in `webpack.config.js` with resolved fully qualified ones.
- Get rid of the `IS_DEV_BUILD` environment because it is only used to select which webpack package to build for development and production. Evaluate the `--mode` commandline parameter from webpack in `webpack.config.js` instead.
- Change the script configuration in `package.json` that it is possible to build the extension from a clean cloned directory only by running `npx vsce package`.

cfehse:
This pull request:

Fixes #346

Tested on following platform(s):
- [x] Windows
- [x] Linux
- [x] macOS

`webpack.config.js`
- Added comments to make the configuration more explainable. 
- Replaced all relative path values with resolved full qualified ones.
- The selection which packages to create is changed to evaluate the webpack `--mode`parameter.

`package.json` / `.vscode/tasks.json` / `.vscode/launch.json`
- Only one build task. This is simpler and  no order of execution required.
- Using the `concurrently` module for the `watch` script allows to see the output from both watchers (`webpack` and `tsc`).
- Removed the environment `IS_DEV_BUILD` because the the webpack `--mode`parameter is now used instead (see `webpack.config.js`).
- The build is now able to build from a clean `git clone` directory by only running `npx vsce package`.

I have:

- [x] Referenced the issue I am fixing/addressing.

Ping: @pez, @kstehn

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -16,8 +16,7 @@
             "sourceMaps": true,
             "env": {
                 "CALVA_DEV_GA": "FUBAR-69796730-4",
-                "DEBUG": "no-debug-universal-analytics",
-                "IS_DEV_BUILD": "YES"
+                "DEBUG": "no-debug-universal-analytics"
             },
             "outFiles": [
                 "${workspaceFolder}/dist/**/*.js"
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -4,31 +4,15 @@
   "version": "2.0.0",
   "tasks": [
     {
-      "label": "Watch TS source",
+      "label": "Calva Watch",
       "type": "npm",
-      "script": "watch-ts-src",
+      "script": "watch",
       "isBackground": true,
       "group": {
         "kind": "build",
         "isDefault": true
       },
       "problemMatcher": "$tsc-watch"
-    },
-    {
-      "label": "Watch REPL Webview src",
-      "type": "npm",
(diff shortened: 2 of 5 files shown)


kstehn:
That is pretty awesome.
From my side i would consider merging this. 

As a reminder the Wiki for how to Contribute should then also be updated.

PEZ:
This is as awesome as it gets. I say add the issue and reference it and then we'll merge it.

I don't think it should be a changelog entry, though. At least, so far, we haven't logged changes to the dev an build processes there.

PEZ:
I tested it on MacOS. Works like a charm! It is something I have wanted to achieve for so long, but have lacked the knowledge about how to do it. You are a blessing to the Calva project, @cfehse!

cfehse:
Housekeeping is done.

Ping: @PEZ, @kstehn

PEZ:
And now I have updated the wiki. A sweet feeling removing that extra step.

PR #352, opened 2019-09-27
Make webview debugable in vscode Webview Developer Tool

cfehse:
One last thing - which was some kind of a "flaw" to me - about the build system I want to change is that the webview is not debugable either in vscode nor in the Webview Developer Tool from vscode.  This can relatively easy be a achieved by changing the property `devtool`  to `'eval-source-map'` in `webpack.config.js` for non production modes. Because of that the content security policy for the webview must now include `'unsafe-eval'` for the `script-src`. 

The other changes are more of a housekeeping kind. I removed the unused scripts in `package.json` to start the webview in the webpack dev server (these scripts did not work anyway for me- at least not in this configuration). I removed the then unused `calva/webview/index.ts` file as well.  I added `npx i` to the `prewatch` script. This way a new developer can simply clone the repository, open vscode in the directory, select `Run Build Script...` and is good to go without further action. The dependency checks make the watch script take a bit longer before ready than before but I don't think the impact is to big. What do you think?

I have:

- [ ] Referenced the issue I am fixing/addressing.
- [ ] Updated the `[Unrleased]` entry in `CHANGELOG.md`, linking the issue(s) that the PR is addressing.

Ping: @pez, @kstehn

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -12,7 +12,3 @@
 !README.md
 !CHANGELOG.md
 !LICENSE.txt
-
-# When not using webpack to bundle these need to be on the whitelist
-# !node_modules/**
-# !cljs-out/**
--- a/assets/webview.html
+++ b/assets/webview.html
@@ -2,7 +2,7 @@
 <html>
 
 <head>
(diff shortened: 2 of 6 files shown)


PEZ:
> I removed the unused scripts in package.json to start the webview in the webpack dev server (these scripts did not work anyway for me- at least not in this configuration)

Yeah, probably a good call. I kept them around because hoping I could get the behaviour back where I could run the repl window standalone, but that will be less important if we get it debuggable win Webview tools.

cfehse:
@PEZ The PR is now ready.

PR #360, opened 2019-09-29
Fix issue 238 argument lists in completion / hover 

### LINKED ISSUE #238: [Feature Request] show argument list of fn
In Cider, when you type out a call to a function, eg 
```clojure
(core/reader? ) ; there is now a space after the `?`
```
Cider will show you the argument list of the function in the mini buffer, like so
Screenshot 2019-06-28 at 10 19 15

It would be very nice if Calva could do something along those lines.

cfehse:
Fixes #238,  #366,  #367,  #368

- Added the argument list to `HoverProvider` and `CalvaCompletionItemProvider`.
- Added a mechanism to load a file into the REPL if the namespace of the file does not exist in the REPL. This is necessary because if the namespace with which the nrepl command 'info' is called (the namespace of the file) no data is returned. So the file must be loaded to retrieve data. Another advantage of this approach is that if the file was loaded information of the elements contained in the file can be displayed as well.
- Added activation events `workspaceContains:**/project.clj` and `workspaceContains:**/shadow-cljs.edn` to allow the extension to be activated faster. 
- Improved the `state.initProjectDir()` function to use `vscode.workspace.workspaceFolders![0]` as a second starting point in case no file is open in the workspace is not multi-root. This conforms to our [wiki entry](https://github.com/BetterThanTomorrow/calva/wiki/Workspace-Layouts) about the supported workspace layout. Made the function robust if no document is defined.
- Made the code robust in case a jack-in task exists prematurely an no REPL client is connected.
- Catch exception during `fs.watch()` in `executeJackInTask()`. These exception is according to the `fs` team by design if a given directory to watch does not exists or the underlying filesystem (network filesystems for example) does not support the operation. During the jack-in process this may occur if the user selects the wrong type of project - e.g. `Leiningen + shadow-cljs` if there is no shadow support in the project. In case the exception is detected, the jack-in task is canceled.
- Added the `file` and `untitled` schemes to all document selectors. These are the vscode supported schemes. The addition prevents the vscode warning message ` Extension '<extension>' uses a document selector without scheme. Learn more about...`.

**Tested on**
- [x] Windows
- [x] Linux
- [ ] macOS

I have:

Ping: @pez, @kstehn


cfehse:
@PEZ, @kstehn

Added some information to the description. This should be ready for testing.

cfehse:
Ping: @PEZ, @kstehn

Just to update you. There are some more timing problems during jack-in (e.g. slow systems starting the REPL processes) which need more testing. I will give you a ping when this is ready.  I will be afk untill monday. So this will have to wait. *gg* Nice holiday tomorrow Kevin and see you both next week.

cfehse:
I trash this PR and start a new one with the same changes based on the new dev source tree,

PR #362, opened 2019-09-30
[WIP] Move Clojure Warrior into Calva

PEZ:
This is just started.

When it's done it will have Introduced the following change(s):
- [x] Moved Clojure Warrior source code into Calva
- [x] Made Calva contribute what CW previously contributed (settings, commands, etcetera)
- [x] Made Calva activate the CW features such as rainbows and comment dimming/highlighting

I have:

- [x] Updated the `README.md`

Ping: @cfehse , @kstehn


PEZ:
Seems like the first thing we need to do is make it compile. :smile:

PEZ:
This is ready for merge afaict, @kstehn and @cfehse . What do you say?

cfehse:
@PEZ I take a look at this tomorrow, okay?

cfehse:
@PEZ  If I checkout the branch, build and run the build in debug - the (comment...) form are no longer dimmed. Original Clojure Warrior Extension is not installed. I changed nothing in my settings. 

image

cfehse:
@PEZ Wouldn't it be better to really integrate the code into the calva namespace. Renaming everything from "Clojure Warrior" into "Calva Warrior" or better to "Calva <whatever it's called, what the extension actually does>"? Now it is a bit of a mishmash to me.

cfehse:
@PEZ @kstehn 

Perhaps we should rearrange the source directory as well to something like that:
`src`
`src/calva`
`src/cljs-lib`
`src/calva-warrior`
and then get rid of the `src`/`'ts` directories inside the modules as well.

cfehse:
@PEZ @kstehn
`calva-fmt` and `calva-warrior` should really be one module, shouldn't they?

kstehn:
Yeah i like the idea of the rearrange 
there are some things that need some restructering in calva itself

PEZ:
> the (comment...) form are no longer dimmed

That's because I changed that so that tehy are now italicized instead. :smile: Ignore forms are still dimmed.

PEZ:
Regarding naming, I'm pretty happy with calling it Clojure Warrior. I don't want to change the settings names, so then it makes sense to call the module something similar to what the settings say.

Regarding rearranging, I am all for that, but it is a different issue than this PR, I'd say.

cfehse:
>Regarding naming, I'm pretty happy with calling it Clojure Warrior. I don't want to change the settings names, so then it makes sense to call the module something similar to what the settings say.

But how does a user now, that these "strange" `Clojure Warrior` settings belong to the Calva extension? In my experience with users things should be named clearly and comprehensibly. Why not name this `Calva Code Formater` (`Clava-fmt`) and `Calva Rainbow Brackets` (`Clojure Warrior`). I personally never heard of `Clojure Warrior` before I installed Calva the first time.  And what about old settings with the same name left over from an installation of the "old original" extension - namespace clash? - or does vscode remove all this entirely?

> Regarding rearranging, I am all for that, but it is a different issue than this PR, I'd say.

Yes but it should be done not too long after the merge I think.

cfehse:
> That's because I changed that so that they are now italicized instead. 😄 Ignore forms are still dimmed.

Thats good! I - and my eyes - like that much better. *gg*

cfehse:
@PEZ 

There are some unused imports and linter warnings (strangely the `import * as isEqual from 'lodash.isequal';` is even shown as an error).

image

PEZ:
> import * as isEqual from 'lodash.isequal'; is even shown as an error

`npm i`?

cfehse:
> npm i? 

No - I tried that first. The npm page for `lodash.isequal` recommended the `require('lodash.isequal')` notation as well. I have no idea why. *gg*

cfehse:
@PEZ  @kstehn This is ready to merge, isn't it?

kstehn:
At a glance seems to be fine to me :)

PEZ:
I decided that it was better to change the name now, rather than later. It is now called ”Calva Highlight”, since that is closest to what it provides. We should probably move the grammar things to this module/subdirectory as well, but that is for later.

In order to not just abandon the users with silently non-functioning settings, I have added a detector for oldly named settings and flash a warning message when those are detected. The message includes information on what the user should now do. To test this, add some settings that are valid for `calva.highlight`, but prefix with `clojureWarrior` instead.

PEZ:
Merged it since you guys were OK with it and I just change things according to previous feedback. Please be alert on any problems that could be related to this change when you have pulled `dev`.

Issue #368, opened 2019-10-02
[Build] The extension should specify the default schemes for document selectors


cfehse:
The extension schould specify the default schemes `file` and `untitled` for all document selectors together with the `language: "clojure"` attribute. This prevents a warning message in vscode and conforms to the standard described here [document selectors](https://code.visualstudio.com/api/references/document-selector).

cfehse:
This is fixed in version 2.0.47.

Issue #367, opened 2019-10-02
[Enhancement] Make code more robust in case Jack-in task fails.


cfehse:
The extension should behave more robust, if the Jack-in task fails either initially (e.g. the user selected an invalid project configuration like `Leiningen + shadow-cljs` if the project is only a Leiningen project) or the REPL process ends prematurely (e.g. a Leiningen project is only used for documentation generation and starting a REPL in the project will fail). 
The extension should always return to a usable state - not leaving die UI in the `connecting` state with the `nREPL` button not usable.

PEZ:
Agree! Maybe this too:

The user should be able to abort the jack-in process at any prompt by pressing <kbd>ESC</kbd>. Today, for some of the prompts, Calva doesn't ”listen” to that and just continues. With bad results, since it doesn't have the answer from the prompt.

If this should be in an issue of its own, let me know and I'll file a new issue instead.

cfehse:
@PEZ

Good point! For the first step in the connection sequence I corrected this already. But I will take a look at the following steps as well.

cfehse:
This is fixed in version 2.0.47.

Issue #366, opened 2019-10-02
[Enhancement] Support Jack-in without file open for single-rooted workspace


cfehse:
The extension should support the a Jack-in process to be started without an open file in a single-rooted workspace environment (e.g. opened via `Open Folder...`). Under these conditions the project file can be reliably detected even without an open file. 
This can not be done in an multi-rooted workspace - at least not until vscode provides an API to get any selected folder in the explorer view. Here a file must be opened farther to be able to detect the root folder of the project.

cfehse:
This is fixed in version 2.0.47.

PR #375, opened 2019-10-04
General stability fixes and arglist support in completion hover

### LINKED ISSUE #238: [Feature Request] show argument list of fn

cfehse:
Fixes #238,  #366,  #367,  #368

- Added the argument list to `HoverProvider` and `CalvaCompletionItemProvider`.
- Added a mechanism to create the namespace from a file in the REPL if that namespace does not exist in the REPL. This is necessary because if the namespace with which the nrepl command 'info' is called (the namespace of the file) does not exists no data is returned.
- Added activation events `workspaceContains:**/project.clj`, `workspaceContains:**/shadow-cljs.edn` and `workspaceContains:**/deps.edn`  to allow the extension to be activated faster. 
- Improved the `state.initProjectDir()` function to use `vscode.workspace.workspaceFolders![0]` as a second starting point in case no file is open in the workspace is not multi-root. This conforms to our [wiki entry](https://github.com/BetterThanTomorrow/calva/wiki/Workspace-Layouts) about the supported workspace layout. Made the function robust if no document is defined.
- Made the code robust in case a jack-in task exists prematurely an no REPL client is connected.
- Catch exception during `fs.watch()` in `executeJackInTask()`. These exception is according to the `fs` team by design if a given directory to watch does not exists or the underlying filesystem (network filesystems for example) does not support the operation. During the jack-in process this may occur if the user selects the wrong type of project - e.g. `Leiningen + shadow-cljs` if there is no shadow support in the project. In case the exception is detected, the jack-in task is canceled.
- Added the `file` and `untitled` schemes to all document selectors. These are the vscode supported schemes. The addition prevents the vscode warning message ` Extension '<extension>' uses a document selector without scheme. Learn more about...`.

**Tested on**
- [x] Windows
- [x] Linux
- [ ] macOS

I have:

Ping: @pez, @kstehn


cfehse:
@PEZ This is ready for testing/merging.

PEZ:
> Added a mechanism to load a file into the REPL if the namespace of the file does not exist in the REPL. This is necessary because if the namespace with which the nrepl command 'info' is called (the namespace of the file) no data is returned. So the file must be loaded to retrieve data. Another advantage of this approach is that if the file was loaded information of the elements contained in the file can be displayed as well.

This is a generally good idea. There are many first time experiences with Calva that get a bit strange because the namespaces are not loaded. However, I think this needs to be a setting. Much as I try to avoid settings, there are quite a few users who have opposed to make namespace loading the default behaviour. They have workflows where their namespaces contain code that can take very long to run, and will be hit badly if Calva runs the code just because a file is opened. I think this setting should default to loading the namespaces.

PEZ:
> Added activation events `workspaceContains:**/project.clj` and `workspaceContains:**/shadow-cljs.edn` to allow the extension to be activated faster.

Can I haz `deps.edn` added too?

Maybe `**/*.{clj,edn}*` would work, even...

cfehse:
@PEZ
 If not loading the file then perhaps just evaluate an "empty file" containing the correct namespace to create the namespace in the repl. Without the namespace known to the repl there will be no information even about the core clojure elements.

Another option is to make the info call in the user namespace if the namespace of the open file does not exist in the repl.

PEZ:
Tested the autoload namespace thing now, and love it.

However, in a `.cljc` file, it only seems to do it for the `clj` repl. If I toggle the file to be backed by the `cljs` REPL, then I need to manually load the file. (We can add this as a separate issue if it looks tricky to solve now.)

PEZ:
I agree that we might do better than just skip loading anything. I still suggest that for now we add a setting that defaults to true and if it is false then the user has to load the file, just as today. We can add an issue for refining it. But it is your call, since your PR. :smile:

cfehse:
@PEZ

Take a good look at all the changes so I can take care of the PR on Monday.

PEZ:
Tested:
* [X] Hover. Works like a charm.
* [ ] Autoloading file. See above comments.
* [ ] Activation events, see above comments. (Can be shipped as is, but a bit strange to leave tools.deps users hanging. :smile: 
* [x] Jack-in in single-root and multi-root namespaces. I look forward to updating the wiki on this!

src/providers/hover.ts
```diff
@@ -1,47 +1,60 @@
...
+                if (res.ns) {
+                    return new vscode.Hover(this.formatDocString(res.ns + "/" + res.name, res["arglists-str"] || [], "No documentation available"));
+                } else {
+                    return new vscode.Hover(this.formatDocString(text, "", "No information available"));
+                }
+            }
+            return new vscode.Hover(this.formatDocString(text, "", "No information available"));
         } else
-            return new vscode.Hover("Not connected to nREPL..");
+            return new vscode.Hover("Please connect to a REPL to retrieve information.");
```
PEZ: Noticed this while testing. Very good! Might seem like a small detail, but it is so much better to get information on what to do, rather than just that something is wrong.

PEZ:
I've skimmed the code changes now. Looks good to me. What do you say, @kstehn ?

src/nrepl/jack-in.ts
```diff
@@ -147,26 +175,43 @@ export async function calvaJackIn() {
...
     vscode.window.showInformationMessage("Not connected to a REPL server");
 }
 
 export async function calvaJackInOrConnect() {
 
     let commands = {};
-    if (!state.deref().get('connected')) {
+    if (!utilities.getConnectedState() &&
+        !utilities.getConnectedState() &&
+        !utilities.getLaunchingState()) {
```
kstehn: There seems to be the same statement two times here `!utilities.getConnectedState()`

src/nrepl/jack-in.ts
```diff
@@ -147,26 +175,43 @@ export async function calvaJackIn() {
...
         // if not connected add the connect commands
         commands["Start a REPL server and connect (a.k.a. Jack-in)"] = "calva.jackIn";
         commands["Connect to a running REPL server in your project"] = "calva.connect";
         commands["Connect to a running REPL server, not in your project"] = "calva.connectNonProjectREPL";
     } else {
-        // if connected add the disconnect command and the
+        // if connected add the disconnect command and the 
         // REPL window open commands if needed.
-        commands["Disconnect from the REPL server"] = "calva.disconnect";
+        commands["Disonnect from the REPL server"] = "calva.disconnect";
```
kstehn: That should be `Disconnect` or not?

src/providers/hover.ts
```diff
@@ -1,47 +1,60 @@
...
-            let res = await client.info(ns, text);
-            if(res.doc)
-                return new vscode.Hover(this.formatDocString(res.ns+"/"+res.name, res["arglists-str"] || [], res.doc))
-            if(res.ns)
-                return new vscode.Hover(this.formatDocString(res.ns+"/"+res.name, res["arglists-str"] || [], "No documentation available"));
+            let client = util.getSession(util.getFileType(document));
+            if(client) {
+                await util.loadFileIfNamespaceNotExist(document);
+                let res = await client.info(ns, text);
+                // I really don't not why result.arglists-str does
```
kstehn: The reason for that is the `-`

src/providers/hover.ts
```diff
@@ -1,47 +1,60 @@
...
+                .replace(/\]/g, ']')
+                .replace(/\[/g, '* [');
+            result += '\n\n';
             result += '\n\n';
         }
-
         // Format the actual docstring
-        if (doc) {
-            result += doc.replace(/\s\s+/g, ' ');
+        if (documentation && documentation != "") {
```
slipset: Jumping in here, so please disregard.
This is basically the same as https://github.com/BetterThanTomorrow/calva/pull/375/files#diff-9205a2ccff45e74e74a392042d83efe8R25 Could/should it be extracted into a `docUtils` (or some such, naming is hard)?

src/providers/hover.ts
```diff
@@ -1,47 +1,60 @@
...
         }
         return result.length > 0 ? result : "";
     }
 
     async provideHover(document, position, _) {
-        let text = util.getWordAtPosition(document, position);
-        if (this.state.deref().get('connected')) {
-            let client = util.getSession(util.getFileType(document));
+
+        if (util.getConnectedState()) {
```
slipset: My knowledge of typescript is limited, but I'd suggest something like:
```
// in-ns util
function with-connection(f, ...) {
  function (args) {
     if (util.getConnectedState() {
       return  f(args);
    } else {
     return "Please connect to a REPL to retrieve information."
  }
}
```

slipset: Also, I'd recommend decoupling this from the `hover` functionality itself.
It seems as `provideHover` "bottoms out" in 
`return new vscode.Hover(someString)`
So I'd suggest:

```
calculateHoverString(document, position) {
 const text = util.getWordAtPosition(document, position);
  const  ns = util.getNamespace(document);

  const client = util.getSession(util.getFileType(document));
  if(client) {
                 await util.loadFileIfNamespaceNotExist(document);
                 let res = await client.info(ns, text);
                 // I really don't not why result.arglists-str does 
                 // not work, but it leads to an compiler error.
                 if (res.ns && res.doc) {
                     return this.formatDocString(res.ns + "/" + res.name, res["arglists-str"] || [], res.doc)
                 }
                 if (res.ns) {
                     return this.formatDocString(res.ns + "/" + res.name, res["arglists-str"] || [], "No documentation available");
                 } else {
                     return this.formatDocString(text, "", "No information available");
                 }
             }
             return this.formatDocString(text, "", "No information available");
}

async provideHover(document, position, _) {
        const message = util.withConnection(calculateHoverString(document, position);
       return new score.Hover(message);
}
```

If you go in this direction, you could potentially unit-test `calculateHoverString`

src/utilities.ts
```diff
@@ -190,6 +190,29 @@ function getWordAtPosition(document, position) {
     return text;
 }
 
+async function loadFileIfNamespaceNotExist(doc) {
+
+    if (getConnectedState()) {
```
slipset: I'd suggest writing this in a guard-clause fashion, and also to possibly throw, since I would assume that it's an error to try loading a ns that we cannot load. Also, by throwing exceptions, we can inform the caller of why we didn't succeed:

```typescript
async function loadFileIfNamespaceDoesNotExist(doc) {
if (!getConnectedState()) {
   throw new Error("Not connected");
}
cont document = getDocument(doc);
if (!document) {
  throw new Error("Couldn't find " + doc)
}
const client = getSession(getFileType(document));
if (!client) {
  throw new Error("No client available for " + document);
}
const ns = getNamespace(document);
const name = getFileName(document);
const dir = document.hasOwnProperty('fileName') ? path.dirname(document.fileName) : ''; 
const nsList = await client.listNamespaces([]);
if (nsList['ns-list'] && nsList['ns-list'].includes(ns)) {
        return;
}
await client.loadFile(document.getText(), name, dir);
}
```

src/utilities.ts
```diff
@@ -237,8 +260,22 @@ function getSession(fileType = undefined): NReplSession {
     }
 }
 
-function getConnectedState() {
-    return state.cursor.get('connected');
+function getLaunchingState() { 
+    return state.deref().get('launching');
+}
+
+function setLaunchingState(value) {
```
slipset: This is 
```typescript
function setLaunchingState(value) {
        vscode.commands.executeCommand("setContext", "calva:launching", Boolean(value));
         state.cursor.set('launching', value);
}
```

PEZ:
Many thanks for chipping in, @slipset !

src/utilities.ts
cfehse: @slipset Good point!

src/providers/hover.ts
cfehse: @kstehn Okay. *gg*

src/nrepl/jack-in.ts
cfehse: @kstehn Corrected that.

cfehse:
@PEZ  @kstehn 

I change the mechanism to enable the info data to be retrieved (autoload namespace thing): if the namespace does not exists in the REPL it is now simply created by evaluating a form like `(ns <namespace>)`. That is enough to get info data from the REPL for everything outside the file. If the user wants to have all information for the file, the file must be loaded. This avoids an option and does not interfere with the workflow of the users.

>However, in a .cljc file, it only seems to do it for the clj repl. If I toggle the file to be backed by the cljs REPL, then I need to manually load the file. (We can add this as a separate issue if it looks tricky to solve now.)

I choose the corresponding REPL for the document using `util.getSession(util.getFileType(document))` so the `info` call goes to the right REPL. If there is an issue I think it is not related to this PR.

> Can I haz deps.edn added too?

Yes you can. *gg*

> Maybe **/*.{clj,edn}* would work, even...

I would not do this because I find it more robust to surely have any supported project file in the workspace to activate the extension.

@slipset 

I incorporated some of your suggestions - thanks for that. Refactor out some code into a new helper unit and decouple some of the hover and completion code I would like to be done in another PR. This is large enough.

src/utilities.ts
```diff
@@ -190,24 +190,19 @@ function getWordAtPosition(document, position) {
     return text;
 }
 
-async function loadFileIfNamespaceNotExist(doc) {
+async function CreateNamespaceFromDocumentIfNotExists(doc) {
```
PEZ: We name functions using `camelCase`, in our undocumented code convention. 😄

cfehse: Good point! Changed...

Issue #228, opened 2019-06-17
REPL window gets sluggish with large data structures


PEZ:
When the repl window is handed a somewhat large dataset, it gets really slow and sluggish.

Reproduce by either evaluating something like:

```clojure
(map (fn [x] {x (range 7)}) (range 1000))
```

You can also paste a large data structure. For instance copy the last result after evaluationg the above, `ctrl+alt+c ctrl+c`. And then paste it.

It takes a too long while for the results to appear.

This could probably partially be handled by coming enhancements to `nrepl`: https://github.com/nrepl/nrepl/pull/106

But it would still leave the case with pasting large chunks of data.

Gnurdle:
perhaps give some thought to something akin to 'cider-inspect' which would be useful in its own right.

PEZ:
I am not familiar with cider-inspect. Can you describe it some, and how you think it would help here?

I started to think about this problem yesterday before falling asleep. My ideas in that stage are mostly crap, but maybe this one could be an interim solution until someone finds out what would bring the REPL window the performance it needs:
1. Detect when the output is too large for comfort.
2. Print only part of the output, and add a button to click at to see the rest. The rest could then open in a regular editor that we enable Clojure mode in.

Obviously needs some more thought around the UI/Ux, but printing results to a regular Clojure editor should be handy for many situations.

brunchboy:
cider-inspect allows you to interactively traverse Clojure and Java data structures, paging through lists and vectors, and diving into and out of objects. It can even handle structures with cycles in them, which would never finish printing at the REPL, because the user guides the search through the object space, and nodes are only expanded as they are reached. I came to this issue trying to determine if Calva had a similar feature, to see if I could recommend it as an alternative to my users (because some of my data structures do have cycles), but so far it is looking like CIDER is the only option.

PEZ:
Thanks, @brunchboy. It sound a bit like REBL, have I understood it correctly if I think like that?

brunchboy:
Hmm, I had not heard of REBL before, but it does look similar, except that cider-inspect works over the nREPL server so you can do it to a remote process (which is the use case that I need, because my users are running my Beat Link Trigger as a standalone Jar file or native application with embedded Java runtime).

PEZ:
[...]

PR #381, opened 2019-10-06
[WIP] Inspect data structures

PEZ:
Console logs out a type annotated object of the evaluation results.
Addresses #228

This is very, very much just work in progress and a request for feedback.

I want to add an inspector to Calva, to make it so that I don't have to use the rather clunky `clojure.inspector/inspect-tree` to navigate data structures when they get a bit large.

My idea is to use a VS Code Tree View for this and populate it with the parsed result. I'm thinking this could happen as automatically as the inline result decorations and hovers. And also be cleared together with those, probably with a way to pin results so that they can survive clearing of the decorations.

For now I have only started to try see what shape the data structure should have in order to be easy to traverse and use to populate the tree view. And right now all I do with it is `console.log()` it out when the evaluation results come in. To try it out, evaluate some code and check the Developer Tools Console.

I'd like to know if you think I am going about this the right way. And if so also what types I should annotate, I just took a few from the top of my head for now.

The interesting stuff happens in `inspector.cljs`. 

I have:

Ping: @slipset, @kstehn, @cfehse, @Gnurdle

### DIFF
--- a/src/cljs-lib/src/calva/inspector.cljs
+++ b/src/cljs-lib/src/calva/inspector.cljs
@@ -2,10 +2,11 @@
   "Utilities for supporting Calva's inspection of data structures."
   (:require [calva.js-utils :refer [jsify parse-edn]]))
 
-(defn- inspector-annotate
-  "Recursively annotates ClojureScript values with .
-  sets/vectors/lists become Arrays, Keywords and Symbol become Strings,
-  Maps become Objects. Arbitrary keys are encoded to by `inspector-key->js`."
+(defn- annotate
+  "Recursively annotates ClojureScript values with Clojure data type information.
+  Sets/vectors/lists become arrays, keywords and symbol become strings. Maps become
+  objects. All these are wrapped in objects containing `type` and `value`.
+  Values that map to JavaScript primitives are not annotated."
   [x]
   (letfn [(thisfn [x]
             (cond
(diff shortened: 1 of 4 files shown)


src/cljs-lib/src/calva/inspector.cljs
```diff
@@ -0,0 +1,43 @@
...
+  "Utilities for supporting Calva's inspection of data structures."
+  (:require [calva.js-utils :refer [jsify parse-edn]]))
+
+(defn- annotate
+  "Recursively annotates ClojureScript values with Clojure data type information.
+  Sets/vectors/lists become arrays, keywords and symbol become strings. Maps become
+  objects. All these are wrapped in objects containing `type` and `value`.
+  Values that map to JavaScript primitives are not annotated."
+  [x]
+  (letfn [(thisfn [x]
```
slipset: Hmm, I think you can avoid the letfn and just call annotate recursively directly
```
(defn- annotate [x]
(cond
               (keyword? x) {:type "keyword" :value (name x)}
               (symbol? x) {:type "symbol" :value (str x)}
               (map? x) {:type "map" :value (into {} (for [[k v] x] [k (thisfn v)]))}
               (coll? x) (let [type (cond
                                      (vector? x) "vector"
                                      (list? x) "list"
                                      (set? x) "set"
                                      :else "coll")]
                           {:type type :value (map annotate x)})
               :else x))
```

You could also solve this by using a multi method like 
```
(defmulti annotate type) 

(defmethod annotate clojure.lang.Symbol [x] {:type "symbol" :value (str x)})
```

You an probably also do this with one of the clojure walk functions https://clojuredocs.org/clojure.walk/postwalk

Anyhow, you should probably try to (and I'm on deep water here) write it in a tail-recursive manner so you can use `recur` instead of calling `annotate` recursively to avoid blowing the stack (stack overflow)

src/cljs-lib/src/calva/inspector.cljs
```diff
@@ -0,0 +1,43 @@
+(ns  calva.inspector
+  "Utilities for supporting Calva's inspection of data structures."
+  (:require [calva.js-utils :refer [jsify parse-edn]]))
+
+(defn- annotate
```
slipset: I'd make this fn public and write a unit-test or two for it, because it's pure and all.

src/cljs-lib/src/calva/inspector.cljs
PEZ: So many options! I'll have a think. Implementing this I took inspiration/stole from `clj->js` which to my understanding has a similar task to perform on similar data.

bpringe:
@PEZ I just noticed this PR is related to the work I am doing with showing structured values in the variables pane during debugging. I'm using orchard.inspect there, and I wonder if that would be of use here too. https://cljdoc.org/d/cider/orchard/0.5.0-SNAPSHOT/api/orchard.inspect

PEZ:
If we return to this I think it needs fresh think anyway.

PR #388, opened 2019-10-10
Fix issue 374

### LINKED ISSUE #374: calva.evalCurrentTopLevelFormInREPLWindow fails in specific condition
If I run `ctrl+alt+c, ctrl+alt+space` with my cursor after the last form inside a comment form like this, the command fails to run, with a little notification saying it failed. It seems to do this only if I have another form above it inside the comment form. By itself, it works.

image

image

cfehse:
Fixes #374.

- `select.ts` function `_adjustRangeIgnoringComment()` changed regex for `postMatch` to find the last ` )`.

I have:

Ping: @pez, @kstehn, @cfehse

### DIFF
--- a/src/select.ts
+++ b/src/select.ts
@@ -13,7 +13,7 @@ function _adjustRangeIgnoringComment(doc, range) {
             postTextLength = 0,
             // find the beginning '(comment...`in the text.
             preMatch = text.match(/^\(\s*comment\s+/m),
-            // find only the last ')' in the text.
+            // find the last ')' in the text.
             postMatch = text.match(/(\))(?!.*\))/m);
         if (preMatch) {
             preTextLength = preMatch[0].length;
--- a/src/select.ts
+++ b/src/select.ts
@@ -22,22 +22,24 @@ function _adjustRangeIgnoringComment(doc, range) {
             postTextLength = postMatch[0].length;
         }
         start += preTextLength;
-        end -= postTextLength - 1;
+        end -= postTextLength;
         return new vscode.Range(doc.positionAt(start), doc.positionAt(end));
     } else {
         return range;
     }
 }
 
-function _getFormSelection(doc, pos, topLevel, ignoreComment = true): vscode.Selection {
+function _getFormSelection(doc, pos: vscode.Position, topLevel): vscode.Selection {
     let allText = doc.getText(),
         ast = paredit.parse(allText),
         idx = doc.offsetAt(pos),
         peRange = topLevel ? paredit.navigator.rangeForDefun(ast, idx) : paredit.navigator.sexpRange(ast, idx);
     if (peRange) {
         let range = new vscode.Selection(doc.positionAt(peRange[0]), doc.positionAt(peRange[1]));
-        if (ignoreComment) {
+        let rangeText = doc.getText(range);
+        if (pos.isAfter(range.start) && pos.isBefore(range.end)) {
             range = _adjustRangeIgnoringComment(doc, range);
+            rangeText = doc.getText(range);
             if (topLevel) {
                 const idxOffset = doc.offsetAt(range.start);
                 ast = paredit.parse(doc.getText(range));
@@ -59,7 +61,7 @@ function _selectCurrentForm(document = {}) {
         selection = editor.selection;
 
     if (selection.isEmpty) {
-        let codeSelection = _getFormSelection(doc, selection.active, false, false);
+        let codeSelection = _getFormSelection(doc, selection.active, false);
         if (codeSelection) {
             editor.selection = codeSelection;
         }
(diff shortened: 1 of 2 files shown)


PEZ:
I think the code should be simplified instead. As we discussed offline. Here is my suggestion:

```typescript
import * as vscode from 'vscode';
import * as util from './utilities';
const paredit = require('paredit.js');


function _adjustTopLevelRangeIgnoringComment(doc, range) {
    const text = doc.getText(range);
    if (text.match(/^\(\s*comment\s+/m)) {
        const start = doc.offsetAt(range.start),
            end = doc.offsetAt(range.end);
        return new vscode.Range(doc.positionAt(start - 1), doc.positionAt(end - 1));
    } else {
        return range;
    }
}

function getFormSelection(doc, pos, topLevel): vscode.Selection {
    let allText = doc.getText(),
        ast = paredit.parse(allText),
        idx = doc.offsetAt(pos),
        peRange = topLevel ? paredit.navigator.rangeForDefun(ast, idx) : paredit.navigator.sexpRange(ast, idx);
    if (peRange) {
        let range = new vscode.Selection(doc.positionAt(peRange[0]), doc.positionAt(peRange[1]));
        if (topLevel) {
            range = _adjustTopLevelRangeIgnoringComment(doc, range);
            const idxOffset = doc.offsetAt(range.start);
            ast = paredit.parse(doc.getText(range));
            idx = idx - idxOffset;
            peRange = paredit.navigator.rangeForDefun(ast, idx);
            range = new vscode.Selection(doc.positionAt(peRange[0] + idxOffset), doc.positionAt(peRange[1] + idxOffset));
        }
        return range;
    }
    else {
        return new vscode.Selection(pos, pos);
    }
}

function selectCurrentForm(document = {}) {
    let editor = vscode.window.activeTextEditor,
        doc = util.getDocument(document),
        selection = editor.selection;

    if (selection.isEmpty) {
        let codeSelection = getFormSelection(doc, selection.active, false, false);
        if (codeSelection) {
            editor.selection = codeSelection;
        }
    }
}

export default {
    getFormSelection: getFormSelection,
    selectCurrentForm: selectCurrentForm
};
```
This fixes another bug as well because we are currently adjusting for the comment form even if we are not doing a top-level eval.

cfehse:
@PEZ 

< This fixes another bug as well because we are currently adjusting for the comment form even if we are not doing a top-level eval.

With my proposed change I do not get that error.

PEZ:
My suggested code doesn't really work, so let's not use that for now. :smile:

PEZ:
> This fixes another bug as well because we are currently adjusting for the comment form even if we are not doing a top-level eval.

The bug I'm talking about is because on line 40 we adjust for comment form before we check for topLevel. It's an ancient bug, but while at it, and all that. :smile:

PEZ:
Also, for `_getFormSelection`, that `ignoreComment = true` argument. There is no case where we would set it to false when `topLevel` is true, so it should just be removed.

src/select.ts
```diff
@@ -22,22 +22,24 @@ function _adjustRangeIgnoringComment(doc, range) {
...
-function _getFormSelection(doc, pos, topLevel, ignoreComment = true): vscode.Selection {
+function _getFormSelection(doc, pos: vscode.Position, topLevel): vscode.Selection {
     let allText = doc.getText(),
         ast = paredit.parse(allText),
         idx = doc.offsetAt(pos),
         peRange = topLevel ? paredit.navigator.rangeForDefun(ast, idx) : paredit.navigator.sexpRange(ast, idx);
     if (peRange) {
         let range = new vscode.Selection(doc.positionAt(peRange[0]), doc.positionAt(peRange[1]));
-        if (ignoreComment) {
+        let rangeText = doc.getText(range);
```
PEZ: `rangeText` is assigned twice, but never used.

PEZ:
I still think we do a lot of hard-to-read work for finding the first and the last character of the range. This does the job for all the uses this function has:

```typescript
function _adjustRangeIgnoringComment(doc, range) {
    let text = doc.getText(range);
    if (text.match(/^\(\s*comment\s+/m)) {
        // Create a new top level context by removing the first and last characters of the range.
        // These will always be the opening and the closing parens of the `(comment ...)` form.
        const start = doc.offsetAt(range.start),
            end = doc.offsetAt(range.end);
        return new vscode.Range(doc.positionAt(start + 1), doc.positionAt(end - 1));
    } else {
        return range;
    }
}
```
Much easier to read and understand what is going on.

cfehse:
> I still think we do a lot of hard-to-read work for finding the first and the last character of the range. Much easier to read and understand what is going on.
(The reason my code in a previous comment bugged out was that I was using start - 1 instead of start + 1, so growing the selection in that end instead of shrinking. The workings of a tired brain...)

Your code behaves differently. The original corrected code does not only removes the first and last character (will say the opening and closing parentheses). The original code removes the complete `comment` block leaving only the content of the block in the range. This gets different result under some circumstances.

cfehse:
@PEZ 

I changed the code as advised. Please test it.

PEZ:
Two bugs less!

Issue #394, opened 2019-10-13
[REPL Window] Completion in REPL window should work like in the editor 


cfehse:
In the vscode editor a selection from the completion editor can be made with the `tab` or `return` key. It the selection is made the completion overlay is closed. The completion in the REPL window should behave the same way. Currently only the `tab` key selects an entry from the list and the overlay is not closed and must be closed with the `esc` key.

PR #395, opened 2019-10-13
Improve paredit in REPL window

### LINKED ISSUE #255: Overly aggressive paredit in REPL window
If I type `[1 2 3]` in the REPL window and then decide that I want it to be `(shuffle [1 2 3])` then if I click at the beginning and type `(` then it adds `)` right after it automatically, which is the sort of thing I've come to expect both in Calva and in other environments... *but* unlike when this happens in other contexts there appears to be no way to delete the `)` no matter what I do.

I know that I'm an outlier in not personally liking any automatic typing of closing brackets ever, and also in not wanting to learn/use Paredit commands (I prefer ordinary typing), and I don't expect the behavior of Calva to change in response to my preferences! But the middle ground adopted in Calva editor windows -- adding closing brackets but letting me delete them -- works well enough for me, and presumably also for others.

So my request is that the REPL window acts like the editor windows in this respect.

cfehse:
## What has Changed?

- Edit message/event mechanism to populate the paredit keymap setting into the REPL webview.
- Made the Paredit implemention in the webview obey the keymap setting and behave like the editor according to this setting.
- Change the completion overlay in the webview to allow the selection of an item using the `tab` or `return` key and close the overlay after the selection (behavior like in the vscode editor)
- Changed the statusbar item for the paredit to toggle between the `strict` and `original` keymap. If the keymap is set to `none` the item is disabled.

Fixes #255, #394 

## My Calva PR Checklist

I have:

Ping @pez, @kstehn, @cfehse


cfehse:
@PEZ 

I have tried to restrict the commands in the REPL implementation of Paredit (`src/webview/repl-console.ts`). I don't know if I did it right. Please take a close look at all the changes.

Thanks!

PEZ:
Sweet! Will have a look at this.

PEZ:
It works like a charm! Both the paredit mode and the enter in completion overlay. That latter thing is saving my sanity, I notice. I so often forget that it doesn't work to use enter, and then strange things happens instead.

I'm pleased that you choose to use the toggle button for strict mode. :heart: Speaking of that button... :smile: : 

1. When the keymap is set to `none`, the button doesn't work. Maybe that is ok. But it says that Paredit is disabled, which is wrong. Maybe it should just say **Paredit keymap is set to `none`, toggling strict mode is disabled**
1. The button should show what mode it is using. So maybe it can be w/o parens in `original` mode and with in `strict`.
1. I want to be able to toggle this with a command, like we have with the `pprint` toggle. Maybe it should be called **Toggle Paredit Mode to `<the-other-mode>`** (Is it even possible to change the text? otherwise, just skip the ”to” part), default shortcut could be `ctrl+alt+c -`
1. I'd like to have the button to the left, together with the other Calva buttons.
    * I'd also like to have the `pprint` button there, but maybe that's OT for this PR :smile:

Will continue testing and try find time to read the source changes too.

cfehse:
>When the keymap is set to none, the button doesn't work. Maybe that is ok. But it says that Paredit is disabled, which is wrong. Maybe it should just say Paredit keymap is set to none, toggling strict mode is disabled

Done.

>The button should show what mode it is using. So maybe it can be w/o parens in original mode and with in strict.

Done.

>I want to be able to toggle this with a command, like we have with the pprint toggle. Maybe it should be called Toggle Paredit Mode to <the-other-mode> (Is it even possible to change the text? otherwise, just skip the ”to” part), default shortcut could be ctrl+alt+c -

Done without changing the title - which is not possible in the moment, I think.

>I'd like to have the button to the left, together with the other Calva buttons

Done for both buttons.

src/webview/repl-console.ts
```diff
@@ -84,6 +92,34 @@ export class ReplConsole {
         }
     }
 
+    private isKeyMay(values: ReplPareditKeyMap[]) :boolean {
```
PEZ: Typo, should be: `isKeyMap`, right?

src/webview/repl-console.ts
```diff
@@ -323,16 +359,20 @@ export class ReplConsole {
 
     commands = {
         "raise-sexp": () => {
-            this.readline.withUndo(() => {
-                paredit.raiseSexp(this.readline);
-                this.readline.repaint();
-            });
+            if (this.isKeyMay([ReplPareditKeyMap.ORIGINAL, ReplPareditKeyMap.STRICT])) {
```
PEZ: Why do we check this here? And, why only on some commands and not on others?

src/webview/repl-console.ts
```diff
@@ -481,10 +539,15 @@ export class ReplConsole {
...
-            this.readline.withUndo(() => {
-                paredit.backspace(this.readline);
+            if (this.isKeyMay([ReplPareditKeyMap.STRICT])) {
+                this.readline.withUndo(() => {
+                    paredit.backspace(this.readline);
+                    this.readline.repaint()
+                })
+            } else {
+                this.readline.backspace();
                 this.readline.repaint()
```
PEZ: How come we do not do this `withUndo`? (I tested it and I can undo also for normal mode, so this is probably correct. I am just wondering.)

src/repl-window.ts
```diff
@@ -93,9 +97,13 @@ class REPLWindow {
...
         })
 
+        paredit.onPareditKeyMapChanged((keymap) => {
+            this.postMessage({ type: "paredit-keymap", keymap: keymap });
+        })
+
         this.panel.onDidDispose((e) => {
             if (this.evaluation)
-                this.evaluation.interrupt();
+                this.evaluation.interrupt().catch(() => {});
```
PEZ: This is a reminder to us, as a team: Lots of the repl window code has these `bare` if statements. When we change stuff in those we should put the braces in, b/c that is part of the, (yet very secret)  code convention we have.

PEZ:
I think the code changes are good. (Put some comments here and there for details, but as a whole it makes sense.) Will test the latest changes and get back with my findings from that.

src/webview/repl-console.ts
cfehse: I tried to mimic the when clauses in the package.json. I put the check around all commands where paredit is involved. I have no idea if this makes sense.

src/webview/repl-console.ts
cfehse: Yes. *gg*

src/webview/repl-console.ts
cfehse: It comes because I forget to do so. No idea why it works. *gg*

src/repl-window.ts
cfehse: Good point. That specific change was to remove the linter warning.

src/webview/repl-console.ts
PEZ: I don't think it makes sense. The when clause is in effect also in the REPL window. Try it in 2.0.49, setting the keymap to `none` disables Paredit shortcuts in the REPL window.

cfehse: Okay what should I do then? Remove the check everywhere except backspace and delete?

cfehse: Or only at 'raise-sexp'? Why do the shortcuts from package.json work in the repl?

cfehse:
@PEZ 

I made the suggested changes from our discussion yesterday.

PR #412, opened 2019-10-19
Fix floating promises

### LINKED ISSUE #411: TS-linting warnings about floating promises
While I'm hacking Calva my ts-lister is showing warnings about floating promises on function calls of `evaluateSelection` in `src/evaluate.ts`.

Here is one of them
https://github.com/BetterThanTomorrow/calva/blob/9193fd9a09222174e3848ed4e2584cabba5856c3/src/evaluate.ts#L120

sauvala:
## What has Changed?

- The promise returned by `evaluateSelection` was not handled correctly. Now the function calls are wrapped with `async/await`.

Fixes #411 

## My Calva PR Checklist

I have:

Ping @pez, @kstehn, @cfehse


PEZ:
Hello, 

I'm super happy that you have started sending PRs!

The changes look good. But you have based them on a pretty old (in Calva time) version of `dev`. Can you update your branch from latest `dev`. I tried it and an everything merges cleanly, except `CHANGELOG.md`, but that should be really easy to fix manually.

In case merging upstream is new to you, here is what I would do, with your feature branch checked out:

```sh
git remote add upstream git@github.com:BetterThanTomorrow/calva.git
git fetch upstream
git merge upstream/dev
```

cfehse:
@kstehn  @PEZ  @sauvala 

A question for the Typescript fans and about tackling this kind of linter warnings in general. I am not a fan of changing the method/function signature to `asyn` and placing an `await` before the call in question because this may often not be what was intended and may lead to other warning in code calling the now `async` functions. The functions mentioned in this PR honestly do not need to be `async`. They only start the called functions of and exit (these are command functions). I fix similar warning in the following way:

``` 
function evaluateTopLevelForm(document = {}, options = {}) {
    evaluateSelection(document, Object.assign({}, options, { topLevel: true, pprint: state.config().pprint })).catch(() => {});
}
```
The reject handler should prevent the code from exception being throw because the rejection is not handled as well - the calling code does not care about any rejection anyway. 

What's your opinion about this?

cfehse:
And if you take a look at the `evaluateSelection()` function there are as well two `await` statements in that function because it is easier to write that code synchronously. This all lead to `async`s bubble up all the way in the calling sequence where it is not necessary I think.

PEZ:
I'm not knowledgeable enough about these promise warnings. I was surprised that async/await would silence them. Agree that if we don't need it to be async, for async reasons, we probably shouldn't make it look like it is needed. Can the warning be dealt with using `.catch()` on the promise instead?

cfehse:
@PEZ

Yes it can.

cfehse:
The warning is generally very useful but we often use promises in a "fire and forget" sense so we do not care about the results in the calling/creating code. In this situations the warning is not very useful I think.

PEZ:
I'm pretty fond of the warnings. 😄 

This spring I had serious struggles releasing Calva 2, much because of dangling promises. I was saved by a PR that enabled the warnings. It helped me chase the bugs down. I suggest we are explicit with the ”fire and forget” by adding no-op `.catch()` to them. Maybe even add a comment in them saying that we are not interested in handling whatever error.

cfehse:
And avoid as much `await`s as possible. Because they very often cause serious trouble - like in the case of "allow user input from the REPL window" solution.

sauvala:
@PEZ thanks for noticing that! I'll update my branch.

@cfehse I'm okay for changing the method signature back to non-async and adding the `.catch(() => {})` handler for it instead. Do you think we could also add logging inside the `catch`-statement to see what was the error? I'm not sure if it's really helpful on the VSCode. Does Calva users see those error's somewhere in the console and could report them?

> I am not a fan of changing the method/function signature to asyn and placing an await before the call in question because this may often not be what was intended and may lead to other warning in code calling the now async functions.

Those methods are called from places where  we can use async-functions. If you think there is going to be other use cases for these in the future where you cannot use async-functions, I do the change back to non-async. 🙂 

The `await` functions can also be wrapped inside `try-catch` to enable logging and error handling.

PEZ:
>  Do you think we could also add logging inside the catch-statement to see what was the error? I'm not sure if it's really helpful on the VSCode. Does Calva users see those error's somewhere in the console and could report them?

Yes, these are printed in the dev console, which the user can open and report to us about. We could make it a convention to do:
```
.catch(e => { console.warn(`Unhandled error: ${e.message}`) }`
```
or some such, in these cases.

sauvala:
Opened a new PR that has clean history

Issue #89, opened 2018-11-28
Red squigglies under compile errors


slipset:
In cider, when evaluating a form with an error, a red squiggly is drawn under the offending form:
![screenshot 2018-11-28 at 08 55 30](https://user-images.githubusercontent.com/5894926/49137203-0b43da80-f2ec-11e8-844c-02f553c80630.png)

I'd think you'd get all the info you need wrt `file` `line` and `col` from the compiler exception
This would make a great addition to Calva!

Also, note that the CompilerException has `ex-data` from version 1.10 which makes this quite a lot easier

PEZ:
Thanks! Of course it would be nice to implement this in a way that it works with older Clojure too, but solving it for only 1.10 would be worth it too.

mseddon:
The way `cider` and `figwheel` handle this sort of behaviour is to catch `clojure.lang/Compiler$CompilerException`  [see here](https://github.com/clojure/clojure/blob/master/src/jvm/clojure/lang/Compiler.java#L6814), which contains the source/column information needed.

sauvala:
Letting others also know in here (in additional to Slack) that I'm working on this.

cfehse:
This is implemented in version 2.0.49.

PR #413, opened 2019-10-19
[WIP/Discussion] Mark compilation errors on form evaluation

sauvala:
### PR is WIP and for Discussion

Hi Calva maintainers!

There was issue #89 where user wanted to have a red marking on compiler errors. @cfehse mentioned that this was implemented in version 2.0.49. 🎉 

However, I tried the feature but I didn't see what I was expecting. This might totally be just a my fault about not knowing how it should show up in the editor. 🤷‍♂

I had an idea how the error should be shown in the editor and implemented it here. I would like to get some feedback if it's a good idea. I added a screenshot how it shows in the editor when evaluating a form with a compiler error.

mark-compilation-error

### DIFF
--- a/src/evaluate.ts
+++ b/src/evaluate.ts
@@ -8,6 +8,7 @@ import * as util from './utilities';
 import { activeReplWindow } from './repl-window';
 import { NReplSession } from './nrepl';
 import statusbar from './statusbar'
+import { stripClass } from './utilities';
 
 function addAsComment(c: number, result: string, codeSelection: vscode.Selection, editor: vscode.TextEditor, selection: vscode.Selection) {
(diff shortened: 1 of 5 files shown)


PEZ:
Hello! This is an awesome suggestion. Please go ahead. We want it to work for ClojureScript as well. Will it?

In the code it looked like you replaced the decoration with the squiggles. I think I want both, but I am not sure until I've tested this some.

sauvala:
Good call! I need to test this with ClojureScript as well. I was too focused working on the Clojure side 😄

Yes, I'm replacing the normal error annotation in the case of `CompilerException`. I can also test how it looks like when we have this and the normal error annotation.

sauvala:
I'll close the PR because it was left untouched for a long time. Feel free to pick this up if there is not this feature yet and if someone wants to work on it.

PR #414, opened 2019-10-19
[WIP] Pretty printing does only that (not also expanding the result)

### LINKED ISSUE #378: pretty printing in repl doesn't emit reader literals
follows is the difference between pretty and non-pretty repl output
clj.user=> 
(ct/now)
#<org.joda.time.DateTime@71436076 2019-10-06T14:13:50.287Z>
clj.user=> 
(ct/now)
#clj-time/date-time "2019-10-06T14:14:22.977Z"
clj.user=>

This is not a huge(tm) deal, but it is a difference.

The particular case that annoys is trying to paste output back into a 'def' etc.

PEZ:
## What has Changed?
- Pretty printing is now performed using `calva-lib` on the plain-printed results from nrepl. Instead of doing it via nrepl.
- Since we need to first parse the result string in order to prettify it, there are situations when the string doesn't parse. (E.g. Datomic `:db-before` values.) In these cases we print the result plain instead and show the parsing error causing the pprint to not happen. 
- This way of doing the pprint also means data readers are left alone. Which should fix #378 
- This also avoids having values ”expanded” by pretty printing, which can get pretty hairy, as #415 shows.

To test the change, you can use this project: https://github.com/PEZ/nrepl-pprint-datomic
See #415 for the repro steps.

Fixes #378 #415

## My Calva PR Checklist
I have:

Ping @pez, @kstehn, @cfehse

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 Changes to Calva.
 
 ## [Unreleased]
+- [Fix Pretty printing Datomic transaction results, dumps the whole database, twice](https://github.com/BetterThanTomorrow/calva/issues/415)
+- [Fix pretty printing in repl doesn't emit reader literals](https://github.com/BetterThanTomorrow/calva/issues/378)
 
 ## [2.0.52] - 10.19.2019
 - [Add info box for VIM Extension users](https://github.com/BetterThanTomorrow/calva/issues/396)
--- a/shadow-cljs.edn
+++ b/shadow-cljs.edn
@@ -1,4 +1,5 @@
-{:dependencies [[pez/cljfmt "0.0.4-SNAPSHOT"]]
+{:dependencies [[pez/cljfmt "0.0.4-SNAPSHOT"]
+                [zprint "0.4.16"]]
 
  :source-paths ["src/cljs-lib/src"
                 "src/cljs-lib/test"]
@@ -14,7 +15,8 @@
                              :jsify calva.js-utils/jsify
                              :cljify calva.js-utils/cljify
                              :parseEdn calva.js-utils/parse-edn-js
-                             :parseForms calva.js-utils/parse-forms-js}
+                             :parseForms calva.js-utils/parse-forms-js
+                             :prettyPrint calva.pprint.printer/pretty-print-js}
                  :output-to "out/cljs-lib/cljs-lib.js"}
                 :test
                 {:target    :node-test
--- a/src/cljs-lib/src/calva/pprint/printer.cljs
+++ b/src/cljs-lib/src/calva/pprint/printer.cljs
@@ -1,0 +1,23 @@
+(ns calva.pprint.printer
+  (:require [zprint.core :refer [zprint-str]]
+            [calva.js-utils :refer [jsify]]))
+
+
+(defn pretty-print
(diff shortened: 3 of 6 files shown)


kstehn:
Instead of ignoring those with literals That we cant read
Maybe supply a default literal handler to the edn Reader?
For example like here https://github.com/clojure-cookbook/clojure-cookbook/blob/master/04_local-io/4-17_unknown-reader-literals.asciidoc

PEZ:
We're not using `read-string` here. Because that would introduce some of the problems that we are trying to avoid (e.g. #378). The strings are parsed using `rewrite-cljs`, which just emit the literals literally (pun intended 😄).

So, literals are handled fine with this change. It croaks on trying to parse the `db` values, which I think is an improvement on what's happening inside nrepl, where the whole database gets dumped.  I'll put the reproduction steps in the OP so that it will be clearer what the change is about and how to test this PR.

On the topic of default literal handler. Here's a slightly less engineered example of it put to use: https://github.com/BetterThanTomorrow/calva/blob/master/src/cljs-lib/src/calva/js_utils.cljs#L20
We use `parse-string` there because I thought it was the safest way to parse `deps.edn`, and other clojure project config files. I'm less convinced today that it is safer than using `rewrite-cljs`, but maybe something is really wrong if we find unparsable values inside those...

PEZ:
Random thought: The reason we need to parse the `eval` result is that we get it as a string. If we could get it as EDN, then `zprint` can probably pretty print even those Datomic value references. I doubt we _can_ get the result as EDN, though, but there is an EDN nREPL transport...

cfehse:
@PEZ 

This implementation does not work for things that worked before. Your new implementation complains about a function in the data (invalid symbol) while the old implementation simply printed the name of the function. I think the parsing is a bit to optimistic about errors - it gives up to quickly. *gg*

[pprint-2.0.52.txt](https://github.com/BetterThanTomorrow/calva/files/3751886/pprint-2.0.52.txt)
[pprint-pr-414.txt](https://github.com/BetterThanTomorrow/calva/files/3751887/pprint-pr-414.txt)

PEZ:
That is actually an example of the problem at hand. The pretty printed version is much more than just pretty printed. It's exploded and huge. That could be a nice feature, but pretty print is a bad name.

PEZ:
So, I think these two (from my Datomic-test repo) illustrate the problem pretty well:

* [original.txt](https://github.com/BetterThanTomorrow/calva/files/3752961/original.txt)
* [exploded.txt](https://github.com/BetterThanTomorrow/calva/files/3752962/exploded.txt)

PEZ:
Agree it is a bit boring with the parser croaking like that. But there is not that much we can do about that...

cfehse:
@PEZ

Are you really sure that my example is expanded in the old pretty printed version? I take a look at this tomorrow.

PEZ:
Yeah, look at the interceptors. They only have `:name` in the original, and then when pretty printed they have more entries.

PEZ:
I changed the title of the PR to reflect the purpose rather than the implementation. I'm exploring solutions from several angles. One is to try solve this in `cider-nrepl`. Another angle is to investigate what options we can use with pretty printers like `zprint`. So have this issue open on zprint right now: https://github.com/kkinnear/zprint/issues/111

Issue #377, opened 2019-10-05
clojure.core/read-line doesn't work in Calva REPL


ozimos:
Repro
1.  `lein new app shopping-list`

2.  Add the following code to "src/shopping_list/core.clj"

         (defn prompt
          "get information from the user"
          [msg]
          (println msg)
          (read-line))

  3.  Jack-in to the project and load the file above `ctrl+alt+c enter`
  4.  Evaluate `(prompt "What is my name")` in the Calva REPL

  Calva REPL becomes unresponsive

  5.  In a new terminal window connect to running nREPL server. `lein repl :connect`
  6.   Evaluate `(prompt "What is my name")` in the terminal. 

All is well. Input is entered at the REPL successfully

PEZ:
Thanks. This should be fixed, even if I right now have no clue how to do it. If anyone else has a clue, I'd welcome help here!

PEZ:
Got this from @bbatsov : 

> There’s a stdin message that gets sent when the REPL is waiting for input.
> See https://nrepl.org/nrepl/0.6.0/ops.html#_stdin

cfehse:
This is fixes in Version 2.0.55.

Issue #237, opened 2019-06-25
[Feature request] Allow interruption of execution


lspector:
Interrupting execution of a long-running (possibly infinite) is a very desirable feature that I would love to have in Calva. 

Shooting for the moon here, but if execution could be interrupted and information about the context of interruption printed (ideally the call stack and super super desirably the values of locals in each stack frame), then that would be truly fabulous.

lspector:
Rationale: Without this feature, any infinite loop cannot be stopped without entirely quitting and restarting VSCode, losing context, etc. To continue working it is necessary not only to be able to regain control of the REPL but also to terminate execution of the process so that it doesn't consume CPU cycles, drain batteries, etc. 

This can be necessary both for accidental infinite loops and for intentional very-long-running processes that the developer realizes aren't doing what was intended, and have to be stopped and restarted after editing code.

bpringe:
@PEZ Isn't this now in the current version of Calva?

PEZ:
Indeed it is. Thanks to @cfehse !

lspector:
How? I just started an infinite loop and don't see any way to interrupt it.

bpringe:
Sorry, I should have specified! In the REPL window, press `ctrl+d` to interrupt a running evaluation. In the editor, I believe you can interrupt with this command:

![image](https://user-images.githubusercontent.com/12722744/67995120-12fbb580-fc06-11e9-86bf-6e62e9721b71.png)

lspector:
`ctrl+d` does not appear to do anything for me, when stuck in an infinite loop in the REPL window, whether I've actually entered the code in the REPL window or selected it in an editor window and used the command that evaluates a selection in the REPL window. 

It would be great if whatever actually does work would be easily discoverable. Ideally the same method for interrupting a running evaluation would also work no matter how/where one started it.

bpringe:
Will have to check this out later. CC @cfehse @PEZ

bpringe:
@lspector Can you provide detailed reproduction steps (commands used, whether in editor or in REPL, etc)? That would be very helpful and much appreciated.

lspector:
@bpinge:

- Launch VS Code
- Jack in Calva
- In the REPL: `(loop [] (println 'looping) (recur))`
- Try aborting with `ctrl-d`, no effect, CPU cranking
- Try adding additional modifier keys, no effect
- Try selecting "Calva: interrupt running Evaluations" from the command palette, no effect

[...]

PR #416, opened 2019-10-20
PROPOSAL enhance REPLwindow handling

cfehse:
## What has Changed?
- `src/connector.ts`
  - Moved code to create/show the REPL windows into a seperate function.
  - Improved error checking during the connect sequence.
-  `NReplEvaluation` in `nrepl/index.ts`
   - Enhanced the class to contain all message parsing.
   - Added properties which contain all information about the evaluation on resolve/reject.
   - Store any normal and error output of the evaluation for later use. 
   - Retrieve the stacktrace of an exception before reject the promise.
- functions `eval()` and `loadfile()` in `nrepl/index.ts`
   - Delegate message parsing and resolve/reject to the `NReplEvaluation`  class to ensure both methods behave identical.
- `repl-window.ts` improved reconnect sequence
   - Use the `NReplEvaluation` class to retrieve error output and display it in the REPL window. Unfortunately with will not show errors on the first connect with shadow-cljs because the repl needs some time to initialize completely and show the error during eval. I the REPL window is closed and reopened the error is shown on reconnect. This is maybe better than nothing. *gg*

calva-error-reconnect

- Added mechanism to add history entries only if they do not already exist.
- Added commands to clear the REPL windows
  - Added a command per REPL type.
  - Clear the history of the REPL window and clear the content in the webview.
- Add robust support for evaluation interruption (ctrl+d  - see #237)
- Added support for `read-line` from the REPL window ( see #377)

## My Calva PR Checklist

I have:

Ping @pez, @kstehn, @cfehse


cfehse:
Ping @pez, @kstehn

Just take a look it this changes. If we decide to include this I can break it down into smaller pieces.

kstehn:
I like this when i have more time will look into the code change.
One thing is i would, like you already said, break it down for now i see two pieces directly.
Your main Change and the "clear Command".

For the clear command i think it would be cool if there would be one command that clears the output for the focused repl window.

cfehse:
@kstehn

It would be cool to have a context menu in the webview or to have the ability to reload the webview. I found this solution with two commands the least confusing - at least less confusing compared to the command that currently exists.

PEZ:
Looks very promising! I can say directly that I think this should be done. (Haven't had a look at the source, but saying this with your descriptions as a basis, and knowing that you know what you are doing code-wise 😄 )

Wondering about this:

> * Added commands to clear the REPL windows
>   * Added a command per REPL type.
>   * Clear the history of the REPL window and redisplay it.
>   * Ensure that the REPL windows stays in the same view column and namespace.

Is this that you are making the command for clearing the window more visible? Asking because there already is such a command. Same question regarding the history.

And not sure what is meant by the last bullet there.

cfehse:
The two new commands replace the old command to clear the history of the active REPL window. The new commands are displayed in the nREPL selection (the button in the statusbar) depending on the state.

The commands clear the history and reload the corresponding REPL window (dispose it an show it again). The old command cleared the history and told the user to close and reopen the REPL windows. This is now done in the command.

PEZ:
Another enhancement of the history I wish for is to only add items if they are not exactly the same as the previous one. It might be because I only use the REPL window when testing it, but I often have to go history up many times before getting past all the `1`s. 😄

cfehse:
> Another enhancement of the history I wish for is to only add items if they are not exactly the same as the previous one. It might be because I only use the REPL window when testing it, but I often have to go history up many times before getting past all the 1s

Added that. *gg*

cfehse:
@PEZ  @kstehn @jmingtan

I added an implementation to retrieve user input from the REPL window. Currently the REPL window statically returns a string. This will be fixed in the next iteration.

cfehse:
@PEZ @kstehn @jmingtan

I now finished the `read-line` support for the REPL Window. The behavior should be exactly as in a console repl. Due to this feature I implemented a robust way to interrupt running evaluations in the REPL using `ctrl-d` (see issue #237) - and added a hint to the REPL HTML file.

cfehse:
@PEZ @kstehn

I moved the PR to the upstream repository.

PR #422, opened 2019-10-23
PROPOSAL enhance REPL window handling

cfehse:
## What has Changed?
- `src/connector.ts`
  - Moved code to create/show the REPL windows into a seperate function.
  - Improved error checking during the connect sequence.
-  `NReplEvaluation` in `nrepl/index.ts`
   - Enhanced the class to contain all message parsing.
   - Added properties which contain all information about the evaluation on resolve/reject.
   - Store any normal and error output of the evaluation for later use. 
   - Retrieve the stacktrace of an exception before reject the promise.
   - Make sure that a running evaluation will be correctly rejected during interruption.
   - Added mechanism to interrupt all running evaluations at once.
   - Added mechanism to provide generic user input for evaluations not running in the REPL window.
- functions `eval()` and `loadfile()` in `nrepl/index.ts`
   - Delegate message parsing and resolve/reject to the `NReplEvaluation`  class to ensure both methods behave identical.
- `repl-window.ts` improved reconnect sequence
   - Use the `NReplEvaluation` class to retrieve error output and display it in the REPL window. Unfortunately with will not show errors on the first connect with shadow-cljs because the repl needs some time to initialize completely and show the error during eval. I the REPL window is closed and reopened the error is shown on reconnect. This is maybe better than nothing. *gg*

calva-error-reconnect

- Added mechanism to add history entries which will not pollute the history. Identical evaluation run many time in a row will only be added once.
- Added commands to clear the REPL windows
  - Added a command per REPL type.
  - Clear the history of the REPL window and clear the content in the webview.
- Add robust support for evaluation interruption (ctrl+d  - see #237)
- Added support for `read-line` from the REPL window ( see #377) and in the editor.
- Added command `"Interrupt running Evaluations"` (`ctrl+alt+c ctrl+alt+d`) to interrupt all running evaluations in editors and REPL windows. 
- Made the message handling and evaluation in the REPL window completely asynchronous (removed all unnecessary async/awaits). 

## My Calva PR Checklist

I have:

Ping @pez, @kstehn, @cfehse


cfehse:
@PEZ @kstehn @jmingtan

I now finished the `read-line` support for the REPL Window. The behavior should be exactly as in a console repl. Due to this feature I implemented a robust way to interrupt running evaluations in the REPL using `ctrl-d` (see issue #237) - and added a hint to the REPL HTML file. 

I moved the PR to the upstream repository so the team can more easily review/change the code.

jmingtan:
Hi @cfehse I tested the change and it works well. Now, when I execute `(read-line)` in the REPL window, I'm able to type in some user input. After pressing enter, my original input is removed and replaced with the repl result.

Just my 2 cents - maybe we could still keep the original user input, and print the repl result on the next line?

cfehse:
@jmingtan 

>Just my 2 cents - maybe we could still keep the original user input, and print the repl result on the next line?

You are absolutely right - can be done. Has been done. Now the behavior is like in a console repl.

Thanks.

cfehse:
@PEZ @kstehn 

I made the interruption in the REPL window robust and added the command `"Interrupt running Evaluations"` (`ctrl+alt+c ctrl+alt+d`) to interrupt all running evaluations in editors and REPL windows. This should be usefull if something like an infinite loop is evaluated in the editor which can easily block the complete editor repl session.

PEZ:
I love the interrupt!

It fails interrupting this though:
```clojure
(ns sandbox.lab
  (:require [clojure.core.async :refer [go]]))

(comment
  (go
    (while true
      (println "hello")
      (Thread/sleep 1000))))
```

It says ”There are no running evaluations to interrupt” and keeps printing! 😄

PR #450, opened 2019-11-04
Use paredit in the signature provider to find correct symbol

### LINKED ISSUE #410: Incorrect red highlights around brackets/paren in specific case
Sorry if the title was oddly worded. I didn't know the best way to state the issue.

In the REPL window, if I create a vector of vars I get red highlights in the following manner, which seem to be indicating mismatches, but there are no mismatches. The following code evaluates just fine, but it looks like it would not.

image

cfehse:
## What has Changed?

- Changed the method to find the symbol for the `info` call to use paredit and to retrieve the symbol from the containing `( )` list.
- Corrected the handling of `'junk'` and `'punk'` in `forwardSexp()` and `backwardSexp()`.

Fixes #410

## My Calva PR Checklist

I have:

Ping @pez, @kstehn, @bpringe


PEZ:
I truly like that you go for the structural way. However, `paredit.js` is abandonware, and I think we should try get rid of it, rather than build more with it. Could this maybe be done using the `LispTokenCursor` like we do here: https://github.com/BetterThanTomorrow/calva/blob/master/src/utilities.ts#L91

cfehse:
This can surely be done with the `LispTokenCursor`. If I have time and the mood I will change that eventuelly - but for now this is better than the current implementation which does not check if the current form is the form for which the symbol can be retrieved.

cfehse:
@PEZ

By the way what is the advantage of the `LispTokenCursor` over `paredit`. I find neither of them especially easy to use - as far as I now know both.

PEZ:
I like paredit as such. 😄 What I am objecting to is `paredit.js`, which is bug ridden and not maintained.

Our own paredit is built with our own `LispTokenCursor`, which might not always have the perfect API , but it is at least maintained, and much more complete and has fewer bugs than `paredit.js`. Also, we can always build some nicer abstraction on top of it. A bit like our paredit.ts module does.

I'll see if I have the time to try modify this PR to use our own stuff instead. If not, we'll merge it and fix that later, as you suggest.

cfehse:
Okay I understand. *gg*

PEZ:
OK. So here is a version using docmirror+LispTokenCursor: https://github.com/BetterThanTomorrow/calva/commit/77e5a97aa57074c162b8d5f614b25e3472bb96c9

Unless I have misunderstood something, it is equivalent to the paredit.js version. I actually think the TokenCursor API is pretty sweet, at least for this task.

When doing this I noticed one more advantage with the docmirror approach: It avoids building a new AST every time the symbol is fetched.

There is a big risk that I *have* misunderstood something though, so therefore I am first pushing this in its own branch, so that you can have a look before I update this PR.

cfehse:
>When doing this I noticed one more advantage with the docmirror approach: It avoids building a new AST every time the symbol is fetched.

For that reason I only builded the AST for the current toplevel form in my approach. But nevertheless your version looks very clean. I will take a look and integrate it. Thanks.

PEZ:
I'll push my changes and then it is ready to merge, I'd say.

cfehse:
> I'll push my changes and then it is ready to merge, I'd say.

Your code is not doing, what it is supposed to to. Wait - I take a look at this.

PEZ:
> Your code is not doing, what it is supposed to to. Wait - I take a look at this.

Too late. 😄 

But at least I haven't merge the PR yet.

cfehse:
@PEZ 

This `docMirror.getDocument(document).getTokenCursor(idx);` returns the Cursor set to the last token of the last line of the doc regardless where the position is.

cfehse:
@PEZ 

I am handle this over to you now - I do not have the time to rework the whole thing. The idea is to find the correct symbol regardless in which place you are in a form.
Examples ( `| ` is the cursor):
````
(def  | (fn   [1   map  ]  {   } )    ) => def
(def    (fn | [1   map  ]  {   } )    ) => fn
(def    (fn   [1 | map  ]  {   } )    ) => fn
(def    (fn   [1   map  ]  { | } )    ) => fn
(def    (fn   [1   map  ]  {   } ) |  ) => def
```

PEZ:
Now I think I got it right. At least for the test cases you provided there. Look:

signature-help-test

Please help sanity check the LispTokenCursor method I added for helping with the task.

cfehse:
@PEZ

I take a look at that tomorrow. Thanks.

PEZ:
Tomorrow is such a long time away, so I added two more things to the signature help:

1. Show active parameter.
   * Only for regular forms so far.
   * It keeps `& more` active when that is reach and more arguments added.
2. Show function name in signature help.

It looks like so:

active-params-in-signature

I also made the `escape` key dismiss the signature hover.

PEZ:
Now the bugs reported in calva-dev should be fixed. They were about:

* argument lists where parameter destruction is going on
* ignored forms in the calling argument list
* macros with optional arguments (of type `arg?`). These don't get active parameter markup now.

As far as I can tell, this should be good to go.

cfehse:
@PEZ 

Am I missing something? The `getSymbol()` method of the `CalvaSignatureHelpProvider()` does not work at all for me. Retruns undefined. I pulled this branch from upstream and made a fresh rebuild. I have no idea.

PEZ:
Does the VSIX work for you?

cfehse:
@PEZ 

No the signature help will not pop up.

PEZ:
How odd. Could it be a Windows thing...

cfehse:
The VSIX does not work on WSL either. 

Now trying native linux...

cfehse:
@PEZ 

On native linux the code works. Very strange...

PEZ:
Works on my Windows in Virtual Box...

Is it the same project you use when it works and when it doesn't?

cfehse:
@PEZ 

It's not a Windows problem. It is a data problem. I have test data here where it is not working. `backwardList()` does return false where it should not I think.

PEZ:
Cool. Are you on it now? Otherwise I'd like that data. 😄

cfehse:
@PEZ 

The code does not work for me after a '(comment ...)` and under some circumstances in that comment form. It works for the first form in the comment.

cfehse:
@PEZ 

Here is some code which confuses the code:

```
(ns brainpie-test-api.test-case) 

(def )

;; Interactive development
(comment
  (def )
  (while true (do (println "start sleeping")
                  (Thread/sleep 1000)))

  (defn prompt
    "get information from the user"
    [msg]
    (println msg)
    (read-line))
  (prompt "What is my name")

  (go
    (while true
      (println "hello")
      (Thread/sleep 1000)))

  (go
    (def x (atom 1))
    (while (< @x 5)
      (do
        (println @x)
        (Thread/sleep 1000)
        (swap! x inc))))

  (defn fehse
    [x]

    (+ x (+ x 3)))

  (println (fehse))

  (def   (fn [1  map]))

  (defn   [] (map println [1 1 1 1] {1 1} #'- (fehse)))

  )
````

PEZ:
Works for me with that test data.

cfehse:
@PEZ

On my systems it is not working. Sadly...

PEZ:
So this is a Windows problem after all. At least related. Setting the line ending of the file in VS Code to `CRLF` will provoke the error. Now the hunt for why this throws things off begins...

PEZ:
Windows/CRLF issue should be fixed now.

Now the only case where I know the wrong argument is marked as active is in thread-first macros, such as `->`, and `some->`. I think I have an idea of how to fix that, but I don't think we need to hold shipping this PR because of that.

PEZ:
And now made it work also when inside thread-first forms.

Here's a demo showing the behaviour working for various test cases:

signature-help

cfehse:
@PEZ 

Now this looks good to me. Very nice feature indeed!

Issue #417, opened 2019-10-20
Expand selection fails at the start and end of the input of the REPL window


PEZ:
Place the cursor at the start or end of this REPL window input line and issue the command **Calva Paredit: Expand selection**:
```
user=> (:foo)
```

**Expected**:
`(:foo)` is selected.

**Actual**:
Nothing happens

Issue #403, opened 2019-10-15
Calva Highlight sometimes incorrectly recognises form as a `comment`  form


PEZ:
Picked this one up here: https://github.com/tonsky/clojure-warrior/issues/19

Something like:

```clojure
(if comment :a :b)
```

Get's highlighted as a comment form.

It's interesting because I'm sure Tonsky fixed one such issue some while ago, but clearly the regex must be worked on a bit more.

PR #456, opened 2019-11-07
Paredit and highlight fixes and improvements.

### LINKED ISSUE #170: [2.0] Paredit inconsistencies
Paredit behaves slightly different between the editor and the REPL window. It will sort itself out when we replace the editor Paredit with our shiny new one, but anyway.

Here are two things that the editor Paredit does better than the new one, IMO:

1. Wrap-around parens `()`, works in the editor, not in the REPL Window.
1. Wrap-around other things work, but in the REPL window they move the cursor next to the closing bracket, in the editor the cursor moves the opening bracket.

The new paredit does some things better than the old one. Here's one:
1. Wrap around with a selection wraps the selection. (The old paredit wraps only the first form in the selection.)

cfehse:
## What has Changed?

- Changed the `paredit.open()` method to wrap around the current selection (#170 ). 
- Keep the selection after wrap on `open()` and `wrapSexpr()`.
- Changed   `insertString()` to use `changeRange()` directly.
- Added a `dblclick` handler to the webview to execute `growSelection()` on mouse double click starting at the current position of the mouse pointer. This allows the selection of words with a double click like in the editor.

calva-paredit-dblclick-repl-window

- Fixed the `growSelection()` method to allow cursor positions outside an sexpr (#417).
- Fixed the regex for the calva highlighter to allow correct rendering of comment forms (#403).
Disables the call to `util.select()` for the `paredit.deleteBackward` command to avoid misplaced cursor position after a line was deleted (#458).

Fixes #170 #417 #403 #458 

## My Calva PR Checklist

I have:

Ping @pez, @kstehn, @bpringe


PEZ:
That double-click is great! We should add it to the editor as well. (Not in this PR, though. 😄 )

PEZ:
Double-clicking inside a string should probably not use grow-selection. It makes it hard to select words in the string.

PEZ:
Undoing wrap selection deletes the selection. I think this might be an old bug.

PEZ:
Those are the only things I find when testing. The rest works wonderfully as advertised.

cfehse:
> Double-clicking inside a string should probably not use grow-selection. It makes it hard to select words in the string.

You mean clicking in something like this `"This is a documentation string"`?

cfehse:
> Undoing wrap selection deletes the selection. I think this might be an old bug.

I take a look at this as well.

PEZ:
> You mean clicking in something like this `"This is a documentation string"`?

Yes. It is both good and bad that the whole string is selected, I guess. But it ”feels” weird.

cfehse:
> Undoing wrap selection deletes the selection. I think this might be an old bug.

I can not reproduce this for the wrap around commands. The behavior with `ctrl+z` does not delete the selection but removes the inserted backets/parens/curlies. What I noticed that the shortcuts for `paredit.wrapAroundSquare` and `paredit.wrapAroundCurly` can not be applied on a german keyboard.

cfehse:
@PEZ, @kstehn, @bpringe

I changed the double click behavior in the webview to select the current word in a string.

src/webview/token-cursor.ts
```diff
@@ -106,6 +107,26 @@ export class LispTokenCursor extends TokenCursor {
         return new LispTokenCursor(this.doc, this.line, this.token);
     }
 
+    /**
+     * Indicates if the current token is inside a string (e.g. a documentation string)
+     */
+    isInString() {
```
PEZ: I thought we already had a method for checking if we're inside a string? Or maybe that is only while parsing... But even if it is, maybe add that fact to the token?

cfehse: If so I didn't found it and the decision if in a string may depend on on the next and previous token so I think this method is correctly placed in the token cursor because only the cursor has the whole picture of the model.

PEZ: I was remembering correctly :smile:

https://github.com/BetterThanTomorrow/calva/blob/786d6dde336dfa53a91dcc1b6bc80eba99e71751/src/webview/token-cursor.ts#L414

cfehse: Hey cool! But I think the code is not sufficient. I will test it.

PEZ: What seems a bit odd to me is that double-clicking words work everywhere on in the REPL window, except the prompt. Is there a reason for that, you think? (I'm talking about the released Calva.)

cfehse: That's because outside the line model the default behavior of the double click is in place. We override the default for mouse events to set the blinking cursor on the right place. My new code takes care that this default behavior is preserved outside the line model.

PEZ: > My new code takes care that this default behavior is preserved outside the line model.

Yes. I saw that.

PEZ:
> I can not reproduce this for the wrap around commands. 

I now tried this a bit more and realise we have quite a few different behaviours. If I use the **Wrap around ()`** command on a left->right selection, the right thing is selected and I can undo the wrap around w/o undoing something else as well. However:

* If the selection is right->left (either by dragging the mouse in that direction, or by holding down the shift-key and expanding the selection with key-left) then the wrap around command does really funny things, and undoing it doesn't bring the cursor back where it was before the wrap.
* If I wrap around parens around a left->right selection by typing an opening bracket, like `(`, the selection is wrapped and keep being correctly selected, but undoing also deletes the selection. (This is probably what I had done when reporting on the behaviour).
* The really crazy things start to happen if I have a right-left selection and wrap using an opening bracket. Then:
   1. the selection is trippled! 
   1. the middle clone is wrapped
   1. the leftmost clone is selected, offset by one character
   1. undoing here:
       1. removes the parens around the middle clone of the original selection
       1. keeps all three clones
       1. undoing _here_:
           1. creates yet another clone of the selection 😂 
           1. selects the leftmost clone
               1. you might think this is the end of the madness, but you'd be wrong!
               1. undoing now:
                   1. does a bit of different things everytime I try, including creating new content that is a mix of the current content and some other entry in the repl history. I haven't found a pattern to it.
                   1. keeps the selection range where the original selection was.

OK. So complicated to follow, here's a gif:

repl-prompt-wrap-issues

Crazy, huh? 🤔

cfehse:
@PEZ 

Try it again Peter - I was not aware of the fact that the selection could be `selectionEnd` before `selectionStart` if selected from right->left. Made the code robust about that and the weirdness should be gone.

PEZ:
Much less crazy now! 😄 

I think a new issue arrived. I can't move the cursor right with key-right. It instead selects the character to the right w/o moving the cursor, and then the cursor won't move either left or right.

Also:

* Having the cursor adjacent to the left of a form and then using the **Wrap around** commands places the brackets adjacent to the right of the form.
* Having the cursor adjacent to the left of a form and then typing an opening bracket, places the brackets adjacent to the left of the form.
* Typing a bracket when there is a selection, wraps it correctly, but undo will delete the selection and misplace the cursor
   * undoing again gives you the selection back and the cursor position correct.
   * (yes, I've already reported this, but including for completeness).

cfehse:
> I think a new issue arrived. I can't move the cursor right with key-right. It instead selects the character to the right w/o moving the cursor, and then the cursor won't move either left or right.

Yes the cursor movement depends on the "direction" of the selection. I fixed that.

>Having the cursor adjacent to the left of a form and then using the Wrap around commands places the brackets adjacent to the right of the form.

Yes that was wrong.

> Having the cursor adjacent to the left of a form and then typing an opening bracket, places the brackets adjacent to the left of the form.

I am really not sure, what you doing. I do this:

problem-one

> Typing a bracket when there is a selection, wraps it correctly, but undo will delete the selection and misplace the cursor

And the same for this one: I am doing this - selecting both directions and execute `open` and `wrapSexpr` for each type of selection followed by undo (`Ctrl+Z`):

problem-two

PEZ:
> I am really not sure, what you doing

I type the character `(`. So, on my Swedish keyboard that is `shift+8`.

cfehse:
>  type the character (. So, on my Swedish keyboard that is shift+8.

Okay then we are doing the same - it is `Shift+8` on the german keyboard as well. 

> Having the cursor adjacent to the left of a form and then typing an opening bracket, places the brackets adjacent to the left of the form.

> Typing a bracket when there is a selection, wraps it correctly, but undo will delete the selection and misplace the cursor

Is this still present for you with the latest commits?

PEZ:
Yes. With some few variations on what happens. Here's a session where select something, then type `(`, then undo several times. And (this I now noticed) redo many times:

wrap-typing-bracket-issue

cfehse:
Did you typed that `repeat` just before selecting and wrap it with `paredit.open()`.? Do you edit in strict mode or not. 

On my system the changes in the model are played correctly back in order (Windows and Linux).

PEZ:
It doesn't matter if I am in strict mode or not. Or if I am doing this on something in the repl history or something I have just typed.

The redo bug is not new, btw. I just tried it in the released Calva. I'll file an issue about it.

PEZ:
Here's that issue. https://github.com/BetterThanTomorrow/calva/issues/462

In this PR I still have this issue:

> Typing a bracket when there is a selection, wraps it correctly, but undo will delete the selection and misplace the cursor

Too bad you can't reproduce it! How about @kstehn and @bpringe ?

cfehse:
On my systems the undo/redo works like a charm in both directions. Looking further into this tomorrow.

cfehse:
@PEZ By the way is there a specific reason that the redo command in the REPL is `Ctrl+Shift+Z` and not `Ctrl+Y` like in the editor?

PEZ:
Maybe because it is `ctrl+shift+z` in the editor in any sane setup? 😄 No, we should probably try to figure out what the setting is in the editor and use the same in at the repl prompt.

On my machine redo is `cmd+shitft+z`, everywhere, and I think it might be quite common with that `shift` ”reverses” the meaning of `ctrl/cmd+z`.

cfehse:
Mayby. *gg* On my stock vscode installation on Windows redo is `Ctrl+Y`. Was just a question.

cfehse:
@PEZ @bpringe 

I think this PR is ready for merging now.

PEZ:
Yes. And I have reached the limit to my testing of this since things are crazy on macOS anyway. Let's merge and we will find out. Especically together with my lexing fix things will work pretty much better than they do today with this PR.

After that, we're almost ready to take on Paredit in the editor. Haha.

PR #468, opened 2019-11-14
[WIP] Lex quoted stuff

### LINKED ISSUE #467: Tokenization errors
@PEZ 

Something similar is happening with the `@` character. The form:
`(print @fehse)` 
will be parsed to this:
```
0: {type: "open", offset: 0, raw: "(", state: {…}}
1: {type: "id", offset: 1, raw: "print", state: {…}}
2: {type: "ws", offset: 6, raw: " ", state: {…}}
3: {type: "junk", offset: 7, raw: "@", state: {…}}
4: {type: "id", offset: 8, raw: "fehse", state: {…}}
5: {type: "close", offset: 14, raw: ")", state: {…}}
````

_Originally posted by @cfehse in https://github.com/BetterThanTomorrow/calva/issues/463#issuecomment-552852481_

PEZ:
## What has Changed?

- In the REPL window, syntax highlighting, and Paredit commands regarding something like ``'`'`'@'(foo)``, now considers that as one token, instead of three, as it used to.
- This also fixes how the signature help matches active parameter for an argument like this.
- _NB: It does not fix Paredit commands in the editor, because they do not use our own Paredit yet._

This is fixed such that the tokenizer now considers all combinations of the characters ``'.`~@``, in any order and with any (one-line) whitespace between as one token, together with the list, symbol, or literal they prepend. I am using this regex: ``/(['`~@][\s\r\n]*)*/``, prepended in front of the token regexes for lists, literals, etc.

It also involves a different check for valid matching pairs. Where we used to have a dictionary of opening bracket character sequences, such as `#?(` matching against list type, I now instead just check if the opening token ends with `(`, `[`, or `{` and return true for `)`, `]`, and `}`, respectively.

This fixes #467, and then some.

## My Calva PR Checklist
I have:

Ping @pez, @kstehn, @cfehse

Also pinging @mseddon for sanity check, since you wrote this stuff originally. 😄

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,13 @@
 Changes to Calva.
 
 ## [Unreleased]
-- [Fix: Lexing regex literal tokenisation](https://github.com/BetterThanTomorrow/calva/issues/463)
-- [paredit.deleteBackward sets cursor position wrong when deleting a line. ](https://github.com/BetterThanTomorrow/calva/issues/458)
-- [Calva Highlight sometimes incorrectly recognises form as a `comment` form](https://github.com/BetterThanTomorrow/calva/issues/403)
-- [Expand selection fails at the start and end of the input of the REPL window](https://github.com/BetterThanTomorrow/calva/issues/417)
+- Fix: [paredit.deleteBackward sets cursor position wrong when deleting a line. ](https://github.com/BetterThanTomorrow/calva/issues/458)
+- Fix: [Calva Highlight sometimes incorrectly recognises form as a `comment` form](https://github.com/BetterThanTomorrow/calva/issues/403)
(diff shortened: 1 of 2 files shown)


cfehse:
@PEZ 

I checked the changed regex expressions for infinite matching - and they are all good. That's a first step to sanity. *gg*

PEZ:
Thanks. I was thinking about that fix you made regarding infinite matching when I added all those Kleene stars. 😄

cfehse:
That's why I checked it. *gg*

PEZ:
I merged dev, and therefore, @cfehse's paredit fixes PR. Please try use this branch (or the VSIX from here) as much as you can.

PEZ:
I'll merge this now. There is probably no better way to discover any issues with it, anyway. 😄

PR #475, opened 2019-11-17
[WIP] Using our Paredit in the editor

PEZ:
## What has Changed?

Work in progress!

All but one Paredit command moved: `transpose`. That will be fun to add.

Also issue with compound commands, where I now chain the next edit on the Thenable that is returned by `editor.edit()`. I think it is because it resolves before we have updated our doc-mirror. For now, I've ”solved” it with a `setTimeout()` 😊 , with the expected result of the commands not always compounding fully. The real fix is to hold the promise until our doc-mirror is updated (if my analysis is correct, of course). To try this hack, use the `SpliceAndKull...` commands.

**NB:** I have also decided that commands like copy/cut this or that paredit-navigation, will be replaced by commands for selecting instead. So less commands, but the user will still be able to easily cut or copy with structural navigation.

### Fix these things

* [x] Indent/format the Paredited code
  * [x] Works for some edits (e.g. slurp, but not barf)
  * [x] Make it work when edits are outside the current list.
    * [x] Add option to format two levels of enclosing lists
    * [x] Use two levels option for Paredit edits formatting
    * [x] Add unit tests for the two levels option
    * [ ] Use two levels option only when necessary (because formatting larger chunks causes flicker)
  * [x] Try get the formatting part of the same undo as the edits.
* [x] Solve the atomic multiple edits issue

#### Editing

* [x] `barfSexpForward`
* [x] `slurpSexpBackward`
* [x] `barfSexpBackward`
* [x] `slurpSexpBackward`
* [x] `splitSexp`
* [x] `spliceSexp`
* [x] `killSexpForward`
* [x] `killSexpBackward`
* [x] `killListForward`
* [x] `killListBackward`
* [x] `spliceSexpKillForward`
   * [x] Moved
   * [x] Also splice
* [x] `spliceSexpKillBackward`
   * [x] Moved
   * [x] Don't miss the splicing
* [x] `deleteForward`
  * [x] Moved
  * [x] Strict mode not working here
* [x] `deleteBackward`
* [x] `wrapAroundParens`
* [x] `wrapAroundSquare`
* [x] `wrapAroundCurly`
* [ ] `transpose` (does not exist in our paredit yet)
* [x] `raiseSexp`
* [x] `convolute`

Not needed:

* [x] `indentRange` (this should be removed)

#### Navigational

* [x] `forwardSexp`
* [x] `backwardSexp`
* [x] `forwardDownSexp`
* [x] `backwardUpSexp`
* [x] `forwardUpSexp`
* [x] `openList`
* [x] `closeList`

#### Selection

* [x] `sexpRangeExpansion`
* [x] `sexpRangeContraction`
   * [x] Moved
   * [x] Fix issue with not contracting back fully after a few grow/shrink sequences
* [x] `rangeForDefun`
* [x] `forwardSexp`
* [x] `backwardSexp`
* [x] `forwardDownSexp`
* [x] `backwardUpSexp`
* [x] `forwardUpSexp`
* [x] `openList`
* [x] `closeList`

Fixes #

## My Calva PR Checklist

I have:


bpringe:
>NB: Growing and shrinking some stops shrinking back fully after a few repeats. This is a bug in the currently released Calva REPL window as well. Probably easy to fix.
Can you provide a gif of this erroneous behavior?

PEZ:
> Can you provide a gif of this erroneous behavior?

Not sure the gif would be able to show it. I need some way to show what keys I am pressing... But it is easy to reproduce:

1. Grow the selection three steps.
1. Shrink four steps.
1. Grow three steps.
1. Shrink four steps.
1. Keep repeating this.

For me the minimum selection just keeps growing.

PEZ:
And now I fixed the bug. To confirm try the grow/shrink dance in this branch and in the currently released Calva's REPL window.

PEZ:
Now you can convolute the form:

convoluted

¯\_(ツ)_/¯

PEZ:
This is about to get ready for release, as far as features are concerned. I have been testing it a bit, but it would be super awesome to get more testing done, and also some sanity check on the code changes.

The command still not moved is `transpose` which we will need to create. I might do it or I might decide we can ship without it, since maybe it is not the most widely used edit.

There is also an unfortunate difference in how `growSelection` selects strings. It used to start with selecting the contents of the string, but now it selects it all, including quotes. Again, not a stopper, VS Code has a built in **Expand selection** command that compensates somewhat. I think the real solution is to make our parser treat strings more like lists.

cfehse:
@PEZ

I can do some testing for this. Is there any description/definition anywhere what the different commands are supposed to do? Would be easier as to figure it all out by myself.

PEZ:
@cfehse Good point. I don't know of such a resource. We should write one. Even if it doesn't help you much with the testing. Just ask here if there is something I can try describe.

PEZ:
I now have added select commands for each movement command: https://github.com/BetterThanTomorrow/calva/tree/wip/editor-using-our-paredit/src/paredit#selecting

bpringe:
I have yet to really use paredit commands for editing, but I intend to start getting familiar with them. I guess now is a good time 😄. I can check this out later tonight myself for further QA.

stefan-toubia:
What a huge improvement!  I'm very excited to see progress here, the editing in particular is a big win.  Not using chords for paredit makes this so much more usable, the previous keyboard mapping was making it pretty difficult to use.

I've spent a little time getting familiar and I've got a few notes/questions.  Keep in mind, I've never been an emacs user and don't have a lot of familiarity with paredit, so I don't know what the standard is.

### Navigating
#### Behavior
- Navigate forward/back puts you at end of current sexp instead of beginning of next.
Calva currently behaves this way and I always wondered why.  My intuition would be that navigating forward a sexp would put the cursor on the begginning of the next sexp.  Right now many of the navigation commands seem to leave the cursor at the end of the sexp.
It would be interesting to see the sexpr forward/back nav have the behavior of aligning the cursor on the start of the next sexp, but if it is the last sexp the cursor goes to the end.  What do you think?  I could be missing something important here.

#### Bugs
- Cursor does not say in scroll view during sexp navigation.
Unlike normal cursor navigation, the cursor does not say in view if you parnavigate down a file.
- Backward sexp doesn't work in some areas
If you put your cursor on some elements such as `let`, `println`, `comment` or `deftest` or `testing` the backwards sexp command does not work.  This is not a problem on other elements, such as variable names.

### Editing
I'm still very slow with parediting but I'll be testing this out more in the coming days.  All I can say right now is it is a great improvement!

PEZ:
@stefan-toubia Thanks for helping to test this!

> It would be interesting to see the sexpr forward/back nav have the behavior of aligning the cursor on the start of the next sexp, but if it is the last sexp the cursor goes to the end.

I have opted to make it behave like word-forward/backward in standard text editing. On my Mac that is `alt+right/left`. My reasoning is that sexps in LISP are like words in plain text. Also, since modifying the movement commands with `shift` makes for selecting the range, we probably want to select to the end of the current form rather than the start of next, and it makes sense for movement and selection to otherwise be equal.

You can try the behaviour you are curious about quite easy. In `rangeToForwardSexp` do `cursor.forwardWhitespace()` before returning the range. Here: https://github.com/BetterThanTomorrow/calva/blob/wip/editor-using-our-paredit/src/cursor-doc/paredit.ts#L62

(And similar in the backwards case.) But, regardless, I think the current behaviour makes the most sense.

> Cursor does not say in scroll view during sexp navigation.

Not good. I'll have a look. Hopefully easy to fix!

> Backward sexp doesn't work in some areas If you put your cursor on some elements such as let, println, comment or deftest or testing the backwards sexp command does not work. This is not a problem on other elements, such as variable names.

Interesting. It seems to be about not moving backward when the cursor is inside the token at the [”function position”](https://www.clojure.org/guides/learn/syntax#_structure_vs_semantics), or rather the first token at the start of a list of any type. I then expected it to behave similar for the last token in a list, but there it works. I think you've found a bug in our Token Cursor. Very good find! It is actually a nice side effect of implementing movement and editing using our token cursor, that it helps us reveal how it works (or not works, in this case) inside there.

PEZ:
I've fixed the two bugs now, @stefan-toubia:

1. Revealing the cursor when it moved outside the viewport was easy.
1. Fixing the **Backward Sexp** command at function position was tricker. It is indeed a bug in the TokenCursor, but it is quite a big job to fix that bug, so I made a workaround for it in the paredit module.

stefan-toubia:
@PEZ that was quick!  Glad I could help out a bit.

I can definitely appreciate the challenge in defining the best behavior of the paredit navigation.  Even after spending some time testing out the current behavior I wasn't confident that there is a better way to do it.  I was mostly coming from the fact that the forward/back is the once I use the most and I constantly find myself having to do a little back and forth dance to get to where I want to be (usually at the start of a sexp).  I imagine this could be a different story if I was more proficient with the other commands.  I could just be biased since I'm a novice with lisps!

Good to know that this should be easy to test out, thanks for the tip.  If I get some time I just might do that.

PEZ:
I just fixed a tricky issue with the cursor positioning after barfing forward: If you barfed with the cursor at the end of what would become the new last form in the list, the caret would move. It shouldn't, and now it doesn't.

Fixing this meant fixing the whole promise chain:

* edit -> formatting -> cursor adjust ->  more edits -> ...

This meant I could get rid of some icky `setTimeout()` things I had added before. Now we can chain edits in arbitrary sequences and thus we can build  more complex paredit commands using simpler ”primitives”. As an example, **Splice and Kill Backwards** is now implemented like so:

```clojure
export function spliceSexpKillingBackward(doc: EditableDocument, start: number = doc.selectionEnd) {
    killBackwardList(doc).then(() => {
        spliceSexp(doc, doc.selectionEnd, false);
    });
}
```

PEZ:
Now I have added gifs for almost all commands to the Paredit documentation page. Looks a bit like so: https://github.com/BetterThanTomorrow/calva/blob/wip/editor-using-our-paredit/docs/readthedocs/source/paredit.md

While doing this I found these bugs:

1. Barf forward still has issues with cursor position after the barf.
1. Barf backward also has issues with this. But with more nasty results when repeating barf.
1. Wrap around parens (all kinds) misbehave quite badly. I'm a bit too tired to describe it.
1. (Already known): **Strict mode Forward Delete** misbehaves.

PEZ:
I've fixed **wrap around parens** now. And did it proper, adding a method on the token cursor that returns the current form according to my strict preferences for how that should be determined. Function doc is like so:

```javascript
   /**
     * Figures out the `range` for the current form according to this priority:
     * 1. If `offset` is adjacent after form
     * 2. Else, if `offset` is adjacent before a form
     * 3. Else, if the previous form is on the same line
     * 4. Else, if the next form is on the same line
     * 5. Else, the previous form, if any
     * 6. Else, the current enclosing form, if any
     * 7. Else, return `undefined`.
     * @param offset the current cursor (caret) offset in the document
     */
    rangeForCurrentForm(offset: number): [number, number] {
```

With this method in place we can now find every place in Calva where the current form is needed and use this one, and then it will be consistent across all the extension. I've tested it quite extensively. And since it is not mutating anything externally it should be quite auto-testable. Might be a good place to start our unit testing of the TS parts of the extension, @kstehn, @bpringe .

PEZ:
Alright. So I continued to have problems with adjusting the cursor position after the editing commands. Tucking it on the promise chain doesn't really help, because the formatter can change the text pretty wildly. So, I have now added a mechanism to the `edit` batches where the desired selection post-edit is passed along, and then can be performed before the formatting. This works, since the formatter is good at figuring out what the cursor position should be after formatting, if the cursor is rightly positioned before the formatting.

**The good news**: Cursor positioning is now really smooth and sane after edit commands.

**The bad news 1**: In the REPL window, some editing commands get performed twice! I have no idea why. And too tired now to figure anything out.

**The bad news 2**: This was a pretty big rewrite of how quite basic things work, so lots of things might have broken. What looked like a Release Candidate, now is alpha again (with bugs see bad news 1).

Please, please, please help test this. Use this guide to see what basic things need to work: 

* https://github.com/BetterThanTomorrow/calva/blob/wip/editor-using-our-paredit/docs/readthedocs/source/paredit.md

And then be creative to try figure out where it might not work.

Since in the REPL window, I know there are nasty bugs now, please help test there as well. It doesn't implement all Paredit commands yet, so it can be a bit tricky to test, but see this list of key bindings for the REPL window to get to know what commands are implemented and what shortcuts they might have if not the same as in the editor. 

* https://github.com/BetterThanTomorrow/calva/blob/wip/editor-using-our-paredit/src/webview/repl-console.ts

Pinging in people I can think of that might be able and willing to help testing: @Gnurdle , @slipset , @bpringe , @cfehse , @kstehn , @stefan-toubia 

Please test using the dev environment or with this VSIX: https://2586-125431277-gh.circle-artifacts.com/0/tmp/artifacts/calva-2.0.62-wip-editor-using-our-paredit-dd3bee27.vsix

Thanks in advance! ❤️

PEZ:
> The bad news 1: In the REPL window, some editing commands get performed twice! I have no idea why. And too tired now to figure anything out.

I'v tested the version before my latest change now, and this problem existed there too. Which is good, because it would have been really strange if yesterdays change had that effect.

PEZ:
OK. So now I know why some commands are being doubled. A while ago, in this PR, I changed the default hot keys in the window to the same as those defined for the editor... It's a bit tricky to fix, because it is undefined which command arrives first. And also registering a hot key is how commands are registered, meaning I can't just disable the repl window's own hotkeys for the commands in question...

PEZ:
Pasting the last commit message details here as well:

* No more doubling of commands
* No chorded key bindings, b/c repl window can't
* Help repl window with selection after edit and with undos
* Editor and repl use separate methods for composing paredit commands

No more chording means changing default shortcuts for some commands, mostly to do with splice, wrap, split and join. See the updated guide: https://github.com/BetterThanTomorrow/calva/blob/wip/editor-using-our-paredit/docs/readthedocs/source/paredit.md 

Now I think things mostly work. I recommend using this version instead of the released one. Not only because I want things tested. 😄 It is actually quite a lot better than current Calva.

bpringe:
I've tested all the commands in the editor, and pending the lexer fixes regarding strings, all of those appear to be working fine.

I started testing the commands in the REPL window and so far I've only gotten through the navigation ones. Here are some issues I found:

image

PEZ:
The REPL window does not have implementations for a lot of the commands that the editor implements. It's mostly a matter of configuring it though, since they share the Paredit ”engine” now. I think most of (maybe all) the issues you reported there are because of those omissions.

bpringe:
I've found these issues in the REPL window with selection commands, which appear to be stemming from some common bug:

image

bpringe:
image

PR #471, opened 2019-11-16
Add stacktrace option to test runner output

stefan-toubia:
## What has Changed?

- Add exception stacktrace to test runner output for tests that failed in error
- Add calva option `outputStackTraceOnTest` to enable outputting stacktrace

## Details
This is still WIP, but here's my first pass at adding stacktrace to the test runner output.  This prints stacktraces in the output pane for each test that had an error.

This uses cider's `test-stacktrace` command, which has been added to `NReplSession`.  This works by requesting the stacktrace for each error individually, I'm not sure if there would be a performance hit if for example you run all tests and there are many errors.  I believe that emacs gives only shows each stacktrace on demand if you want to see it.

What I have so far accomplishes the primary goal of getting a stacktrace for a test error, but there is room for improvement on how the stacktrace is output.  VS code's output window is a bit limited, and a lot of stacktraces could potentially overwhelm the output.  It would be nice if sections of the output were collapsable.  I've thought of a few alternate approaches:
- Write the stacktraces out to temp files in the workspace folder and put a clickable link in the output.
- Write the stacktraces out to the REPL webview.  I'm not sure how easy this would be, but the new stacktrace formatting in the REPL webview is very nice and would be great to reuse.  Ideally there would be a link in the test output that could trigger the REPL window to show the stacktrace, but this is the part that I'm not sure is possible.  This would be a better solution than linking to a log file.
- Use the VS Code Diagnostic output to show the stacktrace on the `Peek Problem` link, similar to how the test failures are already working.  I haven't played with this yet, but I think it should be implemented anyway.  It looks like it should be easy to add.  Maybe this would allow for querying the stacktrace on demand instead of all at once.  I need to read up on this.

### Output
```clj
(deftest error-test
  (is (= 1 (/ 1 0)) "Try dividing by 0"))
```
```
ERROR in hello.hello-test/error-test (line 15):
Try dividing by 0
expected: (= 1 (/ 1 0))
   error: java.lang.ArithmeticException: Divide by zero (Numbers.java)
            Numbers.java: 188 clojure.lang.Numbers/divide
            Numbers.java:3901 clojure.lang.Numbers/divide
          hello_test.clj:  15 hello.hello_test$fn__6529/invokeStatic
          hello_test.clj:  14 hello.hello_test$fn__6529/invoke
                test.clj: 203 cider.nrepl.middleware.test$test_var$fn__6746/invoke
                test.clj: 203 cider.nrepl.middleware.test$test_var/invokeStatic
[...]

PR #480, opened 2019-11-22
Add evaluation status bar item

stefan-toubia:
## What has Changed?

- Add a new status bar item for evaluation status

I've been bitten more than a few times because I missed the fact that I encountered an evaluation error, either because the bottom of my output channel was not visible or the output panel was collapsed.  This adds a simple status indicator to make it idiot-proof.

#### No evaluation
image

#### Evaluating (animated)
image image image

#### Success
image

#### Error
image

#### Action
Clicking the icon will open the output panel.  I couldn't get it to do more than this, I didn't put too much effort into it, it would be nice to make it toggle and ensure that the Calva Says output is shown.

## My Calva PR Checklist

I have:

- [ ] Read [How to Contribute](https://github.com/BetterThanTomorrow/calva/wiki/How-to-Contribute#before-sending-pull-requests).
- [ ] Made sure I am directing this pull request at the `dev` branch. (Or have specific reasons to target some other branch.)
- [ ] Made sure I am changed the default PR base branch, so that it is not `master`. (Sorry for the nagging.)

Ping @pez, @kstehn, @cfehse

### DIFF
--- a/src/evaluate.ts
+++ b/src/evaluate.ts
@@ -1,6 +1,7 @@
 import * as vscode from 'vscode';
 import * as _ from 'lodash';
 import * as state from './state';
+import { EvaluationStatus } from "./state";
 import annotations from './providers/annotations';
 import * as path from 'path';
 import select from './select';
@@ -168,6 +169,11 @@ function evaluateCurrentForm(document = {}, options = {}) {
         .catch(e => console.warn(`Unhandled error: ${e.message}`));
 }
 
+function setEvalStatus(status: EvaluationStatus) {
+    state.cursor.set("evaluation", status);
+    statusbar.update();
+}
+
 async function loadFile(document, callback: () => { }, pprintOptions: PrettyPrintingOptions) {
     let current = state.deref(),
         doc = util.getDocument(document),
@@ -181,7 +187,7 @@ async function loadFile(document, callback: () => { }, pprintOptions: PrettyPrin
     if (doc && doc.languageId == "clojure" && fileType != "edn" && current.get('connected')) {
         state.analytics().logEvent("Evaluation", "LoadFile").send();
         chan.appendLine("Evaluating file: " + fileName);
-
+        setEvalStatus(EvaluationStatus.evaluating);
         let res = client.loadFile(doc.getText(), {
             fileName: fileName,
             filePath: doc.fileName,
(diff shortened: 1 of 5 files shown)


src/evaluate.ts
```diff
@@ -191,18 +197,21 @@ async function loadFile(document, callback: () => { }, pprintOptions: PrettyPrin
...
             } else {
+                setEvalStatus(EvaluationStatus.none);
                 chan.appendLine("No results from file evaluation.");
             }
         }).catch((e) => { 
             chan.appendLine(`Evaluation of file ${fileName} failed: ${e}`);
+            setEvalStatus(EvaluationStatus.error);
         });
     }
     if (callback) {
```
stefan-toubia: This was recently added, I'm not familiar with the purpose.  Would it make sense for this to also be represented in the status bar indicator?

PEZ: I think this might be from @cfehse 's fix for the load file callback. And yes, would probably be material for a status indicator.

cfehse: Yes it is.

cfehse: The purpose was solely to show that an error occurred instead of ignoring the error condition completely.

PEZ:
I think this PR addresses a big and important issue with Calva. We should really do this. However, I don't think adding more status bar items is the way to do it. How about we instead:

* use the `clj` indicator
* Make it red when there are errors
  * Put the error message in the hover.
  * Add an action to it that shows the stack trace in the REPL window. 
* animate its color during evaluation, say pulsate between grey and white (if possible)

This indicator sometimes is a button already. For toggling `cljc` repl between `clj` and `cljs`. So for the stack trace action we will need to show a menu offering both actions. I think we should then change it so that it always shows a menu. We can move the actions for opening the REPL windows there.

stefan-toubia:
@PEZ fair enough about the status bar items, there are enough of them already.  I like the plan of using the `clj` indicator.  This should be a quick change over, I will do some refactoring to the status bar update function to clean this up a bit.

> Add an action to it that shows the stack trace in the REPL window.
I think what would make sense here is to implement this as a separate PR after #471 is finished, which implements some changes to outputting messages and stacktrace to the REPL window, which should make this very straightforward to implement.

PEZ:
> > Add an action to it that shows the stack trace in the REPL window.
> 
> I think what would make sense here is to implement this as a separate PR after #471 is finished, which implements some changes to outputting messages and stacktrace to the REPL window, which should make this very straightforward to implement.

Makes sense. It was that PR that got me to think about it. Let's just have the error message in the tooltip then for now?

stefan-toubia:
I'm closing this PR since it is stale, will re-open a new one when I have made more progress.

Issue #506, opened 2019-12-04
Paredit select forward/backward add to the selection stack even when they don't select anything


PEZ:
Repro:

1. Without anything selected, use *Expand Selection* to select the first form inside a list.
2. Use *Select Sexpr Backwards* (which doesn't have anything backwards to select, and won't.)
3. Use *Shrink Selection*.

**Expected**:
* The selection should go back to just the cursor you started with (at the position you started with).

**Actual**:
* The first form remains selected. (Shrinking again will get you back to where you started.)

PR #507, opened 2019-12-04
Fixing #506 select back/fwd stack issue w/ selection can't grow

PEZ:
WIP: I am fixing the issue and also adding some  unit tests for the Paredit grow and shrink selection stack.

Eventually #506 should be fixed. 😄 

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 Changes to Calva.
 
 ## [Unreleased]
+- Fix: [Select Backward * commands won't grow selection in REPL window](https://github.com/BetterThanTomorrow/calva/issues/498)
+- Fix: [Paredit select forward/backward add to the selection stack even when they don't select anything](https://github.com/BetterThanTomorrow/calva/issues/506)
 
 ## [2.0.66] - 2019-12-02
 - Fix: [Cursor moves forward after undoing wraparound commands in REPL window](https://github.com/BetterThanTomorrow/calva/issues/499)
--- a/docs/readthedocs/source/paredit.md
+++ b/docs/readthedocs/source/paredit.md
@@ -84,8 +84,8 @@ Default keybinding    | Action | Description
 
 Default keybinding                | Action | Description
 ------------------                | ------ | -----------
- `ctrl+right`                        | **Slurp Forward** |  Moves the _closing_ bracket _forward_, _away_ from the cursor, past the following form, if any. <br> ![](_static/images/paredit/slurp-forward.gif)
- `ctrl+left`                         | **Barf Forward** | Moves the _closing_ bracket _backward_, _towards_ the cursor, past the preceding form. <br> ![](_static/images/paredit/barf-forward.gif)
+ `ctrl+right`<br>**NB** On Mac this keybinding might need your attention, see [Quirks](quirks.html).                        | **Slurp Forward** |  Moves the _closing_ bracket _forward_, _away_ from the cursor, past the following form, if any. <br> ![](_static/images/paredit/slurp-forward.gif) 
+ `ctrl+left` <br>**NB** On Mac this keybinding might need your attention, see [Quirks](quirks.html).                        | **Barf Forward** | Moves the _closing_ bracket _backward_, _towards_ the cursor, past the preceding form. <br> ![](_static/images/paredit/barf-forward.gif)
  `ctrl+shift+left`                   | **Slurp Backward** | Moves the _opening_ bracket _backward_, _away_ from the cursor, past the preceding form, if any. <br> ![](_static/images/paredit/slurp-backward.gif)
  `ctrl+shift+right`                  | **Barf Backward** | Moves the _opening_ bracket _forward_, _towards_ the cursor, past the following form. <br> ![](_static/images/paredit/barf-backward.gif)
  `ctrl+alt+s`                      | **Splice Sexp** | Remove enclosing brackets. <br> ![](_static/images/paredit/splice.gif)
--- a/src/cursor-doc/model.ts
+++ b/src/cursor-doc/model.ts
@@ -51,10 +51,36 @@ export class ModelEdit {
     constructor(public editFn: ModelEditFunction, public args: any[]) { }
 }
 
-export type ModelEditSelection = {
-    anchor: number,
-    active: number
-};
+export class ModelEditSelection {
+    private _anchor: number;
+    private _active: number;
+
+    constructor(anchor: number, active?: number) {
+        this._anchor = anchor;
+        if (active !== undefined) {
+            this._active = active;
+        } else {
+            this._active = anchor;
+        }
+    }
+
+    get anchor() {
+        return this._anchor;
+    }
+
+    set anchor(v: number) {
+        this._anchor = v;
+    }
+
+    get active() {
+        return this._active;
+    }
+
+    set active(v: number) {
+        this._active = v;
+    }
+}
+
 
 export type ModelEditOptions = { 
     undoStopBefore?: boolean,
@@ -62,14 +88,6 @@ export type ModelEditOptions = {
     skipFormat?: boolean,
     selection?: ModelEditSelection 
 };
-
-/**
- * Utility to create a selection object representing a caret w/o anything selected
- * @param startEnd 
- */
-export function emptySelectionOption(startEnd: number): ModelEditSelection {
-    return { anchor: startEnd, active: startEnd };
-}
 
 export interface EditableModel {
     /**
@@ -87,9 +105,9 @@ export interface EditableModel {
 export interface EditableDocument {
     readonly selectionStart: number,
     readonly selectionEnd: number,
-    selection: { anchor: number, active: number },
+    selection: ModelEditSelection,
     model: EditableModel,
-    growSelectionStack: { anchor: number, active: number }[],
+    growSelectionStack: ModelEditSelection[],
     getTokenCursor: (offset?: number, previous?: boolean) => LispTokenCursor,
     insertString: (text: string) => void,
     getSelectionText: () => string,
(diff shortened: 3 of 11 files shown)


PEZ:
I got so confused by the names `rangeToSelectionStart/End`, so I renamed them ti `Left/Right` instead, because that is what we care about when growing the selection backward and forward. That name change made me see that their implementation were wrong as well, so that they worked in the editor was a bit of a coincidence.

Right now some things that used to work, doesn't work any longer, and we have new inconsistencies between editor and REPL window. I think the way to go is to continue to challenge the naming and thus the implementations in order to try remove some of the confusion and all of the inconsistencies between the two implementations.

@bpringe : It seems the bug I am trying to fix here has a lot of the causes in common with the bug you are fixing in #508. In this PR I have a bit of better coverage with Unit tests, and the naming change is really necessary. I suggest that if you get time to work on things before I do, that you base your fix off of this branch. Let's keep syncing!

bpringe:
@PEZ I agree that the naming change for consistency is much needed! This is what I was getting at in my PR, and the confusion I was referring to. It's well worth the effort and refactoring.

PEZ:
It's a bit funny, because I just had this flow and added lots of stuff and everything just worked. So I thought everything was dandy, but now it is obvious that there is some serious cleaning up needed. 😄 

I'm starting to think that we could get to a less confusing state by removing the `selectionStart/End` properties entirely and go via `selection` all the time. I'm not sure, because `selection` is just something I added a few days ago, but it seems it is less confusing than dealing with the mix of start/end, forward/backward, left/right, anchor/active...

bpringe:
Yeah I noticed some seemingly redundant properties like that, but didn't have the time to investigate much. I was thinking the same, going off of selection alone and removing the start/end ones would probably be a good idea.

bpringe:
I was just thinking... currently I think the selection is sort of represented in three places in the EditableDocument:
- selectionStart and selectionEnd
- selection
- growSelectionStack (which I think could be just called selectionStack)

Maybe we can have the selectionStack be the source of truth, and the selection is derived from it.

Furthermore we may be able to do away with the anchor and active concept and move to purely two-item arrays for selections, where (first selection) is always greater than (second selection) (to put it in clojure terms 😄). Let me demonstrate the idea briefly, if I can.

Here is an example selectionStack:

```clojure
[[1 2]
 [3 5]]
```
First the user selected from 1 to 2, then from 3 to 5. The selection is derived from the selection stack using a function like getSelection(), which returns `[(min (flatten selectionStack)) (max (flatten selectionStack))]`

The only place we need to be concerned about direction is when moving the cursor inside the forward/backward commands.

This would surely require rewriting some code, but I think a simplified model could simplify the code more, and be worth it in the long run.

bpringe:
There may be some holes in that model like possibly when selecting one way by one increment, then not undoing, but deselecting the other way by a different increment than the previous selection. This could possibly be solved though by not restricting the array numbers to the first one being less than the second. And instead they could sort of indicate direction implicitly like I think happens now in some places (IIRC).

So like a selectionStack of:

```clojure
[[1 2]
 [3 5]
 [5 4]]
```
Means the user selected, in two actions, forward from 1 to 5, then in another action, deselected from 5 backward to 4. In this model the current selection would be calculated as 

```clojure
[(first (first selectionStack) (last (last selectionStack))]
```

Which yields `[1 4]`. And then if needed, that could be sorted ^. This actually might be more intuitive.

PEZ:
Sounds good. Would be possible also to just have `{anchor: x, active: y}` objects on the stack, like we have today, right? It's similar to the semantics that the VS Code editor uses, which is why I introduced it.

PEZ:
Slowly inching my way there, @bpringe . I think my current unit testing tactics won't scale. But they have helped me fix so that paredit  for editor and repl window are bug compatible and to get lateral sexpr selections to work.

## Notes:

I think we could patch it so that all forward/backward selections work... but I also think it is better to take some more steps towards the selection stack being the single source of selection truth like you have thought up there.

One of my fixes involved using the vscode Selection `anchor` and `active` instead of it's `start` and `end` when mapping to EditableDocument selections. So that's one source of this whole confusing. VS Code Selections has both anchor/active and start/end, of which the latter maps to what I have starting to name left/end. Our replreadline document's start/end corresponds to the anchor/active of vscode Selections... gah. I am going to try get it such that we use these names:

* `anchor`, the start of a selection, can be left or right of, or the same as the end of the selection (active)
* `active`, the end of a selection, where the caret is, can be left or right of, or the same as the start of the selection (anchor)
* `left`, the smallest of `anchor` and `active`
* `right`, the largest of `anchor` and `active` 
* `backward`, movement towards the left
* `forward`, movement towards the right
* `up`, movement out of lists
* `down`, movement into lists

This will be in line with vscode when it comes to anchor/active, but introduce our own terminology  for the span of the selection. It will also keep the tradition of paredit with backward/forward and up/down.

If we succeed in getting it there, we can choose to then use start/end instead of left/right, but doing that now will cause chaos, because those currently mean something else in the REPL Window document...

bpringe:
Ah yes, this makes sense if VS Code semantics are anchor and active. And also keeping things the same as paredit convention is a good idea (forward/backward), those were the details I was lacking. But yes, I think we could still simplify the code in the future, and maybe as a first step you could add the above bullet points to the code somewhere in a comment, or to a readme somewhere, so that people working on that part of the code will see it.

PEZ:
I have implemented almost all of the changes now, except that the selection stack isn't yet the only source of truth. That will take some figuring, because today the actual selection in the editor is the source of truth for what is currently selected... We need to mirror that in a reliable way.

I have pasted the naming info as a comment in the code, and it is hopefully telling how things are named now, and not just how we want them to be named. 😄 

In the process I changed things to use more type checks than before.

I think it improves on what is currently released quite a bit so it might be time to merge. If you have some time, @bpringe, have a look and see if you agree.

PEZ:
And now I also pushed the right/left naming changes...

bpringe:
I like this a lot. I think it adds a lot of clarity, both the naming and the comments. I'm always taken aback by object orientated code though, i.e. the ModelEditSelection being a class with constructor, gets, and sets. But maybe it adds a little clarity in some places where used since it must be instantiated as `new ModelEditSelection`. 

Besides the tiny of bit logic for defaulting the anchor to active, I'm not sure if it provides much value. In most places I see it being passed an anchor, and where it's not, it might actually be beneficial to pass one for clarity at the site of the call. And if you were to make `anchor` required in the constructor, it makes the class basically pointless. Put another way, leaving it as it was and requiring it to be created as `{:anchor 1, :active 2}`, might be actually more explicit, clear, and simple. 

What I'm saying may be totally incorrect though! There may me more of a reason to use that class than I'm realizing.

PEZ:
Yes, maybe it retracts clarity rather than add it, which was my intention. I was going to encapsulate the selection stack inside there, was my plan, and I also thought it was good to be explicit about when the selection had the same anchor as active position. The latter could be more clearly expressed though, I ran into problems not being able to figure out how to have two constructors and took the easy way out... Let's remember that we have a little mess here to clean up.

PEZ:
I got fed up with not understanding what the paredit tests actually tested so I invented a small language for it, which is more precise and that I can understand quickly. Now the test tactics feel a bit more scalable to me, even. Please have a look, @bpringe , @kstehn , @cfehse .

kstehn:
That is a cool idea, but when you would describe it as a grammar you would see there is a small problem.
The symbol `|` has two meaning one it could be the position or it could be the range-boundray.
But when they are always exclusive then i think its fine :)

PEZ:
Good catch. I realised it is not exclusive, so changed the language some. Also distinguishing between ”empty” selection positions and ”empty" range positions.

PEZ:
The test results, running `npm run unit-test`, look like so:

```
  paredit
    movement
      rangeToSexprForward
        ✓ ><(def foo [vec]) => |(def foo [vec])|
        ✓ (><def foo [vec]) => (|def| foo [vec])
        ✓ (d><ef foo [vec]) => (d|ef| foo [vec])
        ✓ (def foo [:foo :bar :ba><z]) => (def foo [:foo :bar :ba|z|])
        ✓ (def>< foo [vec]) => (def| foo| [vec])
        ✓ (def foo ><[vec]) => (def foo |[vec]|)
        ✓ (def foo [:foo :bar :><baz]) => (def foo [:foo :bar |:baz|])
        ✓ (def foo [:foo :bar :baz><]) => (def foo [:foo :bar :baz|])
        ✓ (>def> foo [vec]) => (def| foo| [vec])
        ✓ (>def foo> [vec]) => (def foo| [vec]|)
        ✓ (<def foo< [vec]) => (def foo| [vec]|)
      rangeToSexprBackward
        ✓ (def <foo [vec]<) => (|def| foo [vec])
      moveToRangeRight
        ✓ (|def >|foo>| [vec]) => (def foo>< [vec])
        ✓ (|def <|foo<| [vec]) => (def foo>< [vec])
        ✓ (<def< foo >|[vec]>|) => (def foo [vec]><)
      moveToRangeLeft
        ✓ (def >|foo>| ><[vec]) => (def ><foo [vec])
        ✓ (def <|foo<| ><[vec]) => (def ><foo [vec])
        ✓ (<|def<| foo >[vec]>) => (><def foo [vec])
    selection
      selectRangeFromSelectionRight
        ✓ (def foo [:foo >|:bar>| <:baz<]) => (def foo [:foo <:bar :baz<])
      selectRangeFromSelectionLeft
        ✓ (def foo [:foo >:bar> >|:baz>|]) => (def foo [:foo >:bar :baz>])
    selection stack
      ✓ should make grow selection the topmost element on the stack
      ✓ get us back to where we started if we just grow, then shrink
      ✓ should not add selections identical to the topmost
      ✓ should have A topmost after adding A, then B, then shrinking

  string
    keywordize
      ✓ keywordizes non-keywords
      ✓ leaves keywords alone
    unKeywordize
      ✓ un-keywordizes keywords
      ✓ leaves non-keywords alone
```

PR #544, opened 2020-01-11
Cljfmt indent rules

### LINKED ISSUE #80: Custom indentation for some functions
With Cursive it is possible to configure the indentation for each function individually. Is it possible to do the same with Calva?

nbardiuk:
## What has Changed?

- new configration for cljfmt indentation rules `calva.fmt.indentationRules`

Configuration format is the same as https://github.com/BetterThanTomorrow/calva/blob/master/src/cursor-doc/indent.ts#L7-#L76

Example configuration:
```json
"calva.fmt.indentationRules": {
    "comment": [["inner", 0]],
    "let-system": [["inner", 0, 1], ["block", 1]]
},
```

With the rules
```clojure
(comment "comment"
  (prn "too far"))

(let-system [a (resolve-a)]
  (prn a))
```

Default behavior 
```clojure
(comment "comment"
         (prn "too far"))

(let-system [a (resolve-a)]
            (prn a))
```

Semantics of user's rules is the same as in leiningen plugin. 
The user's configuration is merged on top of defaults. 
User don't have to specify all rules and can override existing ones.

Closes #80 

## My Calva PR Checklist

I have:


nbardiuk:
I have doubts about regex syntax of cljfmt. For example https://github.com/weavejester/cljfmt/blob/master/cljfmt/resources/cljfmt/indents/fuzzy.clj

`edn` is more expressive format than `json` so we cannot simply express the same configuration in vscode settings as one can in cljfmt edn. If/when calva will support regex matching for indentation there will be problem to express them in vscode configuration.

Example edn setting
```edn
{ #"^with-" [[:inner 0]]
  comment [[:inner 0]]}
```

Calva will have to decide how to `tag` rules with regexes in json format. 
I can see 4 options:

1. Using custom prefix for regex rules
for example treat `#` as a tag that the rest is regular expression, otherwise it is a symbol
```json
{ "#^with-": [["inner", 0]],
  "comment": [["inner", 0]]}
```

2. Using custom prefix for literal rules
for example treat `'` as a tag that the rest is a symbol, otherwise it is a regex
```json
{ "^with-": [["inner", 0]],
  "'comment": [["inner", 0]]}
```

3. Using custom rule that tags regex
for example treat `["re"]` as a tag that the key is regex
```json
{ "^with-": [["re"], ["inner", 0]],
  "comment": [["inner", 0]]}
```

4. Have separate configurations for symbols and regexes

```json
"calva.fmt.indentationRules": {
    "comment": [["inner", 0]]
},
"calva.fmt.indentationRulesRegex": {
    "^with-": [["inner", 0]]
}
```

I prefer the last one. But regardless of what calva decides it would be good to think about naming now.

PEZ:
We could also read the rules from an `EDN` file, which also already exists in many projects out there already.

PEZ:
Many thanks for starting this, btw! It is really lacking and I think stopping Calva from being used in some projects. I'll check your changes a bit now and see if I have some concrete feedback to give, but from the surface things look great and also very good that you highlight the regex case.

nbardiuk:
I agree that reading config file from project folder is the most flexible and portable option. We can avoid conversion between js and cljs and read file directly in cljs.

Should I create something like `calva.fmt.configPath` that points to `./cljfmt.edn` file instead of json object?

Although I don't know how we can read it in typescript to extend rules in https://github.com/BetterThanTomorrow/calva/blob/master/src/cursor-doc/indent.ts#L7

PEZ:
> Should I create something like calva.fmt.configPath that points to ./cljfmt.edn file instead of json object?

Yeah, maybe that makes the most sense.

PEZ:
About those other indent rules. Not sure you've figured out why we have those, but anyway, there are two indenters in Calva, where this TokenCursor one is meant to be the one used for the format-as-you-type case.

Today they indent the same way because the `TokenCursor` indenter (the rules you linked to) uses `cljfmt` defaults, and we do not have a way to customise the other formatter. Which is a bit brittle, but with this PR we can sort that out so that they both use the same config instead.

You can probably use the `parseEdn()` utility function exposed by `calva-lib`. Check `project-types.ts` for how it is used today.

PEZ:
In the *nice to have*. department we can also honor `cljfmt` settings found in the Leiningen project file.

PEZ:
I now found out why the build fails. `./src/calva-fmt/src/config.ts` is used by the REPL Window webview, which doesn't have access to `vscode` things.

nbardiuk:
I've created another PR #546 because there was almost nothing to reuse from this one

PR #557, opened 2020-01-31
Guard against tokenizing strings > 20K

### LINKED ISSUE #556: Calva locking up when opening files with very long lines
When opening a file with very long lines all Calva commands stop working.

A symptom is that even just hitting backspace stops doing what it should. Another is that the CPU gets very a lot of load.

I think this is because that the tokenizer (probably the `TokenCurosr`, but it could also be the `Calva highlight` tokenizer) locks up trying to split the line up with its regular expressions.

A ”solution” is to do what the VS Code built-in syntax highlighter does: Recognise that tokenisation will take a long time and just skip doing it then.

PEZ:
Work in progress...

## What has Changed?
I've put a guard in `lexer.ts` to guard against tokenization of long lines. Right now using 20K length hardcoded, but probably should be using vscode's tokenization maxlength setting.

The current approach is to recognize when a line is too long and only represent it as a `too-long-line` token.

Fixes #556

## My Calva PR Checklist
I have:

### DIFF
--- a/src/webview.ts
+++ b/src/webview.ts
@@ -4,7 +4,10 @@ var Ansi = require('ansi-to-html');
 import "../assets/styles/webview.scss";
 import escapeHTML = require("escape-html");
 import * as paredit from "./cursor-doc/paredit";
-import { ModelEdit, ModelEditSelection } from "./cursor-doc/model";
+import { ModelEdit, ModelEditSelection, initScanner } from "./cursor-doc/model";
+
+const MAX_LINE_TOKENIZATION_LENGTH = 20000;
+initScanner(MAX_LINE_TOKENIZATION_LENGTH);
 
 declare function acquireVsCodeApi(): { postMessage: (object: any) => void }
 const message = acquireVsCodeApi();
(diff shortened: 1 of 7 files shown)


bpringe:
Sounds like a good approach. I agree that using VS Code's setting would probably be ideal.

PEZ:
Now I fixed a proper scanner object for tokenising the too-long-strings.

I can't figure out how to get hold of the VSCode setting for tokenisation max-length though...

PEZ:
Now added an `initScanner` function that needs to get called before lexing anything. It's a bit messy, but I think it is the right place to pass in the vscode max tokenisation setting.

src/webview.ts
```diff
@@ -4,7 +4,10 @@ var Ansi = require('ansi-to-html');
 import "../assets/styles/webview.scss";
 import escapeHTML = require("escape-html");
 import * as paredit from "./cursor-doc/paredit";
-import { ModelEdit, ModelEditSelection } from "./cursor-doc/model";
+import { ModelEdit, ModelEditSelection, initScanner } from "./cursor-doc/model";
+
+const MAX_LINE_TOKENIZATION_LENGTH = 20000;
```
bpringe: Can we also use the vs code setting here? And do we want to?

PEZ: It would be nice to do that, but it wasn't immediately obvious to me how to do it.

PEZ:
We will have to make it prettier later. This tokenization freeze is causing too much troubles out there.

PR #576, opened 2020-03-05
Tweak handling of nil, true, and false in grammar

### LINKED ISSUE #577: `nil` followed by comma not highlighted correctly
`nil` by itself or followed by a closing delimiter is highligted correctly.

However, `nil` followed by a comma is not.

Working on a [PR](https://github.com/BetterThanTomorrow/calva/pull/576) to address this.

sogaiu:
`nil` followed by a comma was not being handled approriately for highlighting. #577

`true` and `false` had some issues too. #578

Handling of `nil`, `true`, and `false` should be improved.

The test for `nil` and that of booleans was also added to.

I have:

- [x] Made sure I changed the default PR base branch, so that it is not `master`. (Sorry for the nagging.)
- [X] Tested the VSIX built from the PR (well, if this is a PR that changes the source code.) You'll find the artifacts by clicking _Show all checks_ in the CI section of the PR page, and then _Details_ on the `ci/circleci: build` test. (For now you'll need to opt in to the CircleCI _New Experience_ UI to see the Artifacts tab, because bug.)
     - [X] Smoke tested the extension as such.
     - [X] I have used [GitHub's fixes/closes syntax](https://help.github.com/en/articles/closing-issues-using-keywords)

Fixes #577 #578

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Changes to Calva.
 ## [Unreleased]
 - Fix so that Calva treats symbol containing the quote character correctly.
 - [Fix: Parameter hints popup should be off by default](https://github.com/BetterThanTomorrow/calva/issues/574)
+- [Fix: `nil` followed by comma not highlighted correctly](https://github.com/BetterThanTomorrow/calva/issues/577)
+- [Fix: The syntax highlightning fails with symbols named truesomething/falsesomething](https://github.com/BetterThanTomorrow/calva/issues/578)
 
 ## [2.0.79] - 2020-03-01
 - Use scope `variable.other.constant` for keywords, making them highlight nicely
--- a/clojure.tmLanguage.json
+++ b/clojure.tmLanguage.json
@@ -68,11 +68,11 @@
 		"constants": {
 			"patterns": [
 				{
-					"match": "(nil)(?=(\\s|\\)|\\]|\\}))",
+					"match": "(?<=^|\\s|,|\\(|\\[|\\{)(nil)(?=(\\s|,|\\)|\\]|\\}|\\^))",
 					"name": "constant.language.nil.clojure"
 				},
 				{
-					"match": "(true|false)",
+					"match": "(?<=^|\\s|,|\\(|\\[|\\{)(true|false)(?=(\\s|,|\\)|\\]|\\}|\\^))",
 					"name": "constant.language.boolean.clojure"
 				},
 				{
--- a/src/calva-fmt/atom-language-clojure/grammars/clojure.cson
+++ b/src/calva-fmt/atom-language-clojure/grammars/clojure.cson
@@ -97,11 +97,11 @@
   'constants':
     'patterns': [
       {
-        'match': '(nil)(?=(\\s|\\)|\\]|\\}))'
+        'match': '(?<=^|\\s|,|\\(|\\[|\\{)(nil)(?=(\\s|,|\\)|\\]|\\}|\\^))'
         'name': 'constant.language.nil.clojure'
       }
       {
-        'match': '(true|false)'
+        'match': '(?<=^|\\s|,|\\(|\\[|\\{)(true|false)(?=(\\s|,|\\)|\\]|\\}|\\^))'
         'name': 'constant.language.boolean.clojure'
       }
       {
--- a/src/calva-fmt/atom-language-clojure/spec/clojure-spec.coffee
+++ b/src/calva-fmt/atom-language-clojure/spec/clojure-spec.coffee
@@ -84,9 +84,42 @@ describe "Clojure grammar", ->
       for bool in bools
         {tokens} = grammar.tokenizeLine bool
         expect(tokens[0]).toEqual value: bool, scopes: ["source.clojure", scope]
+        {tokens} = grammar.tokenizeLine " " + bool
+        expect(tokens[1]).toEqual value: bool, scopes: ["source.clojure", scope]
+        {tokens} = grammar.tokenizeLine bool + " "
+        expect(tokens[0]).toEqual value: bool, scopes: ["source.clojure", scope]
+        {tokens} = grammar.tokenizeLine "," + bool
+        expect(tokens[1]).toEqual value: bool, scopes: ["source.clojure", scope]
+        {tokens} = grammar.tokenizeLine bool + ","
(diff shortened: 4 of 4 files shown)


sogaiu:
I modified the CHANGELOG.md file after testing, but did not retest.

PEZ:
The tests wouldn't catch a mistake there, anyway. It would just stop the release pipeline, which isn't run on a PR submit. FYI.

PEZ:
Thanks!!!

Had a quick look: Two things:

1. I forgot to mention a thing about this grammar thing: The script for actually updating the file that VS Code uses. I updated the [How to Hack on Calva](https://github.com/BetterThanTomorrow/calva/wiki/How-to-Hack-on-Calva/_edit#updating-the-tmlanguge-grammar) docs with this now... Sorry!
1. I haven't tested this yet, but it does look like something like `nilla` would be highlighted *nil*la, which I don't think we want.

Cheers!

sogaiu:
Will take a look at the script to update the file that VS Code uses.

About `nilla` - then wouldn't similar logic apply to `truea` and `falsea`?

sogaiu:
BTW, the link in your most recent comment to the 'How to Hack on Calva` doc points to a version for editing it.

sogaiu:
Hmm, feeling dense -- not seeing the changes to the `How to Hack on Calva' doc regarding the update script for the grammar.  Any hints?

sogaiu:
Ok, so may be the original pattern for `nil` was actually closer to what is desired.  Though it appears to be missing the case of having a comma come after them.

Should I update all three to account for commas too?

sogaiu:
It appears there are more patterns that don't account for using comma as whitespace.  But testing suggests that commas don't need to be handled in the regular expression pattern.  Is that correct?

May be they are accounted for via some other means?

sogaiu:
Very strange that the behavior of `nil` is different from `true` and `false`.

sogaiu:
May be you don't want to handle the following, but FWIW:
```
$ clj -r
Clojure 1.10.1
user=> nil^{:hi 1}[]
nil
[]
```
Note that there is only one space in that sequence being read -- between the `i` and the `1`.

PEZ:
I'd like that to work. I notice that the token cursor grammar doesn't handle it either, which I think might be a very recent bug introduced by me.

Speaking about the token cursor grammar. It avoids parsing `nila/trua/falsea` as two tokens by the `id` token just being longer than the `lit`(eral) `nil/true/false` tokens. That is obviously not how the tmGrammar deals with things. Maybe it can help us look at the token cursor tokenization of `id` tokens. The regex there ends with ``[^()[\]\{\},~@`\"\s;]*``. Please note that the `^` character is missing there, because I removed it the other day, and that is probably the source of the error with the token cursor failing with `nil^`.

And about the failing build. It has nothing to do with your changes. It is the recently added property based testing that happened to find a problem with how the token cursor's tokenization of keywords.

PEZ:
> Should I update all three to account for commas too?

Yes, please do, if you have the time. I created an issue for it: #578

sogaiu:
The line containing the regex in the token cursor grammar is the following?

https://github.com/BetterThanTomorrow/calva/blob/master/src/cursor-doc/clojure-lexer.ts#L83

PEZ:
Yes. That's the one. And that scanner tries all regexes over the text and then picks the longest one. (And the first of terminals, if there are several longest ones.)

sogaiu:
It may have been clear, but FWIW I mentioned the ^ thing because I was working on enumerating what could come after `nil`.

The obvious things are:

* whitespace `\s`
* right paren `\)`
* right bracket `\]`
* right brace `\}`

The tm grammar file tries to account for these in multiple places, but apparently not consistently.

Slightly less obvious is comma `,` (This tripped me up when working on an externals item in tree-sitter.  I was skipping whitespace but not commas.)

Since you want to support the aforementioned case of:
```
nil^{:hi 1}[]
```
I guess caret `^` should be accounted for as well.

Neither of those seem to be accounted for in the places I looked in the tm grammer.

Then there is the matter of what can come before `nil` (and `true` and `false`) -- which perhaps is only different for delimiters.

I guess there might be other things too...

Does one need to account for end of line and beginning of line too?  May be end of line is accounted for by `\s`.

PEZ:
I think that token cursor grammar list of characters should be used as a start (plus caret). And that we should strive for consistent application of it across the tmGrammar and the token cursor grammar. What has guided me so far in these things are things that the Clojure reader has shown me. So whenever it reads things differently than my grammar (speaking mostly about the token cursor here, because that's where I've been digging lately), I've tried to fix the grammar. I guess it could be stated as: _The reader is my source of truth._

Not sure how newlines are treated in the tmGrammar. But if you use the token cursor grammar for inspiration/consistency know that it parses each line separately. So even if it might look like it accounts for newlines, it actually doesn't. (There is residue there from when I had misunderstood things a bit regarding this.)

sogaiu:
Thanks for the explanation.

Those things mostly make sense to me.

Is there a way to run the token cursor grammar from the command line?

PEZ:
> Is there a way to run the token cursor grammar from the command line?

There probably is. Will involve figuring how to run TS code, or maybe you can run the JS that webpack builds... In any case, looking at the tests is a way to see the simplest way I have found to run the grammar: https://github.com/BetterThanTomorrow/calva/blob/dev/src/extension-test/unit/cursor-doc/clojure-lexer-test.ts

sogaiu:
Thanks for the tip -- I'll take a look.

sogaiu:
Perhaps it is related to the way `scan` works, but the following tests currently pass:

```
    {tokens} = grammar.tokenizeLine ",nil"
    expect(tokens[1]).toEqual value: "nil", scopes: ["source.clojure", "meta.symbol.clojure"]

    {tokens} = grammar.tokenizeLine "nil,"
    expect(tokens[0]).toEqual value: "nil", scopes: ["source.clojure", "constant.language.nil.clojure"]
```

I presume the `meta.symbol.clojure` should really be `constant.language.nil.clojure`.

PEZ:
> I presume the `meta.symbol.clojure` should really be `constant.language.nil.clojure`

Agreed.

PEZ:
Thanks! This looks great. Once the branch is up-to-date with latest `dev` I'll pull this in.

PR #588, opened 2020-03-17
fix link pointing to built-in sequence

jaimesangcap:
Fixed broken link


bpringe:
Thanks @jaimesangcap! Would you mind branching this from dev instead of master, and pointing the merge to dev? We try to keep new things, even small fixes to docs like this, going into the dev branch,

PEZ:
The release workflow is a bit complicated b/c reasons... I think we got this thing merged right now. (If, not, I'll make sure we do. 😄 )

PR #601, opened 2020-04-05
Always send along `ns` w/ nrepl eval

PEZ:
## What has Changed?

- All uses of the nrepl client `eval()` op now are required to pass `ns`.
- Many uses send along the current client `ns`, as that makes more sense than sending `user`.

## My Calva PR Checklist
I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 Changes to Calva.
 
 ## [Unreleased]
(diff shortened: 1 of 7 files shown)


PEZ:
This was discovered using the new `babashka` nREPL server, which doesn't yet allow for a missing `ns` param: https://github.com/borkdude/babashka/issues/331#issuecomment-609434812

PR #616, opened 2020-04-18
WIP - Add debug instrument command

bpringe:
## What has Changed?

- Add command to instrument a form. This adds `#dbg` before the form before eval'ing it, but does not add it to the editor.
- Add decorations that put a border around the symbol name for instrumented defs
- Apply decorations on connect, on eval, on text document change, on editor change, and clear the decorations on disconnect

image

## My Calva PR Checklist

I have:


bpringe:
@PEZ I'm not sure on a reasonable delay(debounce) time for these decorations. They will occur each time a user types in the editor (on text document change), so I didn't want to put it too low, but it's also not a lot of work either and doesn't seem to be an issue.

PEZ:
Sorry, I missed that question, @bpringe . Have you figured out what you want to do about it?

bpringe:
Yes, I can set a reasonable default. I can visually see our debounce time compared to the clj-kondo's debounce time (and could check the clj-kondo code base if desired), and since I'm using clj-kondo now, I think it makes sense to set it close to the same.

bpringe:
@pez My decision to use clj-kondo as a jvm lib, loaded into the repl at jack-in like cider-nrepl, is because this seemed like the best choice for performance, since the linting is called on document change.

borkdude recommended I use it "as a jvm lib" when I mentioned what I was doing, and I'm guessing this is what he meant. I don't think it would be as performant as a command line call, though I have not tested that (and just guessing we could use it that way). I don't really see a downside to loading it into the repl though.

I added a warning message that will show on connect if clj-kondo is not found on the classpath (if the `require` fails), to say that debug decorations will not be enabled. Do you think I should leave this or remove it? I'm not sure if it's a good idea.

src/nrepl/project-types.ts
```diff
@@ -152,11 +152,13 @@ async function leinProfilesAndAlias(defproject: any, connectSequence: ReplConnec
...
     CIDER_NREPL_VERSION = "0.23.0",
     PIGGIEBACK_VERSION = "0.4.2",
-    FIGWHEEL_MAIN_VERSION = "0.2.3";
+    FIGWHEEL_MAIN_VERSION = "0.2.3",
+    CLJ_KONDO_VERSION = "2020.04.05";
 
 const cliDependencies = {
     "nrepl": NREPL_VERSION,
     "cider/cider-nrepl": CIDER_NREPL_VERSION,
+    "clj-kondo": CLJ_KONDO_VERSION
```
PEZ, line 161: This should be added to lein and shadow dependencies as well, right?

Also, I wonder what the user should do if the project already has clj-kondo dependencies. Maybe we need to add an opt-out of Calva injecting it?

src/debugger/decorations.ts
```diff
@@ -0,0 +1,37 @@
+import * as vscode from 'vscode';
+import * as util from '../utilities';
+
+const instrumentedFunctionDecorationType = vscode.window.createTextEditorDecorationType({
+    borderWidth: '1px',
+    borderStyle: 'solid',
+    overviewRulerColor: 'blue',
+    overviewRulerLane: vscode.OverviewRulerLane.Right,
+    light: {
+        // This color will be used in light color themes
```
PEZ, line 12: Minor, minor, but this comment isn't really necessary. =)

bpringe: Carried them over from an example, will remove =)

src/nrepl/project-types.ts
bpringe: Oh yeah, I forgot to go back and do that. Good catch, lol. I don't think it's going to affect anything if we inject it and it's already there, though I'll do some testing like adding a different version to the project deps and see what happens.

bpringe: Added it to the lein dependencies. I'm still unsure that it needs to go into shadow dependencies since that should just be cljs deps right? Since the debugger is just for clojure I figured that was not relevant.

bpringe: Also fyi, I tested with lein jack-in and a diff version of clj-kondo. I'm unsure which version (the injected one or the project one) gets actually used though. It seems unlikely any issues will arise from this, so I'm not sure that an opt-out makes sense to add. But maybe we can just keep it in mind and add it later if needed? I could be missing something though.

bpringe: Also I think if they wanted to opt out they could start their own repl aside from jack-in? This may be the way to go at least for now.

PEZ: A middle ground would be to have the clj-kondo version as a setting.

PR #684, opened 2020-07-21
Output to tmp `.calva-out` document

### LINKED ISSUE #681: Send evaluation results and output to a file
When evaluating forms the output should be sent to a file where Calva editing and highlighting is enabled.

This should probably replace the current behaviour with printing these in the **Calva says**  channel.

PEZ:
## What has Changed?

- Opens a tmp document at REPL connection
- Prints evaluation results and stdout/err to this document

Fixes #681

TODO:
- [x] There are still problems with printing several times during an eval.
- [x] If the output doc is visible it should be scrolled immediately
- [x] Stop printing to Calva says.
- [x] Stop decorating the hover with the results.
- [x] Possibly, open the output doc where it was last placed
- [x] Add command for revealing the output doc
- [x] Stop clj-kondo from linting the output doc
- [x] Make it possible to switch between the CLJ and CLJS REPLs for this file (when used as a REPL).
- [x] Print on a new line when last line is not empty
- [x] Print stack trace in the output window, when available
- [x] Change evaluation shortcuts to something easy (like `alt+enter` for top level)
- [x] Add some usage help to the window greetings
- [x] Add a current namespace to the window
- [x] Stop popping up *Calva says*
- [x] Advice users to leave the output window open
- [x] Stop opening the REPL window by default
- [x] Evaluate connect sequence `afterCLJReplJackInCode` to the output repl-file
- [x] Prepare a version of calva.io with info about this output/repl file
- [x] Avoid repl loading the output window

## My Calva PR Checklist
I have:


PEZ:
@bpringe I'm running in to troubles when evaluating code like

```clojure
(do (println "foo") (println "bar"))
```

Where only `"foo"` gets printed. I think it is that the document is being edited when the next edit `apply` happens, but I do not know how to fix it.  The `"bar"` printing ends up here: https://github.com/BetterThanTomorrow/calva/pull/684/commits/b2c82f92eb672f40d26d97c3807bc7cd6b1abb3d#diff-8d9d1fa036c5755eab055f777e354264R67

If you have an idea, please help. ❤️ 😄

bpringe:
You caught me at a good time because I've started to get an itch for Calva work lately... :smile:. I've pulled it down and started trying it out. Very exciting!

I'll play around with it. I replicated the issue you described above with println and also noticed that with a single println, sometimes the result, `nil`, is printed in the results file (as it should be), but many times it's not. Seems related. I'll let you know if I find something,

bpringe:
Is there a way to set an editor/file to wrap text? I think at least when pprint is turned off, the text should wrap. When pprint is on, I think we may need to adjust the pprint line length to the width of the editor, at least to some reasonable lower limit.

PEZ:
I think wrapping is a general setting. Not something we can configure for one file.

PEZ:
If you evaluate something w/o side effect output, do you still sometimes get no output?

bpringe:
Seems like I always see the output if no side-effect output.

bpringe:
I pushed my changes here. Added a queue for edits and made it apply them one at a time. I want to refactor it more though because I'm not sure if all the logic surrounding the applyEdit call really needs to be in the repeat/close-together calls, like opening the editor.

PEZ:
Thanks!!! It works like a charm. The unpredictable order of prints, that I got with my retry-in-a-jiffy hack, is gone.  I'm so happy now!

bpringe:
One thing to consider here is maybe a max number of lines. What happens when there is some long running process logging to this file? For example, I used to work with data pipelines that might process thousands of messages and log at different steps in the pipeline. This could be left running for hours.

Should we cap the size or number of lines of the file? And how would that work? Would we truncate lines near the top as we reach a limit?

PEZ:
When working with those pipelines, did you observe how the current output channel handled it?

At some amount of output, vscode will stop syntax highlighting the file. And at some point Calva will also stop with its extra highlighting.

That said, already as it is, Calva's highlighting takes a long time on large output structures, so this should be optimised a bit.

glebovmaksim:
Hi! My 2 cents about new output window:

1. After loading current file (`ctrl+alt+c enter`) the output shows in a repl file (which is a good thing), but **Calva says** window still popping up.
2. This one is a little bit tricky. After closing a repl file (which was in a separate editor group) and evaluating some expression inline, a repl file automatically opens in a tab next to mine (same editor group), but have no focus. If I do NOT click on this tab and instead click on an "Open Results Window" link in a result hover, a new (second) repl file tab shows in a new editor group with focus on it. Seems like a strange behaviour to me.

PR #785, opened 2020-09-19
Allow toggling keyboard shortcuts

### LINKED ISSUE #784: Allow toggling keyboard shortcuts
I like the default Paredit keymap, but I find myself needing to use the standard VSCode editing shortcuts a lot as well. For example, when editing a comment or a string, I would like to move by word using the `Ctrl+<ArrowKey>` shortcuts.

I have implemented an additional config option and keyboard shortcut to allow toggling keyboard shortcuts on or off, so users can easily toggle -- similar to modal editing in Vim.

PR incoming! :rocket:

wcerfgba:
## What has Changed?
Closes #784

We introduce a `calva.paredit.keybindingsEnabled` config option and a
`paredit.toggleKeybindingsEnabled` command which allows toggling Paredit
keybindings on and off while preserving the current mode. The Paredit
keybindings are updated to only activate when the
`paredit:keybindingsEnabled` context key is true, in addition to the
existing conditions on the keymap. Finally, the Paredit StatusBar
component is updated to receive a `KeyMapConfig` object containing
`keyMap` and the new `keybindingsEnabled` boolean, and a new path is
added in the UI state management to handle this case as distinct from the
'none' keyMap case.

Inside the Paredit `StatusBar` component, I factored out the code for setting properties on the `_toggleBarItem` into a `updateUIState` fn, to indicate a separation of concerns between the `keyMapConfig` setter and updating the UI: previously the statements in the switch cases were updating the internal `_keyMap` value and also updating the `_toggleBarItem`. I have not touched `visible` setter.

Tested in VSCode 1.48.1 on Arch Linux.

## My Calva PR Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Changes to Calva.
 ## [Unreleased]
 - [Fix: evals should be ignored during parsing](https://github.com/BetterThanTomorrow/calva/issues/763)
 - Fix: [Test runner can't find tests under cursor when using a custom test macro](https://github.com/BetterThanTomorrow/calva/issues/786)
+- [Allow toggling Paredit function independently of mode](https://github.com/BetterThanTomorrow/calva/issues/784)
 
 ## [2.0.124] - 2020-08-31
 - Re-fix: [Can't jack-in when no project file is open](https://github.com/BetterThanTomorrow/calva/issues/734)
--- a/package.json
+++ b/package.json
@@ -504,6 +504,12 @@
                             "none"
                         ],
                         "scope": "window"
+                    },
+                    "calva.paredit.keybindingsEnabled": {
+                        "type": "boolean",
+                        "description": "Activate Paredit keybindings.",
+                        "default": true,
+                        "scope": "window"
                     }
                 }
             },
@@ -753,6 +759,11 @@
                 "title": "Toggle Paredit Mode",
                 "when": "paredit:keyMap =~ /original|strict/",
                 "enablement": "paredit:keyMap =~ /original|strict/"
+            },
+            {
+                "category": "Calva Paredit",
+                "command": "paredit.toggleKeybindingsEnabled",
+                "title": "Toggle Paredit Keybindings Enabled"
             },
             {
                 "category": "Calva Paredit",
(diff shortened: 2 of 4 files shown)


PEZ:
Hello. Thanks for this PR! I had to ponder this a bit to decide whether I thought it is a good idea or not. My default reaction to new settings is that they are bad. 😄 (Only half joking here, as @bpringe can tell you). Anyhow, I've decided that this might actually be a good thing to support: To quickly stove away some troublesome shortcuts and then switch them back back in again. I'll let @bpringe chime in as well.

Here are some thoughts:

I wonder about whether this really is a paredit thing, or rather a Calva in general thing. The problem is that some of Calva's shortcuts are overriding shortcuts that someone might be used to using for something else, right? That it is Calva Paredit that register most of those is due to the fact that Paredit registers the bulk of Calva's shortcuts, and that some of the Paredit shortcuts are of the ”simple” type, with high risk for conflict. So maybe we should name the option ”calva keyboard shortcuts enabled” (or some such) and let the command toggle them all?

Also the current name makes it sound a bit like you are enabling/disabling paredit, which you are not, just the shortcuts. (I have both glanced at the code changes and tried it, so I know that Paredit is still enabled when this option is disabling the shortcuts 😄). 

When toggling the shortcuts off, the paredit button in the status bar shows this nicely. But I can't toggle it back on using that button, which seems a bit odd to me. But I also see how this button isn't used to disable the shortcuts, so it makes sense too. 😄 Your thoughts on this, @wcerfgba and @bpringe ?

As for the code changes, I think they look clean and also improves the code a bit, Kaizen style.

wcerfgba:
Hi @PEZ , thanks for your detailed reply to my PR. :smiley: I wrote this just because it is something I wanted and this solution solves my own use case, but I'm happy to work with you and others to revise this to make something that is more useful for other users. :smile: 

> So maybe we should name the option ”calva keyboard shortcuts enabled” (or some such) and let the command toggle them all?

This sounds like a good idea. For me, I haven't found any other conflicts with any other features in Calva, the Paredit shortcuts are the big one for me, but this idea is more in line with the intention of the PR. I guess there are some edge cases where a user may want to still eg boot a REPL, evaluate a form, or format a form, even though they want to disable the Paredit keybindings, but I am not sure how common this would be, and I think it would make more sense to make all of the Calva shortcuts modal to start, and then more granular toggling of keybindings for specific features could be introduced if there were requests for this?

> Also the current name makes it sound a bit like you are enabling/disabling paredit, which you are not, just the shortcuts.

Great spot, how about "Toggle Paredit Keyboard Shortcuts" or "Toggle Paredit Keybindings" (or "Toggle Calva Keyboard Shortcuts" or "Toggle Calva Keybindings" subject to earlier discussion on modality)?

> When toggling the shortcuts off, the paredit button in the status bar shows this nicely. But I can't toggle it back on using that button, which seems a bit odd to me.

Yes, I did not want to overload or change the behaviour of this button in the PR, only provide a path in the code to render the UI differently for this case as distinct from the 'keyMap none' case, and make it somewhat obvious what is happening via the tooltip -- ideally we could colour the lambda differently or use a different symbol or something to indicate that the keyboard shortcuts are disabled. If we are going to change this to be a toggle for all keyboard shortcuts, we could add another separate toggle/indicator to the main statusbar component, to sit alongside the `prettyPrintToggle`?

> As for the code changes, I think they look clean and also improves the code a bit, Kaizen style.

Thanks! :smiley_cat: If we decide to generalise this change to cover all keyboard shortcuts I can include a separate commit which keeps the extracted `updateUIState` fn in Paredit statusbar but preserves the original behaviour. :metal:

docs/readthedocs/source/paredit.md
```diff
@@ -122,6 +122,8 @@ Care has been put in to making the default keybindings somewhat logical, easy to
 
 Note: You can choose to disable all default key bindings by configuring `calva.paredit.defaultKeyMap` to `none`. (Then you probably also want to register your own shortcuts for the commands you often use.)
 
+You can toggle Paredit on and off using the `paredit.toggleenabled` command. This allows you to switch modally between Paredit keybindings and standard textual keybindings.
```
bpringe, line 125: Just a note here, the current docs are now in the calva-docs repo, and should be updated there. We need to see about removing these old docs so as not to cause confusion.

bpringe:
Thanks for taking the time to make this PR. It's well done. I do agree with @pez that it should reflect that it's disabling/enabling the keybindings and not paredit itself. I also think it's best to keep it at just the paredit shortcuts, so the user can keep them off if desired and still do things like jack-in and evaluate forms.

Is it best to have a toggle button for this in the status bar? I wonder the frequency at which users would actually need to toggle it. If infrequent, then it makes sense to just have it as a checkbox in settings. If you find yourself wanting to toggle it often, then the status bar button makes sense.

docs/readthedocs/source/paredit.md
wcerfgba: Noted, I will remove this change from the PR. I can submit another PR if you like which deletes all the docs and updates the PR template to point people towards calva-docs instead if you'd like?

wcerfgba:
> I do agree with @PEZ that it should reflect that it's disabling/enabling the keybindings and not paredit itself.

Roger, will update. :+1: 

> I also think it's best to keep it at just the paredit shortcuts, so the user can keep them off if desired and still do things like jack-in and evaluate forms.

Interesting. I am happy to leave it as it is, since this is fine for my workflow (and means I don't have to change the PR too much :stuck_out_tongue: ) so long as you are both happy with it from a philosophical and UX perspective. :smiley: 

> Is it best to have a toggle button for this in the status bar? I wonder the frequency at which users would actually need to toggle it. If infrequent, then it makes sense to just have it as a checkbox in settings. If you find yourself wanting to toggle it often, then the status bar button makes sense.

Personally I want to toggle it often and easily, so I would use the keybinding I assigned to the command, rather than clicking a statusbar button, but I think it is nice to provide a visual indication of the modality -- cf vi with no indication of insert/command mode and vim where it indicates `--INSERT--` in the statusbar. :smiley:

PEZ:
>> I also think it's best to keep it at just the paredit shortcuts, so the user can keep them off if desired and still do things like jack-in and evaluate forms.

> Interesting. I am happy to leave it as it is, since this is fine for my workflow (and means I don't have to change the PR too much 😛 ) so long as you are both happy with it from a philosophical and UX perspective. 😃

Yeah interesting. 😄 If I understand the use case correctly this is to _very_ temporarily move shadowing shortcuts away, while the user does a few things and then they would be switched back again. The thing is, we don't know which shortcuts are in the way. Maybe some extension uses `ctrl+enter` or `alt+enter` for something? Then moving just the paredit shortcuts away wouldn't do the trick. All or none, is my take on it.

>> Is it best to have a toggle button for this in the status bar? I wonder the frequency at which users would actually need to toggle it. If infrequent, then it makes sense to just have it as a checkbox in settings. If you find yourself wanting to toggle it often, then the status bar button makes sense.

> Personally I want to toggle it often and easily, so I would use the keybinding I assigned to the command, rather than clicking a statusbar button, but I think it is nice to provide a visual indication of the modality -- cf vi with no indication of insert/command mode and vim where it indicates --INSERT-- in the statusbar. 😃

I think this is not a very common need. Giving it a place of its own in the status bar is a bit too much. A keyboard shortcut is perfect. It was just that the paredit mode icon reflected the disablement that got me to want to use it to toggle back. I think it just leaving that icon displaying the keymap info is OK. Grey it out, but keep it toggling between keymaps.

bpringe:
>The thing is, we don't know which shortcuts are in the way. Maybe some extension uses ctrl+enter or alt+enter for something? Then moving just the paredit shortcuts away wouldn't do the trick. All or none, is my take on it.

I see what your point here, and I agree.

>Giving it a place of its own in the status bar is a bit too much. A keyboard shortcut is perfect. It was just that the paredit mode icon reflected the disablement that got me to want to use it to toggle back. I think it just leaving that icon displaying the keymap info is OK. Grey it out, but keep it toggling between keymaps.

I agree here too. Do you think it would also be wise to add something to the hover tooltip? Like now it might say `Toggle to original mode` but it could also say `Toggle to original mode (paredit shortcuts are off)`, or something when they're off, and maybe also say `... are on` when on. Or maybe this should gray/color thing should only be mentioned in the docs. I'm not sure.

bpringe (changes requested):
As per the discussion. Let's make this toggle all Calva shortcuts, and not just for paredit. Again, thanks for your work on this.

wcerfgba:
Closing in favour of #796 , I changed the branch name on my fork. :+1:

PR #793, opened 2020-09-20
Show the REPL/output with cursor at the bottom (#792)

### LINKED ISSUE #792: Show the REPL/output with cursor at the bottom
Hi! :smile:  This issue based on the fact that the REPL/output window can be easily closed, so we need a convinient way to open it again. There is a keyboard shortcut `ctrl+alt+c o`, which is very nice (and I use it a lot), but it has one drawback: the REPL/output opens with cursor at the top of the file. That means the user must manually scroll to the bottom to use the REPL.

There is one special case, though, when `ctrl+alt+c o` brings back the REPL/output with cursor at the bottom: when the REPL/output contains unread evaluation results. It's implemented by introducing state:
https://github.com/BetterThanTomorrow/calva/blob/5ff07b1584991689156ff103e7e9bc50b0e47480/src/results-output/results-doc.ts#L245-L253
This auto-scrolling happens not only after using the keyboard shortcut, but also after opening the REPL/output file manualy by clicking in the file explorer (not a big deal, but maybe a little bit magical).

I cannot imaging the case when it's necessary to open the REPL/output with cursor at the top, so my proposale is to eliminate the state and always bring it back with cursor at the bottom when using command/shortcut. The meaning of the command: "Show the REPL/output ready to be used". Manual clicking in the file explorer will still open the file in a regular and predictible way.

I've already made this change in my fork and it works very well for me. Will be happy to make a PR.

glebovmaksim:
## What has Changed?
Eliminates the state and always shows the REPL/output with cursor at the bottom when using command/shortcut `ctrl+alt+c o`.

Fixes #792 

## My Calva PR Checklist

I have:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Changes to Calva.
 
 ## [Unreleased]
+- Fix: [Show the REPL/output with cursor at the bottom](https://github.com/BetterThanTomorrow/calva/issues/792)
 - [Fix: evals should be ignored during parsing](https://github.com/BetterThanTomorrow/calva/issues/763)
 - Fix: [Test runner can't find tests under cursor when using a custom test macro](https://github.com/BetterThanTomorrow/calva/issues/786)
 
--- a/src/results-output/results-doc.ts
+++ b/src/results-output/results-doc.ts
@@ -156,7 +156,9 @@ export async function openResultsDoc(): Promise<vscode.TextDocument> {
 
 export function revealResultsDoc(preserveFocus: boolean = true) {
     openResultsDoc().then(doc => {
-        vscode.window.showTextDocument(doc, getViewColumn(), preserveFocus);
+        vscode.window.showTextDocument(doc, getViewColumn(), preserveFocus).then(editor => {
+            util.scrollToBottom(editor);    
+        })
     });
 }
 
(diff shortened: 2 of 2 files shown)


PEZ:
I don't think we can remove the scrolling in that append method. When the window is not visible it should still be scrolled when made visible. So my guess is that this will not work when the output window is in the same tab group as the file where the evaluation happens. I'd be happy to be wrong about it, though. No fan of state. 😄

glebovmaksim:
The output window will always be scrolled to the bottom when made visible via the `ctrl+alt+c o`. It also works with "Open Results Window" button on hover (it's the same command) and with peek to definition. When using another methods (clicking on tab / `ctrl+tab` etc) the scrolling will not work. I'll revert that change if you consider it bad behaviour. This will require extra check to prevent double scrolling and will no longer be simple, but maybe it is necessary.

bpringe:
I like your changes here, but I agree with @PEZ that keeping the auto scroll behavior even when the output window is clicked via a tab and when there are unread eval results, is best. So your change would be additive and not remove any existing functionality. 

I'm not sure that preventing double scrolling is really necessary, as it's a pretty insignificant thing, I would guess, but if it's pretty simple to prevent, then we should.

PEZ:
Will you maybe have time to finish this during Christmas, @glebovmaksim ?  😄

ieugen:
Hi @glebovmaksim ,

I would like to see this through since I usually code with split window and having Calva output there messes with my flow.

Are you interested in continuing this? 
Are you open to let someone else continue? 

I would give it a shot, even though I know quite little about typescript and even less about vscode (that is how much I want to see this though).

PR #848, opened 2020-11-20
Add ns form auto creation

### LINKED ISSUE #844: Adding ns form to a newly created clojure file
Cursive and Emacs have this feature. It would be nice to have it in Calva too.

When I create a new clojure file, I'd like its ns form is created automatically. For example if a new file is `src/my_app/core.clj`, I'd expect to see `(ns my-app.core)` in it.

To do it we need to add a handler for `workspace.onDidCreateFiles`.
https://code.visualstudio.com/api/references/vscode-api#workspace

Also the [vscode-java](https://github.com/redhat-developer/vscode-java) extension already supports [similar feature](https://github.com/redhat-developer/vscode-java/blob/a7854a12cb948b207968e32cfeeac0eec5c788d1/src/fileEventHandler.ts#L44) for java files.

zelark:
## What has Changed?
- Added `classpath` nREPL operation.
- Added `handleNewCljFiles` handler and registered it.

Closes #844

## My Calva PR Checklist

I have:

@pez, @bpringe please review the changes. Please note I don't know TS really well so probably there are a better ways to implement things. Looking forward to your feedback.


src/fileHandler.ts
```diff
@@ -0,0 +1,76 @@
...
+      }
+  }
+  for (let i = 0; i < emptyFiles.length; i++) {
+      const filename: string = resolveFileName(textDocuments[i].fileName);
+      const nsname: string = resolveNsName(sourcePaths, emptyFiles[i].fsPath);
+      const snippets: string[] = [];
+      if (nsname) {
+          snippets.push(`(ns ${nsname}.${filename}\$1)`);
+      } else {
+          snippets.push(`(ns ${filename}\$1)`);
```
bpringe, line 70: What is the point of the `\$1`?

src/fileHandler.ts
```diff
@@ -0,0 +1,76 @@
...
+      }
+      snippets.push("")
+      const textEditor = await vscode.window.showTextDocument(textDocuments[i]);
+      textEditor.insertSnippet(new vscode.SnippetString(snippets.join("\n")));
+  }
+}
```
bpringe, line 76: We can make this a bit more functional like this, I think.
```suggestion
export default async function handleNewCljFiles(e: vscode.FileCreateEvent) {
  const emptyFiles = e.files.filter(file => !isCljFile(file));
  let sourcePaths: string[] = [];
  const client = namespace.getSession();
  if (client) {
    const resp = await client.classpath();
    if (resp && resp.classpath) {
      sourcePaths = resp.classpath.filter(p => fs.existsSync(p) && fs.lstatSync(p).isDirectory());
    }
  }
  emptyFiles.forEach(async file => {
    const textDocument = await vscode.workspace.openTextDocument(file);
    if (textDocument.getText()) {
      return; // We only want to handle empty files
    }
    const filename: string = resolveFileName(textDocument.fileName);
    const nsname: string = resolveNsName(sourcePaths, file.fsPath);
    const snippets: string[] = [];
    if (nsname) {
      snippets.push(`(ns ${nsname}.${filename}\$1)`);
    } else {
      snippets.push(`(ns ${filename}\$1)`);
    }
    snippets.push("");
    const textEditor = await vscode.window.showTextDocument(textDocument);
    textEditor.insertSnippet(new vscode.SnippetString(snippets.join("\n")));
  });
}
```

bpringe (commented):
This looks great so far. I know it's a WIP, but I added a couple comments.

zelark:
> This looks great so far. I know it's a WIP, but I added a couple comments.

That's exactly was my point, to get early feedback. :)

src/fileHandler.ts
zelark: The point is to move the cursor to this position right after an ns form is created. It's needed because the next highly possible step is to add a require block. Since the cursor at the right position, you just need to hit the enter.

$1, $2 and so on are called [tabstops](https://code.visualstudio.com/docs/editor/userdefinedsnippets#_tabstops) and used in snippets.

src/fileHandler.ts
zelark: Thanks for the idea. I think I can come up with something more functionalish. But first I want to simplify the solution a bit.

PEZ: Agree, a bit more functionalish is to prefer. We should write some about that on our [Coding Style](https://github.com/BetterThanTomorrow/calva/wiki/Coding-Style) page.

PEZ:
So good that you are championing this, @zelark! It might seem like a small thing, but I have a feeling that Calva will feel a lot more friendly and professional when this thing is in place.

PEZ:
I checked up on the build failure, btw. It doesn't look like it caused by your changes. I'm rerunning the build now and if it fails the same then @bpringe and I have some work to do, I guess. 😄

PEZ:
Another, general, thing. If you can factor the code such that some logic is free from vscode dependencies, than that is always good, because then it can be tested with our unit test harness. That said, I think a lot of the work here is quite vscode dependant so it is just something to keep in mind.

And now the build passed. Some temporary hick-up in downloading dependencies for the cljs parts of Calva.

src/fileHandler.ts
zelark: @bpringe, @PEZ The suggested functionalish version has one drawback: it doesn't short circuit. If there are no files to process, the array is empty, it still send a message to the server. I tried my best to make it more functionalish, but I failed, it's hard here because of presence of async code. If you write something like `cljFiles.filter(f =? isCljFile(f)).map(async fn)...`, you have to deal with array of promises, and also filter nils then.

But I managed to simplified the code a bit and going to push it in a few minutes.

zelark:
@PEZ, @bpringe 
Factored helpers and added unit tests. I guess it's ready for the final code review now. I'm open to any recommendations or changes if there are ones.

bpringe:
The code looks and tests look good. I noticed when testing the vsix that if I'm not connected to a repl, the ns form is still added, but with just the base file name. Is the auto creation of the ns form intended to occur when no repl connection exists? I think in most cases this would create an inaccurate ns form.

I wonder if we should make the auto creation only occur if there's a repl connection, OR is it possible to not use nrepl to find the correct path so this feature can work without a repl connection? I think the latter would involve statically analyzing the project file, so probably would require a decent amount of extra work.

zelark:
> if I'm not connected to a repl, the ns form is still added, but with just the base file name

Yes, I did it intentionally, but if it's not desired behavior we can throw it away.

> I wonder if we should make the auto creation only occur if there's a repl connection

Good question! We should decided. I would prefer to get at least inaccurate `(ns file-name)` form. Note, there are some cases when users will also get an inaccurate ns form. For example when they don't select a profile with a source path, thus it won't fall into classpath, but then they try to create a namespace under that source path.

> is it possible to not use nrepl to find the correct path so this feature can work without a repl connection?

Yes it's possible, and as I mentioned in the origin issue Cursive does exactly that. It analyzes `project.clj` or `deps.edn` depends on type of project user chooses. Also it allows users to set source paths manually. So in the future I guess a kind of static analysis might be implemented in Calva to read these files.

Btw, in Cursive you'll also get inaccurate ns form if a profile isn't active.

bpringe:
Oh I see, in that case, I think the current implementation is good. Since we plan to add static analysis via clojure-lsp, maybe we can revisit this feature (among other repl-driven features) once we have that tool.

zelark:
Great! I guess I need to update CHANGELOG.md now.

bpringe:
I think this great new feature should be documented as well. This would also help clear up questions people may have like "why is it not creating the full ns path?" when they aren't connected to a repl - details that should be in the docs.

See the How to Hack on Calva wiki page for info on editing docs. I'm not sure exactly where this info should go, but I'm thinking maybe a new page under Features called "Miscellaneous" where we put things like this, or we could just create a page specifically for this.

bpringe:
I've just added a checklist item to the PR template in `dev` to make sure we update docs when appropriate.

zelark:
Added docs, please review it as well. Writing docs is not my strength part. 😊

bpringe:
Great docs! I made a couple of small edits / fixed typo. Time to merge! This will go out with the next release. Great work!

bpringe:
I just realized clojure-lsp has this feature: https://github.com/snoe/clojure-lsp#initializationoptions

>auto-add-ns-to-new-files? if true or not present, will automatically add the ns form in new files.

May revisit this later once we integrate clojure-lsp, but glad we got the feature sooner than later.

PR #886, opened 2020-12-23
885 fstat await problem

### LINKED ISSUE #885: Jack-in prompt never showing on some Windows machines
On my Virtual Box Windows machine, jack-in doesn't work since we added the Live Share changes. The cause is that the way we use `vscode.workspace.fs.stat(uri)` is not the way it is [recommended by the VS Code team](https://github.com/microsoft/vscode-extension-samples/blob/master/fsconsumer-sample/src/extension.ts).

This doesn't seem to affect non-Windows users, and also I see no reports of it causing problems to any Windows users either. Except me then.

The code where this is happening is fundamental, so when it hits it will cause a lot of things in Calva to stop working.

PEZ:
## What has Changed?
Fix jack-in prompt never appearing on Windows.

Doesn't seem to happen to users out there, but on my Windows machine jack-in doesn't work.

The fix is to wrap the `fs.stat()` call in a try/catch instead of using the return value. Quite strange, but it is how the [VS Code team is recommending it to be done](https://github.com/microsoft/vscode-extension-samples/blob/master/fsconsumer-sample/src/extension.ts). (And also, it does fix the problem.)

Fixes #885

## My Calva PR Checklist

I have:

### DIFF
--- a/src/state.ts
+++ b/src/state.ts
@@ -264,7 +264,9 @@ async function findProjectRootUri(projectFileNames, doc, workspaceFolder): Promi
                 catch { }
             }
         }
-        catch (_) { }
+        catch (e) { 
+            console.error(`Problems in search for project root directory: ${e}`);
+        }
         prev = searchUri;
         searchUri = vscode.Uri.joinPath(searchUri, "..");
     }
--- a/src/state.ts
+++ b/src/state.ts
@@ -253,17 +253,18 @@ async function findProjectRootUri(projectFileNames, doc, workspaceFolder): Promi
     let searchUri = doc?.uri || workspaceFolder?.uri;
     let prev = null;
     while (searchUri != prev) {
-        for (let projectFile in projectFileNames) {
-            const u = vscode.Uri.joinPath(searchUri, projectFileNames[projectFile]);
-            try {
-                const stat = await vscode.workspace.fs.stat(u);
-                if (stat) {
+        try {
+            for (let projectFile in projectFileNames) {
+                const u = vscode.Uri.joinPath(searchUri, projectFileNames[projectFile]);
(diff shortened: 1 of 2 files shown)


PEZ:
I added you as a reviewer just so that you can look at the change, @bpringe . You don't need interrupt your workflow with those other changes to test this out, I have tested it quite extensively. But if the changes look good to you, please merge the PR. =)

kstehn:
actually i had some problems with jack-in.
but not with every project but some and i didnt knew why.
So i couldnt write repro-steps for this.

But back to topic:
i tested the vsix for my problem projects and they work now correctly :)

PEZ:
Oh, then we should hurry to get this out. The error behaviour is so silent. Very strange that it only his some projects, but life is full of mysteries, I guess. 😎

Thanks for sharing the info!

src/state.ts
```diff
@@ -253,17 +253,18 @@ async function findProjectRootUri(projectFileNames, doc, workspaceFolder): Promi
...
+                try {
+                    await vscode.workspace.fs.stat(u);
                     cursor.set(PROJECT_DIR_URI_KEY, searchUri);
                     return;
                 }
+                catch { }
             }
-            catch (_) { }
         }
+        catch (_) { }
```
bpringe, line 267: Should we log something useful here in case we have an issue later, so we can easily find the source?

bpringe (changes requested):
Looks good to me, just had a minor request/comment.

src/state.ts
PEZ: Yes we should. Good catch!

PEZ:
I added the logging. Merging this now.

PR #889, opened 2020-12-24
Remove hover info about connecting to REPL

PEZ:
Also use lsp hovers and completions when not connected

Ping  @bpringe


bpringe:
Thanks

PR #904, opened 2020-12-29
Load current repl window namespace

### LINKED ISSUE #808: repl - prompt space ending
The repl prompt currently has a space at the end. Added in this [commit](https://github.com/BetterThanTomorrow/calva/commit/50f517950ade8df62cf77a04d3df364f0e3da292)

If using an extension such as [Trailing Spaces](https://marketplace.visualstudio.com/items?itemName=shardulm94.trailing-spaces) this will colour all repl prompt lines.

image

Was there a particular reason the above commit was added? Happy to do a PR to remove this space but obviously this was added for a reason maybe.

PEZ:
So, two for one in this PR. Just because they go a bit together....

## Load current repl window namespace

Make the **Load Current File** command work in the repl/output window, but then load the current namespace instead.

Fixes: #910 

## Special colon symbols and hard space in the prompt
I find myself wanting to have easy access to the namespaces shown by the prompt in the output window. This change replaces the regular `:` (colon) characters with some unicode version I found that looks the same. Then I could change the grammar some to not treat this as part of symbols.

This makes the ns in the prompt a peekable symbol and also allows for navigating to it.

I also replaced the space with a hard space, hoping that that Trailing Space extension will not warn about it.

Fixes #808

## My Calva PR Checklist

I have:


PEZ:
Marking this as ready for review, @bpringe , even though I do think I need to go over the changes some before I think they are ready to go. But if you find some time for it, please have a look. One of the things you might want to look at is that I remove the jar `TextDocumentProvider` that the `lsp` module added. Since we already have such a provider, which I reuse for the load namespace command.

bpringe:
I'll have to try out / review the rest of it later, but I'm not sure if we should get rid of the clojure-lsp jar content provider. I spent a while fixing it for Windows recently in clojure-lsp, and with the code in this PR, when I try to navigate into a jar dependency on Windows it fails (just a blank file with a progress indicator that doesn't stop).

The clojure-lsp jar content provider (via that dependencyContents request we make), works for Windows. Maybe if we fix Calva's to work for Windows it would be fine, but I also don't see an issue with only using clojure-lsp's jar content provider.

PEZ:
Yes, we could use either or. It is mainly using both that I don't like. I considered using the one in the lsp module, but it had a dependency on the lsp client, and it was late here. I'll have a look.

PEZ:
Trying the `dev` branch on Windows, I don't get any definitions for core or library symbols. In fact there is no a call to our jar content provider. So there is something else going on. And it makes it hard to fix the problem you see of course, since I never get there. I'll make it work on Mac (and probably on Linux) and we make it work on Windows later. Our own jar provider is tried and tested over many years, so I'll go for that one for now.

bpringe:
About to try this out now!

bpringe:
> In fact there is no a call to our jar content provider. So there is something else going on.

FYI, I think there is some caching business that occurs with the content provider. If you don't see a breakpoint in the provider being hit, restart the extension development host and then try it again. I've noticed I have to do that, but after the contents are provided once, the feature works without going into the provider code again.

bpringe:
Ok I have some insights about the jar content loading on Windows. What I know so far:

* Until clojure-lsp finishes starting, using Go to Definition on a jar dep says "No definition found"
* Once clojure-lsp starts, execution goes through the `provideDefinition` middleware and since no repl is connected, it calls `next`
* This puts execution inside `getJarContents`, where the uri looks like:
   image
* After the manipulations to `pathToJar` it ends up being
  image
* This results in an error when `readFile` is called:
  image

bpringe:
Aside from fixing this, we should also log the error so we have more insight if it errors in the future.

bpringe:
Added a fix for the windows issue here: https://github.com/BetterThanTomorrow/calva/pull/918. If you approve, then this PR looks good to go, aside from anything else you wanted to do. :smiley:

PR #916, opened 2020-12-31
Wake up core navigation again

### LINKED ISSUE #915: We can no longer navigate from within a core or library file
The lsp isn't active in files opened outside the project. So when we navigate to a core function or a function in some library we  no longer have lsp help.

This would be OK, I guess, if it wasn't that we also lose functionality we had before the lsp. Our previous DocumentDefinition provider allowed the user to keep navigating from core and library files. There are probably other areas where we have or risk getting this problem.

I'm grasping for ideas here. Best would be if we could make the lsp keep serving the user also with the files opened outside the project. But if we can't, then we need to somehow dispatch these services in some smart way...

PEZ:
Reactivate nrepl definition provider

And then run it from the lsp provider as well to see if it would answer
If so lsp provider returns null

Fixes #915

## My Calva PR Checklist

I have:

### DIFF
--- a/src/lsp.ts
+++ b/src/lsp.ts
@@ -57,7 +57,7 @@ function createClient(jarPath: string): LanguageClient {
             async provideDefinition(document, position, token, next) {
                 const nReplDefinition = await provideClojureDefinition(document, position, token);
                 if (nReplDefinition) {
-                    return nReplDefinition;
+                    return null;
                 } else {
                     return next(document, position, token);
                 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Changes to Calva.
 
 ## [Unreleased]
+- [Reactivate definitions/navigation in core and library files](https://github.com/BetterThanTomorrow/calva/issues/915)
 
 ## [2.0.142 and 2.0.143 - 2020-12-30]
 - No changes besides version number. Released due to vsix publishing issues.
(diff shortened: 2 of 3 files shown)


src/lsp.ts
```diff
@@ -57,7 +57,7 @@ function createClient(jarPath: string): LanguageClient {
             async provideDefinition(document, position, token, next) {
                 const nReplDefinition = await provideClojureDefinition(document, position, token);
                 if (nReplDefinition) {
-                    return nReplDefinition;
+                    return null;
```
bpringe, line 60: Shouldn't we instead check for a nrepl connection here? It seems with this approach, provideClojureDefinition will be called twice. It's not a big deal, but it is duplicated work.

PEZ: Then we would loose local definition navigation when nrepl connects. Nrepl doesn't provide that.

bpringe: I'm a bit confused, because with this change `provideClojureDefinition` is called twice, once here and once from the ClojureDefinitionProvider that we still have registered. I'm not sure why we have it registered, the comment above it says it's not used.

If we remove the ClojureDefinitionProvider (or don't register it), then return `nReplDefinition` here, then `provideClojureDefinition` will only be called once, and we'll get a definition whether nrepl is connected or not.

bpringe: Nevermind, I see now there was an issue with not using that other provider, but there is still a comment above it saying it's not used, when it must be used now, right? :thinking:

PEZ: Yes, the comment is confusing. And the whole setup is confusing, so we should try to factor it in a way that we understand what is going on. 😎

bpringe: https://github.com/BetterThanTomorrow/calva/pull/919 I tried the change I mentioned and it doesn't seem that we lose anything. Please check it out, and let me know if I'm missing something. Maybe I'm misunderstanding what you mean by "local definition navigation."

PR #972, opened 2021-01-21
Fix nrepl port race condition

### LINKED ISSUE #901: Jack-in prompting for nrepl port since terminal jack-in change
Since the recent jack-in change, I'm getting a prompt for the nrepl port at jack-in. This doesn't happen every time though. Maybe it's a race condition related to the .nrepl-port file that's created.

image

clyfe:
## What has Changed?

- Race is: the get-port-from-file code overwrites the port-from-regex with nothing when the nrepl-port file is not yet present at the time is read.
- Fix: don't get-port-from-file when port already present.

Fixes #901 

## My Calva PR Checklist

I have:

- [x] Tested the VSIX built **locally** from the PR (so, after you've submitted the PR).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Changes to Calva.
 
 ## [Unreleased]
 - [Use clojure-lsp for usages for debug instrumentation decorations, and stop injecting clj-kondo at jack-in](https://github.com/BetterThanTomorrow/calva/issues/931)
+- Fix: [nrepl port detection race condition](https://github.com/BetterThanTomorrow/calva/issues/901)
 
 ## [2.0.151] - 2021-01-15
 - Fix: [Debugger is broken on Windows](https://github.com/BetterThanTomorrow/calva/issues/947)
--- a/src/connector.ts
+++ b/src/connector.ts
@@ -451,8 +451,10 @@ export async function connect(connectSequence: ReplConnectSequence,
     state.extensionContext.workspaceState.update('selectedConnectSequence', connectSequence);
 
     try {
-        let bytes = await vscode.workspace.fs.readFile(portFile);
-        let port = new TextDecoder("utf-8").decode(bytes);
+        if (port === undefined) {
+            let bytes = await vscode.workspace.fs.readFile(portFile);
+            port = new TextDecoder("utf-8").decode(bytes);
+        }
         if (port) {
             hostname = hostname !== undefined ? hostname : "localhost";
             if (isAutoConnect) {


PEZ:
Ah, great that you found where this happens! It's consistent with the behaviour that this just happens now and then.

However, we can't delete that code. It is used for the **Calva Connect** command. What you'll need to do is to only execute  it conditionally. So if we already have a port, this should not be done, otherwise it should.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ Changes to Calva.
 
 ## [Unreleased]
 - [Use clojure-lsp for usages for debug instrumentation decorations, and stop injecting clj-kondo at jack-in](https://github.com/BetterThanTomorrow/calva/issues/931)
+- [Fix nrepl port detection race condition](https://github.com/BetterThanTomorrow/calva/issues/901)
```
```suggestion
- Fix: [nrepl port detection race condition](https://github.com/BetterThanTomorrow/calva/issues/901)
```

src/connector.ts
```diff
@@ -438,23 +438,65 @@ export let nClient: NReplClient;
 export let cljSession: NReplSession;
 export let cljsSession: NReplSession;
 
+function filePresence(portFile: vscode.Uri, timeout: number) {
```
PEZ: It is unclear to me why we need a file watcher. There are mainly two paths in to the connect:

1. Jack-in. The port will be known already. No need to read the file.
2. Connect: Assumes the repl is started and and reads the file if it is there.

clyfe:
Tested the VSIX built from the PR locally, but not the one off circleci, nor will I do so.

PEZ:
> Tested the VSIX built from the PR locally, but not the one off circleci

That's perfectly OK. The point is to make sure the release build works in the sense that the tree shaking and munging hasn't broken things. It is tested just as well with your locally built VSIX:

clyfe (commented):
Let's merge.

PEZ (changes requested):
This needs to be addressed before merge.

src/connector.ts
clyfe: Removed.

src/connector.ts
```diff
@@ -439,22 +439,29 @@ export let cljSession: NReplSession;
 export let cljsSession: NReplSession;
 
 export async function connect(connectSequence: ReplConnectSequence,
-    isAutoConnect: boolean, hostname?: string, port?: string) {
+    isAutoConnect: boolean, maybeHostname?: string, maybePort?: string) {
```
PEZ: We currently do not name optional arguments like this in the code base. It might be something we should consider, but  until we have decided on that, let's not introduce it in one place only.

src/connector.ts
```diff
@@ -439,22 +439,29 @@ export let cljSession: NReplSession;
...
-    const portFile = projectTypes.nreplPortFileUri(connectSequence);
-
     state.extensionContext.workspaceState.update('selectedCljsTypeName', cljsTypeName);
     state.extensionContext.workspaceState.update('selectedConnectSequence', connectSequence);
 
     try {
-        let bytes = await vscode.workspace.fs.readFile(portFile);
-        let port = new TextDecoder("utf-8").decode(bytes);
+        let port: string | undefined = maybePort;
+        if (port === undefined) { // when we connect to a repl started externally
```
PEZ: I had to think quite a while before I realised what ”externally” means here. I think we can cope without the comment. Looks clear enough to me that we are looking for a port number in the file because we don't have a port number yet. Then again, maybe the function as such lacks a documentation section which could inform about that if a port is not supplied we will search in the port file.

src/connector.ts
```diff
@@ -439,22 +439,29 @@ export let cljSession: NReplSession;
...
+            const portFile = projectTypes.nreplPortFileUri(connectSequence);
+            // Most of the time the repl started externally is of project provenance
+            // and the port is the one in the nrepl file, use it if there is one.
+            // Noting that's not always the case; if the nrepl is from another project
+            // the port guessed from the opened folder is not the one we want, and
+            // it needs to be manually changed to the right one.
```
PEZ: I think the code is clear enough here. Adding the comments makes bit harder to scan the function.

PEZ (changes requested):
Thanks. We're getting there. 😄 

Generally (and we should add this to the contribution guidelines) we want fixes like this to be quite surgical. You've found the place for the race condition, you've fixed it. If you see other changes that should be made, better send them in a separate PR.

src/connector.ts
clyfe: Reverted.

src/connector.ts
clyfe: Removed.

src/connector.ts
clyfe: Removed.

PEZ (approved):
Thanks. Totally appreciate that you helped fix this and followed through.

Issue #980, opened 2021-01-22
Allow disabling diagnostics


plexus:
Calva receives various diagnostics from clojure-lsp (which gets most but not all of these from clj-kondo), it then displays them as squiggly annotations.

Not everyone wants this, I mostly perceive these as noise. People with ADHD in particular may find these really troublesome.

I could not find a way to disable these. Calva and clojure-lsp don't seem to have a flag for it. I couldn't find a way to tell VS Code to ignore linting/annotations. The best you can do right now is to disable linters in clj-kondo, but not all diagnostics are coming from clj-kondo (e.g. unused namespace, unused public are coming from clojure-lsp itself).

I'm testing out a patch that adds a configuration flag for this, essentially this:

``` json
// package.json
                    "calva.displayDiagnostics": {
                        "type": "boolean",
                        "markdownDescription": "Show linter warnings and other diagnostics from clojure-lsp and clj-kondo.",
                        "default": true
                    }
```

``` typescript
// lsp.ts
        middleware: {
            handleDiagnostics(uri, diagnostics, next) {
                if (!state.config().displayDiagnostics) {
                    return;
                }
```

If it works I can send a PR, meanwhile happy to get some input on this. Is this a reasonable place to handle this? Any opinions on the config flag / description?

bpringe:
Hey @plexus. Thanks for taking time to look into this. That is the appropriate solution, but I've not added that because we have been waiting on https://github.com/clojure-lsp/clojure-lsp/pull/261 to be merged, which I believe uses only clj-kondo for diagnostics, in which case disabling clj-kondo diagnostics as per clj-kondo's instructions will/should work, and this Calva toggle is superfluous, imo.

However, it may still be a good idea to add this setting? @PEZ Thoughts? 

So if you disable all linting via kondo's config, I think it may cause kondo to not do the linting work in the first place (@borkdude correct me if I'm wrong). But with this, the linting work is done, just not shown to the user. So what I mean to say is if that's true, this may not be a good setting to provide as there's a better way to disable diagnostics.

bpringe:
That PR is complete now, so we just need to merge into Calva and do some testing. Once we release then the disabling of all linting via kondo's config should work, if I understand correctly.

PR #981, opened 2021-01-22
Add a configuration option to disable diagnostics

plexus:
## What has Changed?

A boolean config flag `displayDiagnostics` was introduced, it defaults to true. When it is set to false, then any diagnostics (i.e. linter warnings) coming from clojure-lsp are ignored.

See #980 for the motivation behind this.

## My Calva PR Checklist

I have:

- [x] Tested locally by building the vsix, unzipping it, and copying `extension` to `~/.vscode/extensions`

### DIFF
--- a/package.json
+++ b/package.json
@@ -509,6 +509,11 @@
                         "type": "boolean",
                         "markdownDescription": "Don't surface REPL UI elements when avoidable. (For when using another REPL extension together with Calva's static features.)",
                         "default": false
+                    },
+                    "calva.displayDiagnostics": {
+                        "type": "boolean",
+                        "markdownDescription": "Show linter warnings and other diagnostics",
+                        "default": true
                     }
                 }
             },
--- a/src/lsp.ts
+++ b/src/lsp.ts
@@ -29,6 +29,9 @@ function createClient(jarPath: string): LanguageClient {
         },
         middleware: {
             handleDiagnostics(uri, diagnostics, next) {
+                if (!state.config().displayDiagnostics) {
+                    return;
+                }
                 if (uri.path.endsWith(config.REPL_FILE_EXT)) {
                     return;
                 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Changes to Calva.
 
 ## [Unreleased]
 
+- [Add a configuration option to disable diagnostics](https://github.com/BetterThanTomorrow/calva/pull/981)
+
 ## [2.0.171] - 2021-02-10
 - Update clojure-lsp to version 2021.02.09-18.28.06 (Fix: [Auto completion does not work in clojure-lsp only mode (no repl connection)](https://github.com/BetterThanTomorrow/calva/issues/996#issuecomment-776148282))
 - Update clojure-lsp to version 2021.02.10-03.01.19 (Fix: [Project clj-kondo config file not being considered](https://github.com/BetterThanTomorrow/calva/issues/1026))
--- a/src/state.ts
+++ b/src/state.ts
@@ -122,6 +122,7 @@ function config() {
         autoOpenJackInTerminal: configOptions.get("autoOpenJackInTerminal") as boolean,
         referencesCodeLensEnabled: configOptions.get('referencesCodeLens.enabled') as boolean,
         hideReplUi: configOptions.get('hideReplUi') as boolean,
+        displayDiagnostics: configOptions.get('displayDiagnostics') as boolean,
     };
 }
 
@@ -243,7 +244,7 @@ async function findProjectRootUri(projectFileNames, doc, workspaceFolder): Promi
                 catch { }
             }
         }
-        catch (e) { 
+        catch (e) {
             console.error(`Problems in search for project root directory: ${e}`);
         }
         prev = searchUri;


PEZ:
Hello, many thanks for helping out with this!

I see three somewhat separate questions here:

1. Why does configuring this through clj-kondo configs not work?
2. Should Calva have this setting?
3. If so, how should the setting be implemented?

The place for discussing **1** is probably #980, but I'll ping in @bpringe and @ericdallo here anyway for their context. I don't right now recall the status of things in regards to how the clj-kondo inside clojure-lsp is configured, but it would not surprise me if https://github.com/clojure-lsp/clojure-lsp/pull/261 can offer some hope for fixing #980.

As for the setting's be or not be:
 * I think I am for it as something I want to the users to have. It's a bit opaque to the user how clj-kondo fits into the picture, so for many it will be a bit far fetched to figure out that the linting can be switched off with a clj-kondo config file. Let's see what @bpringe's take is. 
 * Something that might make it a bit difficult to implement is if the user wants different settings for different subdirectories/subrepos. Maybe the settings could be considered a master switch here and for more granular control the user would need to have linting enabled and then use clj-config files to control it in per directory tree.

As for implementation of the setting, some things we should consider, imo:

* Most Calva settings take effect when they are changed. This PR's way of implementing the setting adheres to this pattern.
* Can/should the setting be done through configuring the clojure-lsp server?
    * If so, can we configure it on the fly or would it take a server restart?
    * If it takes a server restart, is that instant enough?
* Should the setting be implemented by Calva editing the clj-kondo config file? This would also adhere to the instant effect, I think, because that is how clj-kondo works, right?

As far as a master switch goes, I think it makes sense to try to stop linting from even happening if someone switches it off. 

I will let the answer to question *1*. above guide me a bit. If it is something we can assume is about to be fixed, then maybe wait with implementing a setting until we get indications that many users want to disable and have trouble figuring out how. If the fix is a way away, then maybe pull this PR as it is as a temporary fix. (Not sure, since giving the users the setting easily gets a commitment to keep it.)

ericdallo:
About 1:
We from clojure-lsp are working hard last and this week on https://github.com/clojure-lsp/clojure-lsp/pull/261 which is a huge refactor that will make clojure-lsp use 100% clj-kondo for almost everything, therefore IMO this issue would be closed since user would need just to disable clj-kondo lint as it already has this possibility from clj-kondo side, WDYT?

plexus:
Thanks a lot for the input @PEZ and @ericdallo. It's good to see clojure-lsp using the clj-kondo linters for unused-ns/unused-public, this was definitely a source of confusion.

IMO for something that is so invasive in terms of UX deserves a clear master switch inside the Calva settings. That's where people will look for it. Ideally this should signal to clojure-lsp that no linting is wanted/needed, so that clojure-lsp can skip invoking clj-kondo. No wasted cycles just to start it up and do nothing. No creating `.clj-kondo/cache` which remains empty.

If people want this toggle on a per project level they should be able to do it in `.lsp/config.edn`, something like `{:clj-kondo false}` for instance. If they want to fine tune the linters they can use `.clj-kondo/config.edn` (or `.lsp/config.edn`.

But telling people that if they are not interested in clj-kondo that they should just disable all linters doesn't seem right. Why do I need to configure a tool I don't want? It's also not clearly documented how one is supposed to do that. The clj-kondo docs don't mention it, since indeed why would they? That's not what you're there for. I *think* `{:linters ^:replace {}}` should do the trick, but you have to admit that that's not blindingly obvious. (not to mention `~/.clj-kondo/config.edn` vs `.clj-kondo/config.edn` vs `:clj-kondo` in `.lsp/config.edn`)

So that's my proposal
- have a flag at the clojure-lsp level to not invoke clj-kondo
- allow setting that from `.lsp/config.edn` or from Calva settings
  - if it's OFF in Calva then it's OFF (Calva setting is the master switch)
  - if it's ON in Calva, then honor project-level config (both default to ON)

I'd be happy to make PRs accordingly. Pointers and tips are very welcome since I'm fresh to both projects.

ericdallo:
@plexus clojure-lsp uses clj-kondo not just for linting, but to analyze the code to make most of features work, not sure would be possible to toggle that on clojure-lsp, it's a core part of clojure-lsp now :/ 

If you want just to disable limiting, the LSP client, in this case Calva, should have a feature to not produce diagnostics following [LSP spec](https://microsoft.github.io/language-server-protocol/specification#textDocument_publishDiagnostics) via `textDocument.publishDiagnostics` capability

PEZ:
I agree with your proposal, @plexus. Specifically the version where this is a VS Code setting, which is where most people search for settings, I am pretty sure. So that part of this PR is good already.

If you want, and have the time, @plexus , I'd welcome this PR to implement it with the mechanism that @ericdallo points at. If you don't have the time, leave it open and I can take it from here. Just be informed that I think it is crazy to not use clj-kondo linting so I might not prioritize this for speedy implementation. 😄

ericdallo:
For the implementation, check how to setup `capabilities` on this [VSCode guide](https://code.visualstudio.com/api/language-extensions/language-server-extension-guide) and you may need to change it [here](https://github.com/BetterThanTomorrow/calva/blob/8d79f0bf4bc980d3a45a96a507f9ef1b525acaa4/src/lsp.ts#L11)

bpringe:
Ah I'm late to the discussion here! I saw this after replying on the related issue. I agree with you @plexus about being able to disable diagnostics without having to configure the tool you don't want. I am more certain now that a master switch in Calva settings is probably wise.

bpringe:
Just want to note that if the disabling of diagnostics requires an lsp sever restart, and this isn't easily achieved without a VS Code restart, we could use the initialization method in combination with  the middleware method, so that if the user disables diagnostics, the linting warnings disappear (not sure if this would be immediate though), and then after a restart the initialize request can configure no diagnostics.

bpringe:
I'm thinking of merging this as-is, and creating an issue for figuring out later how to disable diagnostics from the initialization request. We'd want this existing change in combination with that anyway I think (see above comment). If no objections I'll do this soon.

package.json
```diff
@@ -500,6 +500,11 @@
                         "description": "Activate keybindings.",
                         "default": true,
                         "scope": "window"
+                    },
+                    "calva.displayDiagnostics": {
+                        "type": "boolean",
+                        "markdownDescription": "Show linter warnings and other diagnostics from clojure-lsp and clj-kondo.",
+                        "default": true
```
bpringe, line 516: I think "from clojure-lsp and clj-kondo" can be removed from this description since this could be confusing to a user who may want to use the clj-kondo extension for linting but disable linting from clojure-lsp (had someone mention doing this recently). For the purposes of this setting, linting comes from Calva, and the underlying libraries are implementation details (that are apparent in lint messages themselves, at least).

bpringe (changes requested):
@plexus If you don't mind adding the change to the Unreleased section of the changelog and addressing the comment about the description, then I'd be ready to merge it.

Diagnostics are calculated/used in clojure-lsp for a feature besides just reporting the linter warnings, so trying to disable from initialization options I think is not wise, at least at this point. This middleware approach is good.

plexus:
I'll update the description as mentioned and add a changelog entry, as requested. What's up with the back merge? Can I rebase this instead?

bpringe:
I just wanted to bring in the latest changes from the dev branch and resolve a conflict or two. To be honest, I really just stick to merging and don't rebase, since it's the safer option and preserves the complete history. 

But, it seems from my basic understanding that it wouldn't cause an issue here since you're the only one working off this branch, so feel free to rebase. :smile: My apologies if I interrupted your flow, and thanks for the changes!

plexus:
I vastly prefer rebasing feature branches over merging upstream back. Much less confusing, and easier to see what actually changed in a given branch.

I rebased this on current `dev`, added a changelog entry, and dropped the "from clojure-lsp and clj-kondo" from the config flag description.

I'm sorry I couldn't polish this up better, my bandwidth for open source is a little limited at the moment.

src/lsp.ts
```diff
@@ -29,6 +29,9 @@ function createClient(jarPath: string): LanguageClient {
         },
         middleware: {
             handleDiagnostics(uri, diagnostics, next) {
+                if (!state.config().displayDiagnostics) {
+                    return;
```
bpringe, line 33: With this, I noticed when the setting is unchecked, whatever linter warnings existed before are not cleared, even when the user types in the relevant files. If we `return next(uri, [])` then the lint warnings will be cleared when the file is changed (and maybe other events like reopening it).

I'll make this change if you don't have time, just didn't want to modify your PR without checking with you.

bpringe (changes requested):
Thanks for taking the time to make those changes. I noticed one more thing when testing it and commented on the relevant line. As noted there I'll make the change if you don't have the time, no worries.

Just a note for ourselves for later, not to be included in this PR, I think we can make the diagnostics auto clear when the setting is checked by getting the current diagnostics from the lsp client, looping through the URIs, and sending an appropriate lsp notification (like `didOpen`) for each to make it publish diagnostics again and trigger the middleware logic. I'll make an issue for this later.

src/lsp.ts
plexus: Feel free to run with it!

plexus: And thanks a lot!

Issue #1023, opened 2021-02-09
Allow user to override keybindings in comments and/or strings


crispinb:
Many* users report that they prefer default text language mode keyboard shortcuts for use within comments and strings. They could configure this for themselves if calva made 'inString' and 'inComment' when-contexts available.

(* OK, one user)

crispinb:
I'm willing to have a stab at a PR for this, but don't know much about the VSCode extension API or execution model, so I'll have to read up a bit & scan the calva codebase. In the meantime I could do with some guidance for starting points.

As far as I can see from[ the docs](https://code.visualstudio.com/api/references/when-clause-contexts) and [one example from Microsoft](https://github.com/microsoft/vscode/blob/349e31fee9603711ddabfe4c8cb6fb66b72e28f3/extensions/git/src/main.ts#L43), it involves calling a vscode 'setContext' command to toggle the newly added contexts.

The initial questions that arise for me are:

- where do the relevant events come from?
- where will the code for these new contexts go?

PEZ:
I think the relevant event is `vscode.window.onDidChangeTextEditorSelection` (see [`vscode.window`](https://code.visualstudio.com/api/references/vscode-api#window))

There is probably some things to consider so that it is the current active editor that the event is about. And also what should happen when the user activates a document.

The way to examine the code structure of the editor cursor's location is to use a Calva `TokenCursor`. You can create a token cursor for the current selection of a document if you have a reference the document and know the location. The event has a reference to the editor where the event fired and also to the new selection. From the editor you can get the document.

We do something similar here:

https://github.com/BetterThanTomorrow/calva/blob/dev/src/results-output/results-doc.ts

A `LispTokenCursor` instance (which is what you'll get) has a method for checking if the cursor is within a string. You might want to add a method for checking if we're in a comment (if there isn't one already, I don't remember). The methods `getToken()` and `getPrevToken()` should take you quite far, I think...

You can start by adding the event subscription in an anonymous function right in `src/extension.ts`, and we can figure out later where we want to move the implementation.

I hope I am making some little sense here? 😄

bpringe:
Another thing that might help is placing a breakpoint in the code after getting/creating the token cursor, and calling methods on it in the Debug Console to get a feel for how it works.

crispinb:
Thanks guys that all makes sense & should be helpful.

PR #1030, opened 2021-02-13
Wip/add comment string when clauses

crispinb:
## What has Changed?

- introduced couple of functions to `extensions.ts`, called from onDidChangeTextEditorSelection & onDidChangeActiveTextEditor subscriptions
- I presume these could be the basis for any further context-type additions, pulled out into a new file
- the approach taken in 'setNewContext' would be unwieldy if you were to add many more contexts, in which case that function might need something fancier
- added a global 'lastContext'. Not sure what else to do there 
- left a couple of console.logs in there temporarily - it's hard otherwise to see the context changes (though vscode does have a 'developer inspect context keys' action which helps)

Fixes #https://github.com/BetterThanTomorrow/calva/issues/1023

## My Calva PR Checklist

I have:

- [x ] Read [How to Contribute](https://github.com/BetterThanTomorrow/calva/wiki/How-to-Contribute#before-sending-pull-requests).
- [x ] Directed this pull request at the `dev` branch. (Or have specific reasons to target some other branch.)
- [x ] Made sure I have changed the PR base branch, so that it is not `published`. (Sorry for the nagging.)
- [x ] Updated the `[Unreleased]` entry in `CHANGELOG.md`, linking the issue(s) that the PR is addressing.
     - [x ] Tested the particular change
- [x ] Referenced the issue I am fixing/addressing _in a commit message for the pull request_.
     - [x ] If I am fixing the issue, I have used [GitHub's fixes/closes syntax](https://help.github.com/en/articles/closing-issues-using-keywords)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Changes to Calva.
 
 ## [Unreleased]
+- Fix: [Allow user to override keybindings in comments and/or strings](https://github.com/BetterThanTomorrow/calva/issues/1023)
 
 ## [2.0.171] - 2021-02-10
 - Update clojure-lsp to version 2021.02.09-18.28.06 (Fix: [Auto completion does not work in clojure-lsp only mode (no repl connection)](https://github.com/BetterThanTomorrow/calva/issues/996#issuecomment-776148282))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 Changes to Calva.
 
 ## [Unreleased]
-- Fix: [Allow user to override keybindings in comments and/or strings](https://github.com/BetterThanTomorrow/calva/issues/1023)
+- Fix: [Allow keybindings to target when the cursor is inside a comment or a string](https://github.com/BetterThanTomorrow/calva/issues/1023)
 
 ## [2.0.171] - 2021-02-10
 - Update clojure-lsp to version 2021.02.09-18.28.06 (Fix: [Auto completion does not work in clojure-lsp only mode (no repl connection)](https://github.com/BetterThanTomorrow/calva/issues/996#issuecomment-776148282))
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,7 +1,9 @@
 const config = {
     REPL_FILE_EXT: 'calva-repl',
     KEYBINDINGS_ENABLED_CONFIG_KEY: 'calva.keybindingsEnabled',
-    KEYBINDINGS_ENABLED_CONTEXT_KEY: 'calva:keybindingsEnabled'
+    KEYBINDINGS_ENABLED_CONTEXT_KEY: 'calva:keybindingsEnabled',
+    CURSOR_CONTEXT_IN_STRING: 'calva:inString',
+    CURSOR_CONTEXT_IN_COMMENT: 'calva:inComment',
 };
 
 type ReplSessionType = 'clj' | 'cljs';
--- a/src/config.ts
+++ b/src/config.ts
@@ -2,8 +2,8 @@ const config = {
     REPL_FILE_EXT: 'calva-repl',
     KEYBINDINGS_ENABLED_CONFIG_KEY: 'calva.keybindingsEnabled',
     KEYBINDINGS_ENABLED_CONTEXT_KEY: 'calva:keybindingsEnabled',
-    CURSOR_CONTEXT_IN_STRING: 'calva:inString',
-    CURSOR_CONTEXT_IN_COMMENT: 'calva:inComment',
+    CURSOR_CONTEXT_IN_STRING: 'calva:cursorInString',
+    CURSOR_CONTEXT_IN_COMMENT: 'calva:cursorInComment',
 };
 
 type ReplSessionType = 'clj' | 'cljs';
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,9 +1,7 @@
 const config = {
     REPL_FILE_EXT: 'calva-repl',
     KEYBINDINGS_ENABLED_CONFIG_KEY: 'calva.keybindingsEnabled',
-    KEYBINDINGS_ENABLED_CONTEXT_KEY: 'calva:keybindingsEnabled',
-    CURSOR_CONTEXT_IN_STRING: 'calva:cursorInString',
-    CURSOR_CONTEXT_IN_COMMENT: 'calva:cursorInComment',
+    KEYBINDINGS_ENABLED_CONTEXT_KEY: 'calva:keybindingsEnabled'
 };
 
 type ReplSessionType = 'clj' | 'cljs';
--- a/src/cursor-doc/model.ts
+++ b/src/cursor-doc/model.ts
@@ -8,7 +8,8 @@ export function initScanner(maxLength: number) {
 }
 
 /** A cheesy deep-equal function for matching scanner states. Good enough to compare plain old js objects. */
-function equal(x: any, y: any): boolean {
+// TODO: OK to export from here or move to util?
+export function equal(x: any, y: any): boolean {
     if(x==y) return true;
     if(x instanceof Array && y instanceof Array) {
         if(x.length == y.length) {
--- a/src/cursor-doc/model.ts
+++ b/src/cursor-doc/model.ts
@@ -1,34 +1,11 @@
 import { Scanner, Token, ScannerState } from "./clojure-lexer";
 import { LispTokenCursor } from "./token-cursor";
+import { deepEqual as equal} from "../util/object"
 
 let scanner: Scanner;
 
 export function initScanner(maxLength: number) {
     scanner = new Scanner(maxLength);
-}
-
-/** A cheesy deep-equal function for matching scanner states. Good enough to compare plain old js objects. */
-// TODO: OK to export from here or move to util?
-export function equal(x: any, y: any): boolean {
-    if(x==y) return true;
-    if(x instanceof Array && y instanceof Array) {
-        if(x.length == y.length) {
-            for(let i = 0; i<x.length; i++)
-                if(!equal(x[i], y[i]))
-                    return false;
-            return true;
-        } else
-            return false;
-    } else if (!(x instanceof Array) && !(y instanceof Array) && x instanceof Object && y instanceof Object) {
-        for(let f in x)
-            if(!equal(x[f], y[f]))
-                return false;
-        for(let f in y)
-            if(!x.hasOwnProperty(f))
-                return false
-        return true;
-    }
-    return false;
 }
 
 export class TextLine {
--- a/src/cursor-doc/token-cursor.ts
+++ b/src/cursor-doc/token-cursor.ts
@@ -671,6 +671,14 @@ export class LispTokenCursor extends TokenCursor {
     }
 
     /**
+     * Indicates if the current token is in a comment line
+     */
+    withinComment() {
+        const cursor = this.clone();
+        return cursor.getToken().type === 'comment' || cursor.getPrevToken().type === 'comment';
+    }
+
+    /**
      * Tells if the cursor is inside a properly closed list.
      */
     withinValidList(): boolean {
--- a/src/cursor-doc/token-cursor.ts
+++ b/src/cursor-doc/token-cursor.ts
@@ -675,7 +675,8 @@ export class LispTokenCursor extends TokenCursor {
      */
     withinComment() {
         const cursor = this.clone();
-        return cursor.getToken().type === 'comment' || cursor.getPrevToken().type === 'comment';
+        const token_type = cursor.getToken().type;
+        return token_type === 'comment' || (token_type === 'eol' && cursor.getPrevToken().type === 'comment');
     }
 
     /**
--- a/src/cursor-doc/token-cursor.ts
+++ b/src/cursor-doc/token-cursor.ts
@@ -676,7 +676,7 @@ export class LispTokenCursor extends TokenCursor {
     withinComment() {
         const cursor = this.clone();
         const token_type = cursor.getToken().type;
-        return token_type === 'comment' || (token_type === 'eol' && cursor.getPrevToken().type === 'comment');
+        return token_type === 'comment';
     }
 
     /**
--- a/src/cursor-doc/token-cursor.ts
+++ b/src/cursor-doc/token-cursor.ts
@@ -676,7 +676,7 @@ export class LispTokenCursor extends TokenCursor {
     withinComment() {
         const cursor = this.clone();
         const token_type = cursor.getToken().type;
-        return token_type === 'comment';
+        return token_type === 'comment' || token_type === 'eol' && cursor.getPrevToken().type  === 'comment';
     }
 
     /**
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -30,7 +30,7 @@ import config from './config';
 import handleNewCljFiles from './fileHandler';
 import lsp from './lsp';
 import * as snippets from './custom-snippets';
-import * as docMirror from './doc-mirror'
+import setCursorContextIfChanged from './when-contexts'
 
 async function onDidSave(document) {
     let {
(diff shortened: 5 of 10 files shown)


src/extension.ts
```diff
@@ -30,6 +30,9 @@ import config from './config';
 import handleNewCljFiles from './fileHandler';
 import lsp from './lsp';
 import * as snippets from './custom-snippets';
+import * as docMirror from './doc-mirror'
+import { testDataDir } from './extension-test/integration/suite/util';
```
bpringe: I think maybe this was imported accidentally or left after some temporary test? Probably can remove.

src/extension.ts
```diff
@@ -30,6 +30,9 @@ import config from './config';
...
+import { commands } from 'fast-check/*';
```
bpringe: This looks to maybe have been auto imported or something / mistakenly left as well.

src/extension.ts
```diff
@@ -65,6 +68,7 @@ function setKeybindingsEnabledContext() {
 }
 
 async function activate(context: vscode.ExtensionContext) {
+    vscode.commands.executeCommand('setContext', "@AFARK", "farked");
```
bpringe: Is this test code meant to be removed?

src/config.ts
```diff
@@ -1,7 +1,9 @@
 const config = {
     REPL_FILE_EXT: 'calva-repl',
     KEYBINDINGS_ENABLED_CONFIG_KEY: 'calva.keybindingsEnabled',
-    KEYBINDINGS_ENABLED_CONTEXT_KEY: 'calva:keybindingsEnabled'
+    KEYBINDINGS_ENABLED_CONTEXT_KEY: 'calva:keybindingsEnabled',
+    CURSOR_CONTEXT_IN_STRING: 'calva:inString',
```
bpringe: I think this could be a bit more descriptive, like `calva:cursorInString` just so we know it's about the cursor. Maybe that's obvious still, but doesn't hurt to be more explicit.

src/extension.ts
```diff
@@ -231,6 +236,7 @@ async function activate(context: vscode.ExtensionContext) {
     }));
     context.subscriptions.push(vscode.window.onDidChangeTextEditorSelection(event => {
         replHistory.setReplHistoryCommandsActiveContext(event.textEditor);
+        checkForContextChange(event.textEditor);
```
bpringe: Since this isn't just a pure check and does some side-effect, and since it's specifically about the cursor, I think a name like `setCursorContext` might be better. It can still do the check before setting, etc.

bpringe (dismissed):

bpringe:
Thanks for taking the time on this! I just requested a few small changes, but I also think this code should probably be moved out of `extension.ts`. I'm not sure where it should go, though. Let's see what @PEZ thinks.

crispinb:
Thanks - I'll make those changes here (sorry for the clumsy leavings!), but presumably forestall refreshing the PR 'till we hear back from Pez on the structure issue.

crispinb:
Oops. Pushed to my fork on autopilot. Anyway, those minor changes done.

PEZ:
Please keep pushing to your fork. We'll treat this as WIP until it's done and it is much easier if we all can look at and test the same thing.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 Changes to Calva.
 
 ## [Unreleased]
+- Fix: [Allow user to override keybindings in comments and/or strings](https://github.com/BetterThanTomorrow/calva/issues/1023)
```
```suggestion
- Fix: [Allow keybindings to target when the cursor is inside a comment or a string](https://github.com/BetterThanTomorrow/calva/issues/1023)
```

crispinb:
Yep it's all up to date with @bpringe's requested change

PEZ:
Looks good to me. The only thing that had me thinking was the `withinComment` function in the tokenCursor. Two things about it:

1. Does it belong in the TokenCursor? It can be argued it does, but so far it hasn't been needed and the only user is this cursor context janitor. It might be something that should be done there. That means exploiting a bit of the internals of the token cursor (that we know the token type it uses), which is horrible, but something we do all over the code base. (It's a refactoring of its own to use enums or maybe just plain constant strings for that). I'm not sure which way is best here, and I am fine with both. But if this rambling makes you think one or the other is best, @crispinb , then choose whichever you think is best.
2. ”within” is to me different than, say ”adjacent”. The check that is done now checks if it's either within or adjacent.

This second thing then touches the name of the context. I see two ways to deal with it:

1. We name the context something like `calva:cursorTouchesComment`.
2. We create more contexts for it. `calva:cursorTouchesComment`, `calva:cursorLeftOfComment`, `calva:cursorInComment`, `calva:cursorRightOfComment` 

Option 2 would allow to create keybindings that use structural behaviour in one direction and plain text behaviour in the other direction. Option 1 leaves option 2 open so I would pull this PR regardless which option you choose.

I think a good place to put this code is in a file at the root named `when-contexts.ts`. Then you can put the config constants in there as well. That `config` module is another room for Calva refactoring that we should deal with some day. But otherwise we most often just place constants in the modules themselves, and as long as they are well isolated in what they deal with, this seems to work well.

Once we have this in its own module we can get a bit fancier in future PRs. 😄 I've previously been a bit frustrated that I can't read the contexts from the code. So been thinking that maybe some Virtual DOM inspired approach would make sense. But that's for later! Haha.

crispinb:
A few thoughts about the above @PEZ 
- as to where the 'withinComment'  function belongs, I'll take your advice on that. You know  TokenCursor's raison d'etre in a way I don't (I just raided  the existing withinString function, like grabbing a googled SO answer!). So I've moved it out.
- on the  'adjacent/within/touches' comment issue. I'll push back a bit on this, but quite happy to defer if you still don't agree. I think having multiple comment-related contexts, at least now (in advance of being begged for them) is going to confuse matters for users and contributors. It could make for a horrible mess of a shortcuts JSON file, and if we followed that pattern elsewhere the contexts could proliferate madly. 
- you're right that the withinComments function checks for being within and 1 token beyond, but that was just to cope with EOLs . If you're editing a line comment, it feels very strange for keybindings to change when the cursor's at a line end (I tried it). How about this: in https://github.com/BetterThanTomorrow/calva/pull/1030/commits/bae3efffe7c3fafd60fdf0baa880e0614f9fc491 I've make the EOL inclusion more explicit. To the user (if not the token cursor or clojure reader??), a line comment would seem to include the EOL.
- constants also moved out as requested.

As I say, quite happy to revise any of that if you disagree, but I've pushed as-is so you can have a look.

src/cursor-doc/token-cursor.ts
```diff
@@ -670,6 +670,15 @@ export class LispTokenCursor extends TokenCursor {
...
 
+    /**
+     * Indicates if the current token is in a comment line
+     */
+    withinComment() {
+        const cursor = this.clone();
+        const token_type = cursor.getToken().type;
+        return token_type === 'comment' || (token_type === 'eol' && cursor.getPrevToken().type === 'comment');
+    }
+
```
PEZ: Now this seems to be in two places. 😄 That was good this time, because I think I know where I prefer it to be. Which is that we, for now, leave it as a method on the cursor itself. We can refactor this later to have a separate module where `withinString`, this one, and some others can live.

src/when-contexts.ts
```diff
@@ -0,0 +1,61 @@
...
+        case 'calva-standard':
+            break;
+        case 'comment':
+            vscode.commands.executeCommand('setContext', commentContext, true);
+            break;
+        case 'string':
+            vscode.commands.executeCommand('setContext', stringContext, true);
+            break;
+        default:
+        // TODO: throw? log?
```
PEZ: I don't see that we would ever end up here. Which might indicate that we should log it. `console.warn()`, might be a good fit.

src/when-contexts.ts
```diff
@@ -0,0 +1,61 @@
+import * as vscode from 'vscode';
+import { TokenCursor } from './cursor-doc/token-cursor';
+import * as docMirror from './doc-mirror'
+
+type CursorContext = 'calva-standard' | AltCursorContext;
+type AltCursorContext = 'string' | 'comment';
+const stringContext = 'calva:cursorInString';
+const commentContext = 'calva:cursorInComment';
```
PEZ: It is a bit messy in the code base regarding this. But I mostly try to indicate this type of constant by making the variable name all uppercase.

PEZ:
Sorry for late reply. This PR is going to make Calva a lot better, I think.

About the `cursorInComment` context. You make valid points. My initial take on it gets messy. However, I still have this concern:

```clojure
|a|(= (deref an-atom)
     @an-atom|b|)|c|
  |d|;; It's a common mistake to forget to|e| deref|f|
  [g|(|h|first an-atom)|i|
  (first @an-atom)
```

I just made up a syntax where there are cursor positions, `|a|`, `|b|`,  etcetera. 😄 I hope it will be possible to follow this. I am thinking that the default keybindings should make things behave liks this:

* `|c| forwardSexpr => |i|`
* `|d| forwardSexpr => |i|`
* `|d+1| forwardSexpr => |d+2|`
* `|f| forwardSexpr => |i|`
* `|d| backwardSexpr => |a|`
* `|f| backwardSexpr => |e|`
* `|g| backwardSexpr => |a|`

I don't have the time to test this build right now, but I think that it won't allow us to create this behaviour. I do think I know a much less messy way (than my last idea) to achieve it. I could be wrong, but my hunch says that if we have the contexts `cursorAtStartOfLine` and `cursorAtEndOfLine` in addition to a version of `cursorInComment` that doesn't include newline  we are home. We'll need to define `cursorAtStartOfLine` such that `|d|` is considered start of line, but I think that makes sense. Same for `|f|` and end of line, if there are trailing spaces. What'ya'say?

FYI, there is a property on the cursor named `rowCol` that might help you find leading whitespace. And... I think I might have created some code somewhere to get the text of the current line...

PEZ:
This should be covered with some unit tests, I think. To make that possible we need for the functions under test do not require  vscode knowledge, and even live in a module that does not import vscode. Without browsing the code base I am not sure what it entails. But if the non-vscode parts of this lives in the cursor-doc folder. I think the context-figuring functions should work an something implementing `EditableDocument` (which is defined in `cursor-doc/model.ts`). So the `mirrorDoc` you are creating already should work for this. The unit tests have some utility function that can make `MockDocuments` from a string. Iirc these implement `EditableDocument`s.

src/cursor-doc/token-cursor.ts
crispinb: Oops!

src/when-contexts.ts
crispinb: Compile time is the real risk here (missing a case for a newly added context). Not being a Typescript maven, I was a bit puzzled here as I thought it would do an exhaustiveness check. I've found a better way to deal with it.

crispinb:
@PEZ - thanks for spending the time to think that through. I think your `cursorAtStartOfLine/cursorAtEndOfLine/cursorInComment` notion is a good one. It gets around a central confusion between comments & lines in the current implementation (& yes, you're right that the behaviour you illustrate won't work currently). It's also atomic enough to probably be useful in conjunction with other contexts should they be added in future.

I find it hard to think comprehensively about this kind of thing in the abstract so I'll implement your approach here & try using it in working on some clojure stuff later today.

On unit tests - I haven't had a look through your existing test suite yet. Will do soon & then I'll take note of your comments here.

crispinb:
(That commit's not quite working as expected yet - still a wip)

src/cursor-doc/token-cursor.ts
```diff
@@ -676,7 +676,7 @@ export class LispTokenCursor extends TokenCursor {
     withinComment() {
         const cursor = this.clone();
         const token_type = cursor.getToken().type;
-        return token_type === 'comment' || (token_type === 'eol' && cursor.getPrevToken().type === 'comment');
+        return token_type === 'comment';
```
PEZ: Is it that either the token or the previous token is comment maybe? So both `|a|` and `|b|` here would be inComment:

```
|a|;; foo|b|<nl>
```

src/when-contexts.ts
```diff
@@ -1,56 +1,56 @@
...
 
-function determineCursorContext(document: vscode.TextDocument, position: vscode.Position): CursorContext {
+function determineCursorContexts(document: vscode.TextDocument, position: vscode.Position): CursorContext[] {
+    let contexts: CursorContext[] = [];
     const idx = document.offsetAt(position);
     const mirrorDoc = docMirror.getDocument(document);
     const tokenCursor = mirrorDoc.getTokenCursor(idx);
+    const line = document.lineAt(position);
+
+    if (position.character <= line.firstNonWhitespaceCharacterIndex) {
```
PEZ: I think vscode and Clojure might have different definitions of whitespace, (the comma.) You can use the token cursor instead here. So if the previous token is `ws` then `backwardWhiteSpace()` and then if the new position is on the previous row or the column is 0. Or something along those lines.

(I saw you said you're not done with this change yet, just throwing this in here since it is easy to forget about that comma.)

PEZ: I filed a feature request on vscode now. It really would be nice to solve this like you did there.

https://github.com/microsoft/vscode/issues/117174

src/cursor-doc/model.ts
```diff
@@ -8,7 +8,8 @@ export function initScanner(maxLength: number) {
 }
 
 /** A cheesy deep-equal function for matching scanner states. Good enough to compare plain old js objects. */
-function equal(x: any, y: any): boolean {
+// TODO: OK to export from here or move to util?
+export function equal(x: any, y: any): boolean {
```
bpringe: Probably best to move this to util

crispinb:
Excuse my absence here - I haven't forgotten about this. I'll be back to it sometime this week.

src/cursor-doc/token-cursor.ts
crispinb: My mistake - misinterpretation of what we had decided.

src/when-contexts.ts
crispinb: You're right  - hadn't thought of commas. Still something of a clojure rube really

src/cursor-doc/model.ts
crispinb: Done

crispinb:
Sorry it took a while to get back to this. It's still a wip. 2 remaining issues in addition to anything you spot:

1.  still to address unit tests

2.  the idea had been to use the new when contexts in customised shortcuts, adding clauses along lines like: `(!calva:cursorInComment || calva:cursorAtEndOfLine || calva:cursorAtStartOfLine)` 
Unfortunately it turns out this won't work, as (unintuitively!) [when clauses don't support parentheses](https://github.com/microsoft/vscode/issues/91473).  
I guess until that's addressed we'll  have to supply a few chosen composite when clauses (eg calva:cursorWithinCommentLine, which would be present when  calva:cursorInComment is true and calva:cursorAtEndOfLine and calva:cursorAtStartOfLine are both false)?

crispinb:
>     2\. we'll  have to supply a few chosen composite when clauses

I've added a POC of this - it adds 2 'compound' where clauses, `calva:cursorInCommentExcludingSOL` and `cursorInCommentExcludingEOL`. These can be used in keybindings like the following to achieve the behaviour @pez you described above:

```json
{
        "key": "ctrl+left",
        "command": "paredit.backwardSexp",
        "when": "calva:keybindingsEnabled && editorTextFocus && editorLangId == 'clojure' && paredit:keyMap =~ /original|strict/ && !calva:cursorInCommentExcludingSOL"
    },
    {
        "key": "ctrl+left",
        "command": "-paredit.backwardSexp",
        "when": "calva:keybindingsEnabled && editorTextFocus && editorLangId == 'clojure' && paredit:keyMap =~ /original|strict/"
    },
    {
        "key": "ctrl+right",
        "command": "paredit.forwardSexp",
        "when": "calva:keybindingsEnabled && editorTextFocus && editorLangId == 'clojure' && paredit:keyMap =~ /original|strict/ && !calva:cursorInCommentExcludingEOL"
    },
    {
        "key": "ctrl+right",
        "command": "-paredit.forwardSexp",
        "when": "calva:keybindingsEnabled && editorTextFocus && editorLangId == 'clojure' && paredit:keyMap =~ /original|strict/"
    },
```

I haven't been doing much clojure lately so this is imperfectly tested as yet.

PEZ (approved):

PEZ:
Looks great. I will be testing it today. I'm starting to think the default key bindings should be like the examples you have there.

PEZ:
Whoo, feels great! There are two quirky things I notice

1. forward from the end of line of a comment, jumps to the end of next sexpr, but backwards from there then jumps to the start of the previous sexpr. I don't think this is wrong, or anything we can really do anything about as long as we don't want to always step through comments. (Which I don't think I want.)
2. Jumping from the end of a comment line that has further comment lines, jump to the next sexpr. My brain expects it to jump past the next word on the next comment line. I think this can be achieved by defining `inComment` as being in a block of comment lines. So if at a  comment-EOL and the next non-whitespace line is also a comment, then we are `inComment`.

I'll be testing this further today, as I am writing my  *dram* thing mostly in comments I happen to be the perfect tester. 😄

crispinb:
Thanks for checking that out @PEZ . I had the same reaction re your 2nd point. It's all a bit semantically weird really -in effect kind of jumping between different languages. I haven't played around with the results much yet, but there's scope for using the new contexts with other keybindings (deletes etc).

Anyway let me know what else you come up with. I should be able to have a crack at some fixes/changes  Mon and/or Tue.

PEZ:
Yeah, totally interesting problem with the two languages. I like that with your work here we stop just ignoring that there are languages within the language, even if we'll have some boundary cases left (like my first observation above).

Also agree that the contexts will be of broader use. I've already added that for deletion on my branch. `alt+delete` is second nature to me, but it is weird to have to do that inside non-structured text

Btw, if I push that branch, will you be able to use that then? I've resolved the merge conflicts too. Might just try it and see what happens. Apologies in advance if it messes things up for you.

PEZ:
Seems it just became a regular branch. You can pull it in, if you like: https://github.com/BetterThanTomorrow/calva/tree/crispinb-wip/add-comment-string-when-clauses

bpringe:
> alt+delete is second nature to me, but it is weird to have to do that inside non-structured text

I just came across this (again?) recently. Glad we have this work being done by @crispinb!

PEZ:
I can pick this up from here if you have run out of time to spend on it, @crispinb . It seems close to finished and it is good stuff!

crispinb:
Hey @PEZ  - I did drop the ball a bit there (apologies - I don't want to be the annoying disappearing contributor). I won't get back to it for at least a week so please feel free if you have the time & inclination. If you do I'll see if there's something else I can pick up. Either way works for me.

PEZ:
Haha, don't worry. You are not an annoying contributor just because you have a life. 😄 

We'll see what the week brings. Today I realised I need your changes for fixing an issue, so if I find myself fixing that issue I will first finish this work. You'll see when you come back if the work is still waiting for you.

PEZ:
Continuing this in #1110. Please have a look, @crispinb and @bpringe . It has been a lot of late night hacking on this and mistakes might have happened.

PR #1042, opened 2021-02-22
Translate lsp module to cljs

### LINKED ISSUE #1025: Translate clojure-lsp integration to cljs
This is something I want to play around with to see what it's like to use cljs, not just for core functionality but also integration. The lsp related code is pretty sectioned off, not too large, and is my current area of work, so I think it's a good candidate.

bpringe:
## What has Changed?

- Added lsp.cljs to the cljs-lib and removed lsp.ts. Functionality remains the same and was tested.

Closes #1025

## My Calva PR Checklist

I have:


src/cljs-lib/src/calva/lsp.cljs
```diff
@@ -0,0 +1,220 @@
...
+                      extra-param (when extra-param-fn
+                                    (<p! (extra-param-fn)))]
+                  (when (or (not extra-param-fn) extra-param)
+                    (.. (.. client (sendRequest
+                                    "workspace/executeCommand"
+                                    (clj->js {:command name
+                                              :arguments (if extra-param 
+                                                           (conj params extra-param)
+                                                           params)})))
+                        (catch js/console.error))))))))]
```
bpringe, line 142: Note to self: test that this catch works as expected.

src/cljs-lib/src/calva/lsp.cljs
```diff
@@ -0,0 +1,220 @@
...
+    (.. client stop)))
+
+(defn get-references
+  [^LanguageClient client uri ^Position position include-declaration]
+  (.. client
+      (sendRequest "textDocument/references"
+                   (clj->js {:textDocument {:uri uri}
+                             :position position
+                             :context {:includeDeclaration 
+                                       (if (nil? include-declaration)
```
bpringe: Refactor to multi-arity function, passing true for this param. See if this works when called from js.

src/cljs-lib/src/calva/pprint/printer.cljs
```diff
@@ -1,43 +1,12 @@
...
-                   (is (> width 70))
-                   (is (<= width 80)))
-                 (is (not (re-find #"#" (:value (pretty-print deep nil))))))
-               (testing "Settings"
-                 (let [width (apply count (pretty-line-of 25 "foo" {:width 40}))]
-                   (is (> width 30))
-                   (is (<= width 40)))
-                 (let [width (apply count (pretty-line-of 25 "foo" {:max-length 2}))]
-                   (is (< width 20)))
-                 (is (re-find #"#" (:value (pretty-print shallow {:max-depth 1}))))))))}
```
bpringe, line 40: This was moved into the test directory in conjunction with removing the ns-regexp from the shadow-cljs :test target, so that an error wasn't thrown because "vscode" couldn't be found, due to lsp.cljs now living in the `cljs-lib/src/calva` directory.

bpringe: All cljs tests should go in the test directory now.

bpringe: I don't know/remember why that one test living in the src file caused an issue, but there are other tests in other source files that don't, so I left those, and they're still run by the test script.

shadow-cljs.edn
```diff
@@ -17,11 +17,16 @@
...
+                             :deactivateLsp calva.lsp/deactivate
+                             :getReferences calva.lsp/get-references
+                             :getDocumentSymbols calva.lsp/get-document-symbols
+                             :LSP_CLIENT_KEY calva.lsp/client-key}
                  :output-to "out/cljs-lib/cljs-lib.js"}
                 :test
                 {:target    :node-test
                  :output-to "out/cljs-lib/test/cljs-lib-tests.js"
-                 :ns-regexp ".*"
+                 :ns-regexp "-test$"
```
bpringe, line 30: With this set to `.*` as it was before, we'd get an error during the release script like "cannot find module 'vscode'" since adding the lsp.cljs file. With this regexp, we get the same number of tests and assertions as we currently do on dev.

bpringe: With regard to the above, I think that is because I moved a test from the pretty-print function to its own test namespace in order to make sure it was run.

PEZ: I really want to retain the possibility to have the tests inlined in the functions. This change looks like it is closing that. option. Can we factor te code such that the vscode using modules have something in their namespace names that we can exclude?

PEZ:
This is a tough one. Initially it will get harder for me to work with this code, with all the interop going on. Then I wonder what it will cost when we have bugs that we would have wielded the debugger against, which will not work with this, if things haven't changed in that respect. Oth we would have the REPL, which is great...

bpringe:
The interop is a small thing to get used to (kind of like getting used to parens and lisp structure when learning a lisp, but less so). I find in spite of the interop plus the lack of the node debugger it's still well worth using cljs, even for integration code. Having a repl replaces some of this, and even improves the experience beyond what we have with TS in other respects. It's a different methodology for sure, but I think the pros outweigh the cons.

This is a test, too. We can leave this module in cljs and see how we fair with it over time. This will give us a better idea of how things would work if more of the code base were in cljs. We have to gain more experience to know how it'll go, but I'm fairly confident in this way, or else I wouldn't have spent time on it.

shadow-cljs.edn
bpringe: We can still inline tests with this regexp. I don't know exactly how it works, but if you run the tests you'll notice the inline tests are run. I just added the test I moved out of the `calva.pprint.pprinter` namespace back into that namespace and re-ran the tests and verified that it ran, along with other inline tests in other namespaces. So this seems to work for what we need.

bpringe: I believe what it does is compile namespaces that either end in `-test` _or_ that have inline tests.

bpringe:
I plan to implement the downloading of the graalvm binary in cljs in this new lsp namespace as well. So I'll wait to see what you think before spending time on that. I hope you agree that this is at least worth trying. :smiley: I enjoy working in cljs, even where interop is involved, and I think fixing bugs could even be easier even though we don't have the debugger. We have inline defs, a repl, etc.

It also seems easier to experiment with editor side-effects from the repl, as we can just call `(.. vscode (showInformationMessage "test"))` etc. (but rather, a more useful usage, register a command from the repl to test it, apply an edit, etc.)

PEZ:
That sounds like a good thing to try. I think we should push forward and persue this. Also like the idea to pick a module that you are familiar with and use it for the test. We should maybe add a module where we have more issues than in this one to the test so that we get to feel how it is to fix bugs.

PEZ:
Used the shadow-cljs inspector a bit at work today. So awesome. I want to hack on the formatter a bit now. Haha. =)

bpringe:
I didn't know about the inspector but just looked into it. Really cool.

PR #1046, opened 2021-02-25
Fix cljs js interop release issue

### LINKED ISSUE #1044: State issue with released cljs/js interop
This issue didn't pop up in dev, and unfortunately I didn't catch it before I released.

From what I can tell, the issue is related to how we output the code for release, and it revolves around the state module, though it could pop up elsewhere. In order for the `js/require` calls to work, the individual .js files for those modules need to exist, so I added a step to the release script to compile with tsc before bundling with webpack. However, it seems this causes an issue in that two different modules can be used for the same intended module. 

I believe the js code uses the bundled state module, but the cljs code uses the unbundled state.js file, so when js updates a state value, cljs can't see the update, and vice versa. This leads to issues such as the cljs not being able to see if the repl is connected (because that's updated from js) and the js not being able to get the lsp client from state (because that was set in cljs). Put simply, there's two versions of state, one for js and one for cljs.

I think there's a way to solve this, besides writing the state module in cljs, which is an option, but it'll require some thought.

bpringe:
## What has Changed?

- Stop bundling with webpack so that the same modules are always used by both cljs and js

Fixes #1044 

## My Calva PR Checklist

I have:


bpringe:
@PEZ By not bundling with webpack and including the necessary node modules in the vsix, the vsix size comes in at  ~80 MB. Once we switch to downloading clojure lsp and remove clojure-lsp.jar, it'll be reduced to ~53MB.

This is the simplest approach to ensuring that the same modules are shared between cljs and js in all cases. Over time if we continued down the cljs path, we'd have fewer js files and the vsix will decrease in size (though this size is already not a big deal in my opinion, as far as I can tell).

There is a warning when packaging with vsce that the vsix includes a lot of files and a recommendation to bundle, but like I said above, more would be bundled over time as more is written in cljs, and I don't see any issues from this anyway.

I left in the webpack config just in case we want to switch back in the near future.

bpringe:
Those sizes I mentioned above actually were for the `extension` folder inside the vsix. The vsix itself is smaller. The one I downloaded from CircleCI for this PR was ~50 MB.

PEZ:
I think we should investigate more how we can keep bundling the extension in fewer files. There are probably good reasons vsce warns like that.

bpringe:
Ok. I've spent some time investigating how we can keep bundling, including the shadow :esm stuff, which I don't think would help us much here anyway as we might still have two different bundled versions of modules.

This is from the vscode extension docs:

>Visual Studio Code extensions often grow quickly in size. They are authored in multiple source files and depend on modules from npm. Decomposition and reuse are development best practices but they come at a cost when installing and running extensions. Loading 100 small files is much slower than loading one large file. That's why we recommend bundling.

It's merely an install and load time benefit, the difference of which is hardly noticeable at all in our case The difference in install time I'd say is noticeable but it's still a small amount of time in total (maybe 3 to 5 seconds). I don't notice any issue with the performance or loading of the extension though. And as stated, we'd aim to decrease this size over time, bundling more with shadow as more is written in cljs.

But there's one more thing that I know of to try, and that's using multiple [entry points](https://webpack.js.org/concepts/entry-points/) in the webpack config so it'll produce individual js files for our cljs to consume. I don't know if this would work for what we need, though. The problem I see here is that while we may only have to do this for the state module, we still leave ourselves open to the same issue happening again with another module, in which case when we discover the issue, we'd have to add another entry for that module.

If the above doesn't work, then we're left having to rewrite the state module in cljs, and I'm not sure how complicated that would be, or if it would in turn require other rewrites, amassing a larger rewrite that we probably don't want to do.

PEZ:
I do not agree about ”hardly noticable”. Loading Calva goes from being instant to taking quite a while. Maybe it is worth it, but it does add to the cost of moving vscode integration into cljs land.

I have my doubts about how much we will/should prioritize porting code in favor of adding user value. Seriously doubt we will be able to port the Token Cursor stuff in a maintainable way at all. We would have to complete the porting 100% until we can stop bundling the whole `node_modules` directory....

However. We should continue the experiment, as we do both think we have big benefits to reap from having the REPL and the dynamics of Clojure. If the Webpack options seem exhausted, we probably should run a while without bundling and see what it feels like. Maybe we will rediscover why I felt it necessary to start bundling. Maybe not. 😄

bpringe:
How are you measuring load time? I may not be doing so accurately. In current Calva, if I have a .clj file open and I reload the VS Code window, it takes about 1 second for the rainbow highlighting to show (with me clicking around the editor to make them activate), and with the vsix from this PR, it takes about 2 seconds.

If looking at when the Calva says output shows, it's the same for both this vsix and current Calva, for me at least - 3 seconds. There may be a much better way to measure load time though :smile:.

> I have my doubts about how much we will/should prioritize porting code in favor of adding user value. 

While I agree that user value is the most important thing, I think we can accomplish that better in cljs, even if not just with our own contributions but with those from others. I'm not advocating for a big rewrite though. My strategy would be along the lines of a type of freeze on the current typescript code like:

* If adding a new module/file, make it cljs, so we don't expand the current use of ts.
* Bug fixes of ts code stay in ts, of course, unless it involves rewriting a large portion of a ts module, in which case use cljs
* Additions to existing ts modules still happen in ts too, unless it's a large addition that can be extracted to another module, in which case use cljs and import it from the ts

> We would have to complete the porting 100% until we can stop bundling the whole node_modules directory....

Just so you know, not all the node_modules files are added to the vsix. I don't know what determines which ones are added, but in the vsix for this PR only 164 items are in node_modules, whereas in dev there are 786. The directory sizes are 40 MB in the vsix and 213 MB in dev. Maybe this is something to figure out, but I haven't seen any issues. I was thinking maybe only specific/imported modules are being included, but I don't know what is deciding that.

Anyway, I'll check out the multiple entry points thing with webpack.

bpringe:
Update: Multiple entry points does not seem work for our situation (or be what we need). I'll be looking into `externals` in the webpack config next.

bpringe:
#1053 resolved this

PR #1053, opened 2021-03-03
Use cljs for state

### LINKED ISSUE #1052: Use cljs for state
This is an improvement that will help with #1044, but I want to isolate this change so it's easier to manage.

bpringe:
## What has Changed?

- Adds cljs state namespace to be used for ephemeral state, in place of Immutable.js
- Moves some things out of state.ts, but ultimately I left the rest as I want to keep these changes as minimal as possible. There are already enough other changes :smile:
- Moves some things out of namespace.ts into a new repl-session.ts. This was to avoid the cljs lib issue noted below, before I found a workaround, but I think the changes make sense so I left them.

### The cljs lib issue

In two files that I've seen so far, annotations.ts and namespace.ts, if we import the cljs lib directly and call functions from it, during runtime we get an error that the function is not defined, even though they work in many other files. If we bundle the extension for release and run the vsix, it works fine. Something is not being linked correctly, but I have not figured out the root of the issue.

#### The workaround

I've found that if we export the cljs lib from utilities.ts, then import it from utilities.ts in those troublesome files noted above, the function calls work fine during runtime. I tried to make a separate cljs-lib.ts and import+export it from there, then import the lib from that module, but again it doesn't work. Since I know it works from the utilities module I left that workaround in place for now. 

Maybe later we can figure out the root issue, but at some point it might not matter anyway. In the meantime I'm not sure if we should make all imports of the cljs lib come from the utilities module just to be safe. There are many places in the code where we call functions from the cljs lib now since we use it for state, and I haven't yet verified every single one of those calls works. If we discover some don't, we can switch it to be imported from the utilities module.

Closes #1052 

## My Calva PR Checklist

I have:


bpringe:
@PEZ I'm having a pretty odd issue that I can't yet figure out. It seems that any function from the cljs-lib called inside the namespace.ts function `updateREPLSessionType` throws an error on Calva loading like:

```
Activating extension 'betterthantomorrow.calva' failed: parseForms is not a function.
```

This originally was happening for the new `getStateValue` function, so I thought I did something wrong, so then I started trying things and noticed I can call that function right at the beginning of the extension.ts `activate` function and it works fine, but if I call it from that namespace.ts function I get that error. Something about it being called from that location / point in time is causing the issue.

[This branch](https://github.com/BetterThanTomorrow/calva/tree/use_cljs_for_state_troubleshooting) has my troubleshooting code with just minimal changes from dev. If you could take a look and see if you see anything obvious that would be great. This is really odd to me, especially since `parseForms` is called elsewhere in that file, and that's been that way. :thinking:

bpringe:
You can see in the Log (Extension Host) output the stack trace of the error when trying it out. I also noticed that if I move the calls to the cljs functions up one level in the call stack to state.ts/`update` ([here](https://github.com/BetterThanTomorrow/calva/blob/c8543d7a35bbd067597740210a149554d97a93bd/src/status.ts#L16)), they work fine. So that seems to narrow it down more to being something about namespace.ts that throws off the execution of the cljs-lib functions.

bpringe:
If we can't figure out why the cljs functions won't run inside `updateREPLSessionType`, since I've verified they do work in the calling code, we can just pass a boolean to that function [for `connected`](https://github.com/BetterThanTomorrow/calva/blob/c8543d7a35bbd067597740210a149554d97a93bd/src/namespace.ts#L114), getting it from state beforehand, removing the need to make the call in the function.

bpringe:
This is really odd. If I copy all of namespace.ts into another file, namespace2.ts, and I change the status.ts `update` function to this:

```javascript
function update(context = state.extensionContext) {
    vscode.commands.executeCommand('setContext', 'calva:showReplUi', shouldshowReplUi(context));
    namespace2.updateREPLSessionType();
    //namespace.updateREPLSessionType();
    statusbar.update(context);
}
```

The call to `getStateValue` in `updateREPLSessionType` works, but then if I go and update all imports of `namespace` to `namespace2` instead, the call fails again.

```
Activating extension 'betterthantomorrow.calva' failed: cljs_lib_1.getStateValue is not a function.
```

bpringe:
Just noticed it works fine when I package it as a vsix and install it, it just doesn't work in dev :thinking:. Must be something weird with bundling/compiling going on.

bpringe:
Just want to add here that this problem occurs even on `dev`, without any of this new cljs state stuff added, so it's not anything caused by that, at least. If you call `parseForms` in namespace.ts/`updateREPLSessionType`, it throws a similar error to one mentioned above, saying that `parseForms` is not a function.

bpringe:
@pez I've worked around the above issue by a little refactoring. I also posted an [issue in the shadow repo](https://github.com/thheller/shadow-cljs/issues/849) about it to see if Thomas has any insights, but I think we can move forward now.

If you wouldn't mind looking over these changes and letting me know if you'd like any changes that would be great. Definitely want to test this a good bit before merging.

I also still need to split up the state module / refactor it into other modules. I don't plan to change the code there though. I want to change as little as needed for this PR, but I think leaving the state.ts module named / grouped as it is could be misleading.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 Changes to Calva.
 
 ## [Unreleased]
+- [Use cljs for state](https://github.com/BetterThanTomorrow/calva/pull/1053)
```
PEZ: I don't think it belongs in the changelog really, since to the user things should stay the same. But it gets relevant if something break. Then our users (since they are programmers) could connect the dots and help us find issues.

All in all, I think it should stay, but be marked as being of internal tech nature.
```suggestion
- Implementation detail: [Use cljs for state](https://github.com/BetterThanTomorrow/calva/pull/1053)
```

src/calva-fmt/src/extension.ts
```diff
@@ -1,15 +1,15 @@
 import * as vscode from 'vscode';
-import * as state from '../../state';
 import { FormatOnTypeEditProvider } from './providers/ontype_formatter';
 import { RangeEditProvider } from './providers/range_formatter';
 import * as formatter from './format';
 import * as inferer from './infer';
 import * as docmirror from "../../doc-mirror/index"
 import * as config from './config'
+import { documentSelector, getWorkspaceConfig } from '../../config';
```
PEZ: I have come to prefer `config.documentSelector` for most things like this.

src/config.ts
```diff
@@ -1,13 +1,68 @@
...
+ * @param  {string} name
+ * @return {string} The trimmed name
+ */
+ function _trimAliasName(name: string): string {
+    return name.replace(/^[\s,:]*/, "").replace(/[\s,:]*$/, "")
 }
 
-export default config;
+// TODO find a way to validate the configs
+function getWorkspaceConfig() {
```
PEZ: This should probably just be called `getConfig()`, as it is not just workspace configs in here.

src/extension.ts
```diff
@@ -26,16 +26,18 @@ import * as debug from './debugger/calva-debug';
 import * as model from './cursor-doc/model';
 import * as outputWindow from './results-output/results-doc';
 import * as replHistory from './results-output/repl-history';
-import config from './config';
+import  { KEYBINDINGS_ENABLED_CONTEXT_KEY, KEYBINDINGS_ENABLED_CONFIG_KEY,
```
PEZ: Again. I think we should rather pick it up from `config.` explicitly in the code below.

src/repl-session.ts
```diff
@@ -0,0 +1,63 @@
+import { NReplSession } from "./nrepl";
```
PEZ: The creation of this module and separation of these things out of namespace.ts make perfect sense to me. Question, is this required for this PR, or just a tidying up of things? I'm just curious. I might have forgotten som thing you have told me about this. Also, maybe it should go into the `nrepl` subdir? Not sure about that, so your call.

PEZ (approved):
This is great stuff. I have a few nitpicks, that's all.

src/config.ts
bpringe: It's actually only workspace config that we're dealing with here, that's why I named it that way.

```javascript
const configOptions = vscode.workspace.getConfiguration('calva');
```

bpringe: But if you plan to maybe use this for other config in the future, then it makes sense to change it.

src/repl-session.ts
bpringe: It started out as me moving things out of namespace.ts where the calls to cljs lib functions were failing, but after the workaround it's not necessary. I can move the functions back to namespace.ts if you want to keep changes minimal, but seems you may be okay with leaving them separate? I'll consider moving to the nrepl subdir. That may be wise.

src/config.ts
bpringe: Maybe I misunderstand here though. I was just going off `vscode.workspace.getConfiguration` and the fact that we only pull things from that thereafter, though I see `.globalValue` at one point, so I'm unsure.

src/repl-session.ts
bpringe: Moved repl-session to the nrepl directory.

src/config.ts
PEZ: For custom command snippets we merge the user config and the workspace config, since you can have commands that you want globally and commands that only make sense in a certain project. Therefore it is not just workspace config, as I see it. And also for the other settings, if there is no workspace config for a setting, the user config will be used. I think this is just config.

bpringe: Got it! Thanks. Will change.

PR #1068, opened 2021-03-15
Remove nrepl definition provider.

bpringe:
@PEZ I continue to run into the "is not a function" issue while trying to complete the native binary PR. Since I've yet to figure out the root cause, I've been working around it. This time it required a bit more extensive rework. The error was occurring when the lsp middleware calls `provideClojureDefinition`, because that calls `util.getConnectedState`, which calls into the lsp state module (then the "is not a function error" shows up).

I know you may not want to only provide definitions from lsp, so I separated this change out so we could discuss.

I really want to figure out the root cause of the issue, but it's been blocking my progress on work for so long that I'm trying to at least get this native binary PR out with workarounds. But maybe I should just shelve the change until we can figure out the root cause of the interop issue?

Btw, the issue is only occurring with packaged vsix's now, which is the opposite of what happened before - really strange.


PEZ:
I don't think we dare release until we know the root cause. Seems like any change can cause it to stop working either in dev or the VSIX or both.

And you are right about me not wanting to lose the definitions that only nREPL provides. That would really be sacrificing user value over something where the user value is not all that obvious.

bpringe:
> And you are right about me not wanting to lose the definitions that only nREPL provides

Ah yes, I forgot nrepl provides some definitions lsp does not.

Back to the drawing board with the issue then.

PR #1070, opened 2021-03-17
Use :npm-module shadow target

bpringe:
@PEZ I've come up with a new solution to the interop problem, but it also comes with trade-offs.

1. We import the functions from their cljs namespace js file and call them by their compiled js name, so hyphens are replaced with underscores. I actually kind of like this approach because we don't have to grow an export list like with the :node-library target.
2. We have to run shadow-cljs `compile` instead of `release` for the release build. When running `release`, for some reason I see some outputted js files with nothing in them, like the calva.state.js file. This of course resulted in an error that a function from that namespace was not found. I'm not sure why that occurs, but using `compile` instead, like in dev, makes things work with the release build.
    - As a result of using `compile` (and even with `release` with this target, even if it worked) , there are many files produced in the `out` directory, so `vsce` warns that: `This extension consists of 495 files, out of which 220 are JavaScript files. For performance reasons, you should bundle your extension: https://aka.ms/vscode-bundle-extension . You should also exclude unnecessary files by adding them to your .vscodeignore: https://aka.ms/vscode-vscodeignore`. This number is a bit lower with `release` but we still get the warning, and the release build doesn't work anyway.

Overall, the vsix is 15 MB, but inside it, the extension directory is 42 MB. I tried sending the shadow module target output to `node_modules`, knowing webpack should bundle them in like it does with other deps, but with that, requiring our js files from cljs doesn't work because it's a relative path and the cljs file is no longer in the `out` directory.

I know this is probably not an acceptable solution, but I just wanted to show you the approach. I may copy this text over to the issue for the record, and consider the options left. Unfortunately, there may not be a reasonable solution left besides a large rewrite, and I don't intend to do that. I may have to give this up, but I'll think a bit more on it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1960,7 +1960,7 @@
         "watch": "tsc -watch -p ./tsconfig.json",
         "release-cljs": "npx shadow-cljs release :calva-lib :test",
         "update-grammar": "node ./src/calva-fmt/update-grammar.js ./src/calva-fmt/atom-language-clojure/grammars/clojure.cson clojure.tmLanguage.json",
-        "release": "npm i && npm run clean && npm run update-grammar && npm run release-cljs && tsc -p ./tsconfig.json && webpack --mode production",
+        "release": "npm i && npm run clean && npm run update-grammar && npm run compile-cljs && tsc -p ./tsconfig.json && webpack --mode production",
         "compile-test": "tsc -p ./",
         "postrelease": "",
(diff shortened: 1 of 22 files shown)


bpringe:
For the record I do not think this should be used / merged :smile:. I'll go ahead and close it now, though you can still view it. @PEZ

PR #1087, opened 2021-03-31
Make find test under cursor handle namespaced def macros

### LINKED ISSUE #1086: Calva not detecting tests with aliased `clojure.test` namespace
Calva is unable to detect tests if created with aliased `t` namespace instead of direct `refer`:

```clojure
(ns ...
      (:require ...
            [clojure.test :refer [deftest is] :as t]))

(deftest foo
  (is (= 1 1)))

; Running test: foo…
; 1 tests finished, all passing 👍, ns: 1, vars: 1

(t/deftest bar
  (t/is (= 1 1)))

; No test found at cursor
```

PEZ:
## What has Changed?

We had some pretty complicated way of finding ”the current test”, which also fails when using namespace qualified definition macros. Fixing that here, reusing the ”current top level definied thing” that we use for custom code snippets. It besically disregards whatever is in the function position of the toplevel form and just picks the next non-meta symbol. I can't see when this should be worse than what we previously did, and it certainly is better for the namspace qualified macro case.

Fixes #1086

## My Calva PR Checklist

I have:

Ping @bpringe


bpringe:
I noticed that `Run Current Test` on the `with-test` macro does not run the test with this vsix, but it does in current Calva.

```clojure
(with-test
  (defn foo []
    #{{:id           1010101
       :display-name "Campaign 1010101"
       :state        "enabled"}})
  (is (= #{{:id 1010101
            :display-name "Campaign 1010101"
            :state "enabled"}}
         (foo))))
```

Output:
```clojure
; Running test: defn…
; No tests found. 😱, ns: 0, vars: 0
```

PEZ:
Good catch!

PEZ:
Can you test this again, @bpringe ? And extra points if you have a look at the code again, of course. 😄

src/util/get-text.ts
```diff
@@ -41,18 +42,11 @@ export function currentFunction(editor: vscode.TextEditor): SelectionAndText {
...
-                const symbol = tokenCursor.getToken();
-                if (symbol.type === 'id') {
-                    return [selection, symbol.raw];
-                }
-            }
+        const cursorOffset = document.offsetAt(editor.selection.active);
+        const tokenCursor = docMirror.getDocument(document).getTokenCursor(cursorOffset);
+        const [range, text] = cursorTextGetter.currentTopLevelFunction(tokenCursor);
+        if (range) {
+            return [select.selectionFromOffsetRange(document, range), text]; // TODO: Figure out which selection to return...
```
bpringe: I don't understand this TODO. Is it still a todo? Looks like there's only one selection to return.

bpringe (approved):
Looks good to me. Just had a question about a code comment.

bpringe:
Also, I tested it and it works fine for me.

src/util/get-text.ts
PEZ: Thanks. Yeah, we were previously returning the text of the function name, but the selection of the whole function. Which seemed really weird, but I didn't know if that was intentional. I don't think it was, and for this particular function no-one is using the selection so I fixed it. And forgot the TODO. 😄

Issue #1079, opened 2021-03-24
Auto deletion of closing paren on deletion of opening paren in shorthand function syntax causes newline to be removed


bpringe:
![auto_removal_newline](https://user-images.githubusercontent.com/12722744/112347646-45d39d00-8c84-11eb-8781-358719883d3a.gif)

Not sure if this is related to #1024.

PEZ:
As always, text repro's appreciated. 😄 

```clojure
{::foo #(|)
 ::bar :foo}
```

<kbd>backspace</kbd>

**Expected:***

```clojure
{::foo |
  ::bar :foo}
```

**Actual:***

```clojure
{::foo #|::bar :foo}
```

And no, not related to that feature request. Which I would also love to implement, btw.

PEZ:
This also affects <kbd>delete</kbd>.

Issue #650, opened 2020-05-16
Prevent extra closing brackets in strict mode


gerred:
Currently, if you open a form, the editor gives you the closing bracket as expected:

`()`

However, in either mode, you currently can add more closing brackets to this. These brackets are then not deleted (despite there being no form to balance) unless you perform a force delete:

`()))))]])} ; backspacing or deleting over these will not remove the unbalanced trailing brackets!`

Should this automatically work in strict mode? I go back and forth on this, as force delete gives me a mechanism to still do what I want.

PEZ:
My gut thinks it makes sense that strict mode prevents this unbalance to enter. Force delete is easy to use and all, but it should be a last resort and not against something Calva has almost helped happen.

Is there a case where someone would want to add extra closing brackets? If it is not a common case that someone could use Cave Man mode to enter them and then switch back to Strict, right?

PEZ:
I added the `fun-task` label, because I think this would be pretty entertaining to implement. I'm available to help get you started via Zoom or something if you want to take it on.

PR #1099, opened 2021-04-03
Assorted Paredit quality fixes

### LINKED ISSUE #1098: Paredit slurp sometimes leaves an extra space
I think the moved paren might be replaced by a space:

```clojure
(s|tr )#(foo)'
```

<kbd>slurp</kbd>

**Expected**

```clojure
(s|tr #(foo))
```

**Actual**

```clojure
(s|tr  #(foo))
```

PEZ:
- [x] [Replace extra whitespace with one when slurping](https://github.com/BetterThanTomorrow/calva/issues/1098)
- [x] [Delete empty literal function causes newline to be removed ](https://github.com/BetterThanTomorrow/calva/issues/1079)
- [x] Add experimental setting to: [Prevent extra closing brackets in strict mode](https://github.com/BetterThanTomorrow/calva/issues/650)

Fixes #1098

There are some serious changes in paredit `backspace` and also generally. I've added tests to guard against the regressions I could think up.

My testing says that the `backspace` changes should be for the better. Deleting an extraneous closing bracket now doesn't require force-backspace.

For the prevention of adding unbalanced closing brackets this applies:

> Adding a setting to prevent inserting unbalanced closing brackets. As it isn’t really possible to do this in VS Code, it is a bit clumsy, the bracket is added (since we can’t stop it), then removed, then only added if there is unbalance in the document. The rational being that if there is unbalance, all bets are off, the human at the keyboard is best suited to heal it. The solution doesn’t behave well when it is a selection that is replaced with the closing bracket, so don’t do that if you have this setting enabled.

(And with the backspace fixes even leaving this setting disabled should not be as painful as it used to be.)

TODO:

- [x] Give `delete` the same love as `backspace` has gotten.
- [x] Document the new setting

## My Calva PR Checklist

I have:

Ping @bpringe

### DIFF
--- a/src/extension-test/unit/cursor-doc/token-cursor-test.ts
+++ b/src/extension-test/unit/cursor-doc/token-cursor-test.ts
@@ -1,6 +1,7 @@
 import * as expect from 'expect';
 import { LispTokenCursor } from '../../../cursor-doc/token-cursor';
 import * as mock from '../common/mock';
+import { docFromTextNotation, textAndSelection } from '../common/text-notation';
 
 describe('Token Cursor', () => {
     const docText = '(a(b(c\n#f\n(#b \n[:f :b :z])\n#z\n1)))';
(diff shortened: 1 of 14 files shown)


src/extension-test/unit/cursor-doc/token-cursor-test.ts
```diff
@@ -76,11 +77,26 @@ describe('Token Cursor', () => {
...
         });
 
+        xit('Does not move past unbalanced top level form', () => {
+            //TODO: Figure out why this doesn't work
+            const d = docFromTextNotation('|(foo "bar"');
+            const cursor: LispTokenCursor = d.getTokenCursor(d.selectionLeft);
+            cursor.forwardSexp();
+            expect(cursor.offsetStart).toBe(d.selectionLeft);
+        });
+        xit('Does not move past unbalanced top level form', () => {
```
PEZ, line 87: @bpringe, this is the test that breaks coming tests. Something runs amok when there is no closing quote. But the way it runds amok!

src/extension-test/unit/cursor-doc/token-cursor-test.ts
```diff
@@ -76,11 +77,26 @@ describe('Token Cursor', () => {
             expect(cursor.offsetStart).toBe(14);
         });
 
+        xit('Does not move past unbalanced top level form', () => {
+            //TODO: Figure out why this doesn't work
```
PEZ, line 81: This is utterly strange. It actually does work in the vscode editor, but not like this. I suspect it might have to do with the ”runs amok@ thing, since also here we have a missing close. But that is just a wild guess.

src/extension-test/unit/cursor-doc/token-cursor-test.ts
bpringe: Weird, so, maybe something to do with the `"` after `bar`, whereas the one above doesn't have that? No clue what it could be though.

PR #1100, opened 2021-04-03
Add projectRootDir for monorepo connection sequences

### LINKED ISSUE #579: Add option to manually specify project root directory
When working in a monorepo project, it's useful to manually specify which directory is the project root.  There should be a property added to replCOnnectSequences that allows you to set this project root directory.  This value will override the project root found by `initProjectDir()`.

stefan-toubia:
## What has Changed?
Adds a `projectRootDir` option for custom connection sequences to specify a root dir to use instead of resolving the root dir based on the current open file context.

Fixes #579

## My Calva PR Checklist

I have:


stefan-toubia:
There's still one issue to resolve with this, currently when you jack-in the results doc is created before we prompt for the connection sequence and potentially change the project root dir based on this new setting. This is causing the results doc to become disconnected due to the directory change. There's a couple potential solutions:

1. Reorder the sequence within `jackIn` so that `initResultsDoc` is only called after we get the connection sequence and potentially reinitialize the project dir. This would require changing the 'aborting jack-in' message since this currently writes out to the window. This could be an alert popup instead.
2. Instead of picking the results doc output location based on project root dir, make the location static. This could be based on the workspace folder, a user-defined path in settings, or just cache the first file path returned by `outputFileDir`. This would also solve the issue of there being multiple output file windows open at a time, which seems like an improvement to me but I'm not sure if anybody finds it beneficial to have more than one results doc open at a time.

My preference would be 2 since it also solves the multiple results doc issue, WDYT?

PEZ:
I think option 2 would be strange when you are working with more than one project. And option 1 sounds like it would risk opening a can of worms.

Do we have other options?

bpringe:
> I think option 2 would be strange when you are working with more than one project

@PEZ Can you elaborate more on this? I don't work with multiple projects simultaneously much, so maybe it's hard for me to see the issue. Do you mean contextually? Like, I'm working on A and B, and I run some things in A and then switch to B and run some things, and now there's output from two different projects in the output window, but it's unclear from which project they came?

If that is the case, I don't see that being an issue, since output is output and happened in the past. I may be way off here from what you're thinking though. Maybe you also mean the context for working directly in the output/repl window could be confusing ("which project am I in?") - though there is the namespace in the prompt, at least.

PEZ:
@bpringe you basically elaborated it to what I was thinking about. The difference in viewpoint being that I see a huge issue with it. 😄 I have five to ten different projects open almost all the time and having some system.log kind of output would be awful.

I think that can of worms (option 1) is interesting. We've pulled off similar things in the past. Maybe it is not as scary as I first thought it was.

bpringe:
> I have five to ten different projects open almost all the time and having some system.log kind of output would be awful.

Yeah, I see. Good point.

stefan-toubia:
@PEZ did you intend to delete this branch?

bpringe:
I was wondering the same. I know he did some cleanup earlier. :thinking:

PEZ:
My apologies! Not intended. At. All.

package.json
```diff
@@ -3,7 +3,7 @@
     "displayName": "Calva: Clojure & ClojureScript Interactive Programming",
     "description": "Integrated REPL, formatter, Paredit, and more. Powered by cider-nrepl and clojure-lsp.",
     "icon": "assets/calva.png",
-    "version": "2.0.185",
+    "version": "2.0.185-monorepo-proj-root-dir",
```
PEZ, line 6:
```suggestion
    "version": "2.0.185",
```
(I usually pick up the VSIX from CircleCI instead. Maybe we should make a script that build a local VSIX...)

package.json
```diff
@@ -373,6 +373,11 @@
                                     "description": "Here you can give Calva some Clojure code to evaluate in the CLJ REPL, once it has been created.",
                                     "required": false
                                 },
+                                "projectRootDir": {
+                                    "type": "string",
+                                    "descripion": "The root directory to execute the Jack-in command from.  This will override the default behavior for determining the root directory relative to the current open file.  This is useful for working in monorepos.",
```
PEZ, line 378: FYI (and this should go to the dev docs): I've started to use `markdownDescription` to boost the schema help a bit.

src/nrepl/connectSequence.ts
```diff
@@ -208,7 +209,10 @@ function getConnectSequences(projectTypes: string[]): ReplConnectSequence[] {
     const customSequences = getCustomConnectSequences();
     const defSequences = projectTypes.reduce((seqs, projectType) => seqs.concat(defaultSequences[projectType]), []);
     const defSequenceProjectTypes = [...new Set(defSequences.map(s => s.projectType))];
-    const sequences = customSequences.filter(customSequence => defSequenceProjectTypes.includes(customSequence.projectType)).concat(defSequences);
+  const sequences = customSequences.filter(customSequence =>
+      !!customSequence.projectRootDir
+      || defSequenceProjectTypes.includes(customSequence.projectType)
+  ).concat(defSequences);
```
PEZ, line 215: I don't quite remember where the result from this function is used. Can you explain why you filter away sequences with the setting? (I'm sure it is the right thing to do, just want to understand why. 😄 )

PEZ:
Good morning. Again my apologies for deleting the branch. Thanks for restoring, @bpringe!

I left some comments in the review. Overall it looks good. I wonder if you have tested this with the **Connect to a running REPL** scenario? I expected that there would need to be some changes in the `connect.ts` module as well.

mrkam2:
@stefan-toubia bump

PR #1132, opened 2021-04-19
Restore the right selection after eval to comment

### LINKED ISSUE #1131: calva.evaluateTopLevelFormAsComment results in vim Visual mode
When using Calva with VSpaceCode, doing "_evaluate top-level form as comment_" (, e ; in VSpaceCode) leaves me in Vim Visual mode after the comment has been pasted.
Is there a setting I can update so that _evaluate top-level form as comment_ leaves me back in Normal mode?

Some dicussion from #calva on Slack:
> practicalli_john:
> Seems like the selection used by Calva keeps the form selected in VSpaceCode, e.g in visual select mode.
In that case a manual press of the Esc key is required it seems, unless there is a way to send an unselect at the end of the call to that function.
> ...
> It seems strange, because the previous two functions will leave the expression selected.
However, calva.evaluateSelection and calva.evaluateCurrentTopLevelForm do not leave the expression selected.

>pez:
>Indeed:
```
    vscode.workspace.applyEdit(wsEdit).then((_v) => {
        editor.selection = selection;
    });
```
> I wonder why I thought that was a good idea…

I don't know much about VSCode extensions, but I'm happy to try and help out or test possible solutions.

PEZ:
## What has Changed?

Seems we had a regression in the code for evaluating to comment. I've restored the previous behaviour of restoring the previous editor selection after having inserted the comment.

While at it I also made it decorate the evaluated form, just as with ”normal” evaluation. It is a bit erratic, but still good to see what it was that was evaluated, I think.

Fixes #1131 

## My Calva PR Checklist
I have:

Ping @pez, @bpringe

### DIFF
--- a/src/evaluate.ts
+++ b/src/evaluate.ts
@@ -42,9 +42,8 @@ function interruptAllEvaluations() {
 async function addAsComment(c: number, result: string, codeSelection: vscode.Selection, editor: vscode.TextEditor, selection: vscode.Selection) {
     const indent = `${' '.repeat(c)}`, output = result.replace(/\n\r?$/, "").split(/\n\r?/).join(`\n${indent};;    `), edit = vscode.TextEdit.insert(codeSelection.end, `\n${indent};; => ${output}\n`), wsEdit = new vscode.WorkspaceEdit();
     wsEdit.set(editor.document.uri, [edit]);
-    await vscode.workspace.applyEdit(wsEdit).then(_v => {
-        editor.selection = selection;
-    });
+    await vscode.workspace.applyEdit(wsEdit);
+    editor.selection = selection;
 }
 
 async function evaluateCode(code: string, options, selection?: vscode.Selection): Promise<void> {
--- a/src/evaluate.ts
+++ b/src/evaluate.ts
@@ -39,10 +39,10 @@ function interruptAllEvaluations() {
     vscode.window.showInformationMessage("Not connected to a REPL server");
 }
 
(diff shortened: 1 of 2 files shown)


src/evaluate.ts
```diff
@@ -39,10 +39,10 @@ function interruptAllEvaluations() {
     vscode.window.showInformationMessage("Not connected to a REPL server");
 }
 
-function addAsComment(c: number, result: string, codeSelection: vscode.Selection, editor: vscode.TextEditor, selection: vscode.Selection) {
+async function addAsComment(c: number, result: string, codeSelection: vscode.Selection, editor: vscode.TextEditor, selection: vscode.Selection) {
     const indent = `${' '.repeat(c)}`, output = result.replace(/\n\r?$/, "").split(/\n\r?/).join(`\n${indent};;    `), edit = vscode.TextEdit.insert(codeSelection.end, `\n${indent};; => ${output}\n`), wsEdit = new vscode.WorkspaceEdit();
     wsEdit.set(editor.document.uri, [edit]);
-    vscode.workspace.applyEdit(wsEdit).then((_v) => {
+    await vscode.workspace.applyEdit(wsEdit).then(_v => {
```
bpringe: Since you're using `await` here, you could get rid of the `.then` and just add the inner code below the `await`. You probably know that though. It's just a consistency thing in my head. :smile:

PEZ: This is some old code, when I understood even less of promises and await/async than a do today. 😄 I'll fix!

PR #1187, opened 2021-05-26
Add command for switching between test and implementation files

### LINKED ISSUE #1168: Feature request: Jump back and forth between Clojure source and its test
Dear Calva developers,

[As discussed on Clojureverse](https://clojureverse.org/t/we-created-and-maintain-calva-ask-us-anything/7131/18?u=rudolfvesely), I'd like to kindly ask to add the following feature:

* A command called "Clojure source/test jump" with or without specified key binding.
* When executed for example from a file `src/a/b/c.clj` Calva will open a file `test/a/b/c_test.clj` and then when executed again Calva will focus to the the source file `src/a/b/c.clj`.
* If the test file is missing, Calva will ask to create it with the following content:
``` clojure
(ns a.b.c-test
  (:require [clojure.test :refer :all]
            [a.b.c :as t]))
```
* If the source file is missing, Calva will ask to create it with the following content:
``` clojure
(ns a.b.c)
```
* Files are created with all parent directories in case they're missing

There is already a similar extension for Elixir called [Elixir Test](https://marketplace.visualstudio.com/items?itemName=samuel-pordeus.elixir-test) that has exactly this functionality so you may consider to take a look at its code (MIT licensed) to see how it's implemented.

Thank you.

Kind regards,

Rudolf Vesely

pratik97:
## What has Changed?

- Adds a command between toggling implementation and tests.

Fixes #1168 

## My Calva PR Checklist

I have:

Ping @pez, @bpringe

### DIFF
--- a/src/file-switcher/file-switcher.ts
+++ b/src/file-switcher/file-switcher.ts
@@ -1,5 +1,6 @@
 import * as vscode from 'vscode';
 import * as path from 'path';
+import * as util from './util';
 
 function openFile(file) {
     return vscode.workspace
@@ -39,41 +40,10 @@ function askToCreateANewFile(dir, file) {
     });
 }
 
-function isFileValid(openedFilename) {
-    return openedFilename.includes('src')
-    || openedFilename.includes('test')
-    || openedFilename.includes('main');
-}
-
-function getNewFilename(fileName, extension) {
-    if (fileName.includes('_test')) {
-        const strippedFileName = fileName.replace('_test', '');
-        return `${strippedFileName}${extension}`;
-    }
-    return `${fileName}_test${extension}`;
-}
-
-function getNewSourcePath(sourcePath) {
-    let replacedSourcePath = '';
-    const srcMainPath = path.join('src', 'main');
-    const srcTestPath = path.join('src', 'test');
-
-    if (sourcePath.includes(srcMainPath)) {
-        replacedSourcePath = sourcePath.replace(srcMainPath, srcTestPath);
-    } else if (sourcePath.includes(srcTestPath)) {
-        replacedSourcePath = sourcePath.replace(srcTestPath, srcMainPath);
-    } else if (sourcePath.includes('src')) {
-        replacedSourcePath = sourcePath.replace('src', 'test');
-    } else if (sourcePath.includes('test')) {
-        replacedSourcePath = sourcePath.replace('test', 'src');
-    }
-    return path.dirname(replacedSourcePath);
-}
-
 export async function toggleBetweenImplAndTest() {
     const activeFile = vscode.window.activeTextEditor;
     const openedFilename = activeFile.document.fileName;
-    const valid = isFileValid(openedFilename);
+    const valid = util.isFileValid(openedFilename);
 
     if (!valid) {
         return;
@@ -84,8 +54,8 @@ export async function toggleBetweenImplAndTest() {
     const fileName = fullFileName.split('.')[0];
     const extension = '.' + fullFileName.split('.')[1];
 
-    const sourcePath = getNewSourcePath(openedFilename);
-    const newFilename = getNewFilename(fileName, extension);
+    const sourcePath = util.getNewSourcePath(openedFilename);
+    const newFilename = util.getNewFilename(fileName, extension);
 
     const filePath = path.join(sourcePath, newFilename);
     const fileToOpen = vscode.workspace.asRelativePath(filePath);
--- a/src/file-switcher/file-switcher.ts
+++ b/src/file-switcher/file-switcher.ts
@@ -1,4 +1,5 @@
 import * as vscode from 'vscode';
+import * as path from 'path';
 
 function isWindows(openedFilename) {
     return openedFilename.includes('\\');
@@ -26,7 +27,7 @@ async function createNewFile(dir, file) {
 
 function askToCreateANewFile(dir, file) {
     function srcOrTest() {
-        if(file.includes('_test'))
+        if (file.includes('_test'))
             return 'test';
         return 'src';
     }
@@ -44,9 +45,9 @@ function askToCreateANewFile(dir, file) {
 
 function isCodeFile(openedFilename) {
     if (isWindows(openedFilename)) {
-        return openedFilename.match(/(.*\\)(test|src)(.*\\)(.*)(\.\w+)$/);
+        return openedFilename.match(/(.*\\)(test|src|main)(.*\\)(.*)(\.\w+)$/);
     }
-    return openedFilename.match(/(.*\/)(test|src)(.*\/)(.*)(\.\w+)$/);
+    return openedFilename.match(/(.*\/)(test|src|main)(.*\/)(.*)(\.\w+)$/);
 }
 
 export async function toggleBetweenImplAndTest() {
@@ -62,23 +63,30 @@ export async function toggleBetweenImplAndTest() {
     const testOrSrc = openedFile[2];
     const postDir = openedFile[3];
     const fileName = openedFile[4];
-    const replacedSrcOrTest = testOrSrc === 'src' ? 'test' : 'src';
+    const extension = openedFile[5];
+    const isMavenStyle = (openedFilename.includes(path.join('src', 'main'))
+        || openedFilename.includes(path.join('src', 'test'))) ? true : false;
+
+    let replacedFolderName = (testOrSrc === 'src' || testOrSrc === 'main') ? 'test' : 'src';
+    if (isMavenStyle && replacedFolderName === 'src') {
+        replacedFolderName = 'main';
+    }
 
     let newFilename = '';
 
     if (fileName.includes('_test')) {
         const strippedFileName = fileName.replace('_test', '');
-        newFilename = `${strippedFileName}.clj`;
+        newFilename = `${strippedFileName}${extension}`;
     } else {
-        newFilename = `${fileName}_test.clj`;
+        newFilename = `${fileName}_test${extension}`;
     }
-    const fileToOpen = vscode.workspace.asRelativePath(
-        startDir + replacedSrcOrTest + postDir + newFilename,
-    );
+
+    const filePath = path.join(startDir, replacedFolderName, postDir, newFilename);
+    const fileToOpen = vscode.workspace.asRelativePath(filePath);
 
     vscode.workspace.findFiles(fileToOpen, '**/.calva/**').then((files) => {
         if (!files.length) {
-            askToCreateANewFile(startDir + replacedSrcOrTest + postDir, newFilename);
+            askToCreateANewFile(startDir + replacedFolderName + postDir, newFilename);
         } else {
             const file = files[0].fsPath;
             openFile(file);
--- a/src/file-switcher/file-switcher.ts
+++ b/src/file-switcher/file-switcher.ts
@@ -1,0 +1,87 @@
+import * as vscode from 'vscode';
+
+function isWindows(openedFilename) {
+    return openedFilename.includes('\\');
+}
+
+function openFile(file) {
+    return vscode.workspace
+        .openTextDocument(vscode.Uri.file(file))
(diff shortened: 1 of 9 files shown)


pratik97:
I have tested below listed scenarios on MacOS 10.15.7 with the VSIX build.

- Having a test and its implementation files, running this command(via keybinding or not) from source file should switch the active file to test and vice versa.
- Having only a source file foo.clj, and running this command should open a dialogue within VSCode that asks for creating a new test file, if suggested it should create a new test file. If the respective test directory is missing, it should also create the directory. For example, if source file is `src/domain/foo.clj` and there is no test directory, running this command should create a file `test/domain/foo_test.clj`.
- Having only a test file foo_test.clj, running this command should similarly create a source file.

PEZ:
This is looking really good. Will test tonight and check the code changes a bit closer (at a glance it looks super duper fine).

Please also consider how we would document this feature. Maybe https://calva.io/test-runner/ is a the best current page to place the info. There is a VS Code task configured for building the docs and running the docs site locally. There's a section about this here: https://github.com/BetterThanTomorrow/calva/wiki/How-to-Hack-on-Calva

PEZ:
I think it makes sense to create an `ns` form automatically in the newly created file. I'm not really suggesting that it needs to be done in this PR, more noting it down.

src/file-switcher/file-switcher.ts
```diff
@@ -0,0 +1,95 @@
...
+    let newFilename = '';
+
+    if (fileName.includes('_test')) {
+        const strippedFileName = fileName.replace('_test', '');
+        newFilename = `${strippedFileName}${extension}`;
+    } else {
+        newFilename = `${fileName}_test${extension}`;
+    }
+
+    const filePath = path.join(startDir, replacedFolderName, postDir, newFilename);
```
PEZ: I think this can be extracted to one or more unit-testable functions. To make something unit-testable the code needs to be in a module that does not import `vscode`.

The codebase does not have very high code test coverage, but lately we have been better at that. For this particular feature, I predict that the logic around choosing the right paths and file names might get more complicated/involved than it is now. Good to start it off with testing, I think. Let me know if you want my help to wire the tests up.

pratik97:
@PEZ I think it's happening right now, does it not?

src/file-switcher/file-switcher.ts
pratik97: Hey @PEZ I am working on @seancorfield's suggestions, and it's cleaner and working fine except for one case, this code is slightly old now. I have posted a question on calva channel, let me know if you have any thoughts on that. I have not pushed that code, because it still requires refactoring(also it doesn't work with the case I described on the channel) and I'll try to refactor it with more unit-testable functions and once it gets reviewed, I will try to write unit-tests(Too unfamiliar with TypeScript and calva ecosystem to do TDD :/).

PEZ:
>  I think it's happening right now, does it not?

Oh, so it could be that `clojure-lsp` is having troubles starting in my dev environment. It doesn't like that I restart things so often, I think. 😄

PEZ:
Awesome with the unit tests!

seancorfield:
Thanks, @pratik97 -- the tests make it much easier to verify the intended behavior and now they can be expanded as needed once this is out in the wild and folks start asking for additional structures to be supported!

src/file-switcher/file-switcher.ts
```diff
@@ -0,0 +1,71 @@
...
+    const openedFilename = activeFile.document.fileName;
+    const valid = util.isFileValid(openedFilename);
+
+    if (!valid) {
+        return;
+    }
+
+    const fullFileName = openedFilename.split(path.sep).slice(-1)[0];
+    // Assumption: File names can't contain dots.
+    const fileName = fullFileName.split('.')[0];
```
bpringe: Though this may not be common, I believe file names can have dots, so that should be accounted for here. (I can create a `one.two.three.clj` file and use it as usual with a REPL, for example.)

pratik97:
@PEZ don't have a clue on how to fix this test-grammar failure, any inputs?

PEZ:
It has nothing to do with your PR. I'll try to figure it out.

bpringe:
I've fixed the CI job that was failing. If you merge dev and push, it should pass.

bpringe (changes requested):

src/file-switcher/file-switcher.ts
pratik97: Added the fix for this as well as extension-less files(which should be no-op).

bpringe: Awesome. Thanks!

bpringe (approved):

bpringe:
@PEZ, I haven't tested this PR myself, but if it looks good to you then I think we can merge.

pratik97:
@PEZ @bpringe only things that are remaining here is fixating the command keybindings(+ docs update) and testing on windows. Could you help me with that?

bpringe:
> fixating the command keybindings(+ docs update)

Sorry, could you explain what you mean here? I see the keyboard shortcut and a mention of it in the docs. What else should be done?

@PEZ should be able to help with the Windows testing since I believe he now has a Windows machine at hand.

pratik97:
@bpringe me and @PEZ discussed about what keybindings default should be better, currently it's `ctrl+option+s t` but can we change it to something other IDE uses or something that fits with other current/future Calva keybindings. We also, need to update them in the docs as well as we need to make it explicit in the docs that for the projects which doesn't have leiningen style folder structure, users may have to declare sourcepaths in `.lsp/config.edn` for the file switching to work.

PEZ:
There is something off with how paths for created files are determined. Especially when having a test file and creating an implementation. This is from a project created with the Luminus Template:

File path:

```
/Users/pez/Projects/tests/luminus-tests/a-luminus-shadow-2/test/clj/a_luminus_shadow_2/foo_test.clj
```

Resulting path starts from the parent of the project root and adding `srcs`:

image

Same thing happens on windows. Here with a minimal clj-new created project:

```
C:\Users\Peter Strömberg\win-proj\tests\clj-new-app\test\pez\foo_test.clj
```

image


WIth the Luminus project I get something similar when the file is not in `src`. Here's a file from the `env` source path:

```
/Users/pez/Projects/tests/luminus-tests/a-luminus-shadow-2/env/dev/clj/a_luminus_shadow_2/dev_middleware.clj
```

image


I don't know what is the right path for the test file here (with the `env` case). Maybe @seancorfield has an idea?

pratik97:
@PEZ there are 2 problems here:
1. The full path here, has a "tests" keyword in it, and we are blindly replacing any "test" with "src" keyword. The quick fix would be to check for "/test/" and replace it with "/src/" and vice versa. That would fix one problem, but again if you have a "src" or "test" folder in which your project resides then it would be a problem. Of the top of my head, I can't think of any idea where we can easily fix the second case, but maybe we could add our assumption in the docs? pushing the fix for the first case.
2. The luminous project structure has hierarchy in src paths(src/leiningen/new/a.clj) and on test path every test module is placed under test folder(test/a_test.clj) as far as I understand. This will be hard to capture with our current implementation where we just replace source with test. I don't have much idea about such project structures but if there is any way we can find that it's a luminous project structure, happy to implement that.

PEZ:
I can confirm that the bug with creating files in `<project-root>/../srcs` is gone with the latest build.

Now when using the command from the `env/dev/clj` in the Luminus project it offers to create the test file in `.`.  Then it doesn't create any file. I don't know what to do with this case really, but this feels a bit broken. Maybe if we don't recognize the structure we flash en error message saying this, instead of offering to create the file somewhere arbitrary? I don't know if it is easy or hard to detect this, though.

I think that instead of saying:

> Create the _test_ file in `<full-directory-path>`?

or

> Create the _src_ file in` <full-directory-path>`?

We could use just one message:

> Create `<project-relative-file-path>`?

That the file should be in the project is a given, so that is not necessary to show in the dialog, and it is also easier to deicide if I want a particular file created than figuring this out from the info about _where_ it would be created.

pratik97:
Figured couple of things we need to do by talking with @PEZ,
- If we want to support luminus style project structures, we can ask its creator about where to create test files for the special case of `env` folder. In general, we can also show error messages to users on running the commands for extension-less files or files whose path doesn’t have `src` or `test` keyword.
- We need to get the root project path and only replace the directories inside that.

seancorfield:
I seem to recall talking with @yogthos about the Luminus structure around tests but don't remember the details -- but I think that it is changing, along with the new edition of the book?

yogthos:
I recall updating tests to follow the latest convention, but that was a little while back and I haven't changed anything recently.

PEZ:
@yogthos the test structure `src/test` seems pretty standard. This question was more about source files outside the `src` directory, such as `env`.

yogthos:
The idea behind `env` source path is mainly to manage any environment specific configuration such as dev specific middleware. Any business logic in the app that warrants testing should live in the `src` source path. The tests will include `env/dev` source path by default, and the users can create a separate test profile for `env/prod` if that's warranted. The tests should live in the `test` source path though.

PEZ:
I've tested the last VSIX some. Will test some more edge cases and also on Windows again. I think we need to test it in the Remote setup as well, though i am not sure how to do that.

rjerue:
Testing on WSL on something like `lein new re-frame rftest2 +garden +test +routes`.

Happy Path works!

When one has multiple folders in their vscode workspace, I get a prompt to make the file even if it does exist. It does not make a new file, just takes me to the already made one.
image

At work, we have some projects that have `src-something` and... `test-something` When I change the folders like this and try it, it puts a core or core_test file in the root folder.
image
image

I'd be nice too if it maybe told me the exact file path it was making something.

We also have repos that don't have any sort of `_test` suffix. In these, the file names in `src` and `test` mirror each other in name. When I tried this case on `test/rftest2/core.clj` it made `src/rftest2/core_test.clj`, which I found a bit odd. This is despite it implying that it would be made under `test`
image
image

It may be useful to add more documentation to the rules surrounding this with the rules of the feature.

pratik97:
Thanks @rjerue for the detailed feedback. 
The issue with multiple projects open in a workspace, I am planning to fix that.
About the folder structure with `src-something` and `test-something` and the test files without prefix `_test` currently we don't have a support but can discuss with @PEZ if we are planning to support it. Supporting the `src-something` feature would mean that if someone have a `src-something` folder it will _always_ be replaced with `test-something` folder, which might not always be a desired outcome depending on the project (For example, `<project-root>/src/src-new/thisfeature.clj` and `/src/src-old/thisfeature.clj`).
Totally agree that we can add docs on the rules of this feature.

bpringe:
> if it maybe told me the exact file path it was making something

I think this is a good idea, to show the entire file path. Instead of saying `Create the src file at <path-without-file-name>?` it could say `Create <path-with-file-name>?`. This would be more clear, I think, and remove the need for the prompt to say things like "src" and "test" conditionally.

PEZ:
I now tested this in my work project, which has a non-standard structure like so:

```
- src
- test/src/foo/test
```

source-paths in the project are setup as `src` and `test/src`. The test file for something like `src/foo/bar.cljs` (namespace `foo.bar` are then placed in `test/src/foo/test/bar.cljs` (namespace `foo.test.bar`).

For this example, jumping from an implementation file to its test file offers to create a new file `test/bar-test.cljs`.

Jumping from the real test file to the implementation offers to create `test/test/foo/test/bar-test.cljs` (some of the extra weirdness is presumably because we think it is an implementation file).

This is probably quite hard to try fix and we do write in the documentation that non-standard structures are not supported. Is there a way we can check the structure and then just present an error message if we don't find a structure we support?

I am thinking that we should get this out to users so that we can get feedback from the field. We should prepare for merging. With these changes:

1. We don't provide a default keyboard shortcut. Then it will be more of an active choice to test the feature.
2. That structure check (if you  find it feasible, @pratik97, otherwise we can skip that, I think)

PEZ:
Thanks! And extra thanks for the persistence with this tricky issue.

PR #1200, opened 2021-06-14
Add nrepl and clojure-lsp versions to Calva says greetings

### LINKED ISSUE #1199: Show jack-in dependencies and clojure-lsp version at startup
When asking questions and filing issues users are often asked about what versions of nrepl and clojure-lsp they are using. If we print this to the greetings from Calva says, we can always tell people to look there for the answer.

PEZ:
## What has Changed?

We show configured jack in dependencies, and clojure-lsp versions in Calva says greetings message.

Fixes #1199

## My Calva PR Checklist

I have:

Ping @bpringe


bpringe:
Using the configuration to get the clojure-lsp version can be misleading if the user set a binary to be used with a custom path. https://calva.io/clojure-lsp/#using-a-custom-clojure-lsp-native-binary

The server info command is really the one source of truth for the clojure-lsp version. I do see you're saying "this is the version configured," so that would sort of be true, but not necessarily the version being used.

The same goes for nrepl. Depending on their classpath and tooling used (lein vs tools.deps), I think a different version of nrepl could be used than the one configured in jack-in deps. We'd also need to be aware of whether they connected to a repl without using jack-in, because in that case the nrepl version could be different too.

PR #1206, opened 2021-06-16
Add batching to output window printing

### LINKED ISSUE #942: REPL is Slow and Performance Degrades as the Output Grows
Evaluating this takes over 30 seconds to finish all output to the REPL:
(doseq [n (range 1000)]
     (println n))
     
And performance degrades gradually as the output to .calva/output-window/output.calva-repl grows. :/

brdloush:
## What has Changed?

When `editQueue` is filled rapidly by many one-line `println` outputs, `append` function in `results-doc.ts` actually processed that queue one-by-one, always appending a single line to document (`output.calva-repl`) and scrolling to the bottom of the document. This was very slow. (even 25 seconds for 500 outputted lines)

This change basically does what klauswuestefeld suggested here:

https://github.com/BetterThanTomorrow/calva/issues/942#issuecomment-759049579

> If there is a queue, can't we read all pending output events from the
> queue, concat them and apply them as a single batch to the output file?
> That would maybe make the console jumpy but at least it would not be
> absurdly slow.

Even after the change it's likely that first few executions might end up appending just a few lines, as the queue is not filled much. This could definitely be improved somehow (probably by some waiting a few hundred ms), but it would introduce much more complexity. And it's likely that this whole output processing/redirection to `output.calva-repl` might deserve a bigger refactoring/rewrite.

I tried to make this change as little and as simple as possible.

Fixes #942 (I'd say drastically improves, not completely fixes. After 30k of lines in `output.calva-repl` it takes let's say 5 seconds to append 500 lines. It's not super quick, but it's bearable. Without the change, it's not bearable, as it takes minutes.)

## My Calva PR Checklist

I have:

- [x] ~Read [How to Contribute](https://github.com/BetterThanTomorrow/calva/wiki/How-to-Contribute#before-sending-pull-requests).~
- [x] ~Directed this pull request at the `dev` branch. (Or have specific reasons to target some other branch.)~
- [x] ~Made sure I have changed the PR base branch, so that it is not `published`. (Sorry for the nagging.)~
- [x] ~Updated the `[Unreleased]` entry in `CHANGELOG.md`, linking the issue(s) that the PR is addressing.~
- [x] ~Figured if **anything** about the fix warrants tests on Mac/Linux/Windows/Remote/Whatever, and either tested it there if so, or mentioned it in the PR.~
- [x] ~Added to or updated docs in this branch, if appropriate~
- [x] ~Tested the VSIX built from the PR (so, after you've submitted the PR). You'll find the artifacts by clicking _Show all checks_ in the CI section of the PR page, and then _Details_ on the `ci/circleci: build` test. NB: *You need to sign in/up at Circle CI to find the _Artifacts_ tab.*~
     - [x] ~Tested the particular change~
     - [x] ~Figured if the change might have some side effects and tested those as well.~
     - [x] ~Smoke tested the extension as such.~
- [x] ~Referenced the issue I am fixing/addressing _in a commit message for the pull request_.~
     - [x] ~If I am fixing the issue, I have used [GitHub's fixes/closes syntax](https://help.github.com/en/articles/closing-issues-using-keywords)~
     - [x] ~If I am fixing just part of the issue, I have just referenced it w/o any of the "fixes” keywords.~
- [x] ~Created the issue I am fixing/addressing, if it was not present.~

Ping @pez, @bpringe

### DIFF
--- a/src/results-output/results-doc.ts
+++ b/src/results-output/results-doc.ts
@@ -296,28 +296,18 @@ export function append(text: string, onAppended?: OnAppendedCallback): void {
                             new vscode.Location(DOC_URI(), doc.positionAt(Infinity)));
                     }
                     
-                    const queueLength = editQueue.length
-                    if (queueLength > 0) {                        
-                        var numOfQueueItemsWithoutCallback: number = 0;
-                        for (var i=0; i<queueLength; i++) {
-                            var [_, onAppendCallback] = editQueue[i];
-                            if (onAppendCallback != null) {
-                                numOfQueueItemsWithoutCallback = i;
-                                break;
-                            }
-                        }
-                        if (numOfQueueItemsWithoutCallback > 0) {
-                            const maxTextsToGroup = 1000;
-                            var textGroup = [];
-                            for (var i=0; i<Math.min(queueLength, maxTextsToGroup); i++) {
-                                var [text] = editQueue.shift();
-                                textGroup.push(text);
-                            }
-                            return append(textGroup.join('\n'), null)
+                    if (editQueue.length > 0) {
+                        const batchSize = 1000;
+                        const textBatch = takeWhile(editQueue, (value, index) => {
+                            return index <= batchSize && !value[1];
+                        }).map(value => value[0]);
+                        if (textBatch.length > 0) {
+                            editQueue.splice(0, textBatch.length);
+                            return append(textBatch.join('\n'));
                         } else {
-                            var [text, onAppendCallback] = editQueue.shift();
-                            return append(text, onAppendCallback)
+                            return append.apply(null, editQueue.shift());
                         }
+                    }
                     }
                 });
             }
--- a/src/results-output/results-doc.ts
+++ b/src/results-output/results-doc.ts
@@ -13,6 +13,7 @@ import * as replHistory from './repl-history';
 import * as docMirror from '../doc-mirror/index'
 import { PrintStackTraceCodelensProvider } from '../providers/codelense';
 import * as replSession from '../nrepl/repl-session';
+import { max } from 'lodash';
 
 const RESULTS_DOC_NAME = `output.${config.REPL_FILE_EXT}`;
 
@@ -296,15 +297,27 @@ export function append(text: string, onAppended?: OnAppendedCallback): void {
                     }
                     
                     const queueLength = editQueue.length
-                    if (queueLength > 0) {                    
-                        var allTexts = [];
-                        var lastOnAppendCallback: OnAppendedCallback;
-                        for (var i=0; i<queueLength; i++) {                            
-                            var [text, onAppendCallback] = editQueue.shift();                            
-                            allTexts.push(text);                            
-                            lastOnAppendCallback = onAppendCallback;
-                        }                                                 
-                        return append(allTexts.join('\n'), lastOnAppendCallback)
+                    if (queueLength > 0) {                        
+                        var numOfQueueItemsWithoutCallback: number = 0;
+                        for (var i=0; i<queueLength; i++) {
+                            var [_, onAppendCallback] = editQueue[i];
+                            if (onAppendCallback != null) {
+                                numOfQueueItemsWithoutCallback = i;
+                                break;
+                            }
+                        }
+                        if (numOfQueueItemsWithoutCallback > 0) {
+                            const maxTextsToGroup = 1000;
+                            var textGroup = [];
+                            for (var i=0; i<Math.min(queueLength, maxTextsToGroup); i++) {
+                                var [text] = editQueue.shift();
+                                textGroup.push(text);
+                            }
+                            return append(textGroup.join('\n'), null)
+                        } else {
+                            var [text, onAppendCallback] = editQueue.shift();
+                            return append(text, onAppendCallback)
+                        }
                     }
                 });
             }
--- a/src/results-output/results-doc.ts
+++ b/src/results-output/results-doc.ts
@@ -294,8 +294,17 @@ export function append(text: string, onAppended?: OnAppendedCallback): void {
                         onAppended(new vscode.Location(DOC_URI(), insertPosition),
                             new vscode.Location(DOC_URI(), doc.positionAt(Infinity)));
                     }
-                    if (editQueue.length > 0) {
-                        return append.apply(null, editQueue.shift());
+                    
+                    const queueLength = editQueue.length
+                    if (queueLength > 0) {                    
+                        var allTexts = [];
+                        var lastOnAppendCallback: OnAppendedCallback;
+                        for (var i=0; i<queueLength; i++) {                            
+                            var [text, onAppendCallback] = editQueue.shift();                            
+                            allTexts.push(text);                            
+                            lastOnAppendCallback = onAppendCallback;
+                        }                                                 
+                        return append(allTexts.join('\n'), lastOnAppendCallback)
                     }
                 });
             }
--- a/src/results-output/results-doc.ts
+++ b/src/results-output/results-doc.ts
@@ -13,7 +13,6 @@ import * as replHistory from './repl-history';
 import * as docMirror from '../doc-mirror/index'
 import { PrintStackTraceCodelensProvider } from '../providers/codelense';
 import * as replSession from '../nrepl/repl-session';
-import { max, takeWhile } from 'lodash';
 import { splitEditQueueForTextBatching } from './util';
 
 const RESULTS_DOC_NAME = `output.${config.REPL_FILE_EXT}`;
--- a/src/results-output/results-doc.ts
+++ b/src/results-output/results-doc.ts
@@ -13,7 +13,7 @@ import * as replHistory from './repl-history';
 import * as docMirror from '../doc-mirror/index'
 import { PrintStackTraceCodelensProvider } from '../providers/codelense';
 import * as replSession from '../nrepl/repl-session';
-import { max } from 'lodash';
+import { max, takeWhile } from 'lodash';
 
 const RESULTS_DOC_NAME = `output.${config.REPL_FILE_EXT}`;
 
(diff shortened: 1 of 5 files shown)


src/results-output/results-doc.ts
```diff
@@ -294,8 +294,17 @@ export function append(text: string, onAppended?: OnAppendedCallback): void {
...
+                    const queueLength = editQueue.length
+                    if (queueLength > 0) {                    
+                        var allTexts = [];
+                        var lastOnAppendCallback: OnAppendedCallback;
+                        for (var i=0; i<queueLength; i++) {                            
+                            var [text, onAppendCallback] = editQueue.shift();                            
+                            allTexts.push(text);                            
+                            lastOnAppendCallback = onAppendCallback;
+                        }                                                 
+                        return append(allTexts.join('\n'), lastOnAppendCallback)
```
bpringe: It seems with this code only the last `onAppendCallback` is run for a given batch, and the others are discarded. Each call to `append` can come from a different caller / location, and we need to make sure every callback is executed.

bpringe (changes requested):
Thanks for the PR. Besides the comment I made about callbacks, do we know that this looping mechanic won't continue indefinitely if a process is printing continuously and rapidly, resulting in nothing being printed?

bpringe:
I'd like to discuss this with @PEZ too before we decide on what to do here. In the linked issue we discussed adding an option to send stdout to an output channel. I still think that should be done, however, I'm unsure what, if anything, we should change about appending to the output window, since I think the output channel would solve the speed issue for those that need it.

If we did add batching to the output window appending, I think there does need to be a time period cutoff at which collected outputs will be printed no matter what's left in the queue.

If I run:

```clojure
(doseq [i (range 30000)]
    (println i))
```

it will print some, then a little more, then maybe pause until it spits out the rest up to 30000. While this might be okay for fixed size printings, what about for streaming output? Say I'm connected to some continuous stream of data like from a Kafka topic with a high volume of inputs, and I'm printing them as they're received. Obviously this would freeze up with the current setup too, but it also would with this batching setup, even more so if we don't have a delay to flush collected outputs.

Having a flush delay then adds complexity, as you mentioned. What should the delay be? Probably not more than 100 or 200 ms, but then:
* For single or a few output lines, appending would take the length of time as the delay, even if it didn't need to, OR
* We somehow make it print without delay if only a certain number of outputs are in the queue. I'm not sure how well this could be done or would work.

The above points lead me back to wondering if we should keep the implementation as it is, and only add the option to send stdout to an output channel. This seems to be the simplest of solutions. I'm open to discussion on this though, of course.

PEZ:
I'll have to read this when I am a bit clearer in my mind than I am currently. But my spontaneous sense for it is that just printing everything in the queue at each time there is any printing consider should be better than just printing one line at a time.

Disclaimer. I couldn't really grasp this mechanism when I started to implement it, and it was saved by @bpringe who could grasp it. So my gut feeling might not be worth much at all. I'll read this first thing tomorrow after I have had my morning coffee.

bpringe:
It's hard for me to tell the possible impact of calling the callbacks after a batch of prints has been combined and printed at once, but I'm thinking we'll rarely if ever run into an issue with that. Most of the parts of code that use the callback are related to printing results, and there aren't that many places, though the callback is important to keep for some of those usages.

stdout printing will not be using the callback, so the only time I can see callback usage possibly being a problem is if a result is printed in the middle of stdout printings, or something else that uses a callback is called in the middle of stdout printing.

Anyway, I think it could be worth trying this change, at least for a while. I do think we need to make sure we call every callback in the queue though, not just the last one.

bpringe:
Here's code that could work for combining all callbacks into another callback that would, when called, execute all callbacks for the batch in order, assuming `callbacks` is an array containing the collected callbacks, similar to the one that contains all text edits:

```javascript
const callback: OnAppendedCallback = callbacks.reduce((cb, f) => {
    return (il, np) => {
        cb(il, np);
        f(il, np);
    };
}, (il, np) => {});
```

There may be a more elegant way to do that in JS. :man_shrugging:

bpringe:
Another thing to be aware of is that in a couple of places we use the `insertPosition` and `newPosition` of the callback after appending something. Those will now correspond to the end of the batch, not to the end of the item passed in the original `append` call. I don't really know if this will cause any issues, though.

PEZ:
That with us keeping track of the the positions is probably worth exploring a bit. We rely on it a bit less today, than we used to, but still...

bpringe:
Just to add some info, places where we use the arguments to the callback are:

* https://github.com/BetterThanTomorrow/calva/blob/35f1490fe0a8e43b3c19ecab37a83edf2a1973d3/src/evaluate.ts#L79 - I think this was used when we were storing locations of results in the output window to use with the peek feature in the editor. Since we've added the results back to hovers, these locations aren't needed any more, and don't look like they're still being used.
* https://github.com/BetterThanTomorrow/calva/blob/35f1490fe0a8e43b3c19ecab37a83edf2a1973d3/src/evaluate.ts#L106 - for managing stack trace code lenses in the output window after an error occurs
*  https://github.com/BetterThanTomorrow/calva/blob/35f1490fe0a8e43b3c19ecab37a83edf2a1973d3/src/evaluate.ts#L115 - same as above

So, it seems we're only using the callback `Location` arguments for the stack trace code lenses.

brdloush:
Sorry, I wasn't around last few days. 

It seems to me that the problem is much more complex than I originally thought - I should have looked around first and see where those callbacks are being used and how.

If I understand correctly, every appending operation (to `output.calva-repl`) can actually notify original caller (which asked for append) via its callback. This callback is supposed to be called after the appending was applied and "committed" (saved), and the called should effectively receive a position "just before the text was applied" and "just after the text was applied". 

If I were to pass correct before/after positions for each individual appending when using this "groupping append optimization", I 'd probably have to somehow analyze the individual grouped texts, count their own number of lines etc. Which is probably doable, but might complicate things a bit and perhaps even itroduce bugs. Moreover what I'm a bit more worried about is that the callback function passed by caller might even try to do its own writing to `output.calva-repl`. In such case everything would probably break, as the line numbering of all those groupped appends would change.

If I were to keep things simple (which I'd definitely prefer), I'd probably try following approach: 

1) Peek into the queue and try to find a block of queued text appends which actually don't have any callbacks registered on them. Such block must reside at the the very beginning of the queue (idx 0).
2) if such "block of no-callback queued appends at the beginning of the queue" was found, we can simply batch them as one append as no calling of callback is needed. (perhaps we can also limit the max number of grouped operations if that makes sense).
3) otherwise we simply take first item from the queue, append it, let the callback to be called (ie. current behavior).

Not sure if this makes sense to you and whether I should change the PR accordingly. If you have other plans on more complex redesign/rewrite, I'll be more than happy to close this PR. I'm just hoping to see at least some improvement in appending speed as logging via multiple `println`s is currently quite slow.

bpringe:
I think that's a pretty great idea. I don't see anything wrong with it. I do think we should set a maximum number of items to a batch as you mentioned so printing doesn't just hang for a while when appends occur rapidly. I'm open to trying this out. 

I think calls to append with a callback happen infrequently while rapid printing is happening, but I guess it depends on the user's dev workflow and project. Anyway, the implementation you described sounds like it would handle that situation gracefully when it does happen.

brdloush:
I've implemented the discussed change and tried the built vsix. Seems to work fine, but definitely needs more testing.

bpringe:
Thanks for the changes. I'm away on a trip, but I'll test/review next week.

src/results-output/results-doc.ts
```diff
@@ -294,8 +295,29 @@ export function append(text: string, onAppended?: OnAppendedCallback): void {
...
+                            var textGroup = [];
+                            for (var i=0; i<Math.min(queueLength, maxTextsToGroup); i++) {
+                                var [text] = editQueue.shift();
+                                textGroup.push(text);
+                            }
+                            return append(textGroup.join('\n'), null)
+                        } else {
+                            var [text, onAppendCallback] = editQueue.shift();
+                            return append(text, onAppendCallback)
+                        }
```
bpringe: Nice work on this. Seems to work fine. I'd like to make it a little more functional, though. I added a suggestion below. If you like it feel free to merge it in. Also feel free to make a suggestion or alter it. This requires importing `takeWhile` from lodash.

If you go with this, I think a test for the logic that gets the `testBatch` (the `takeWhile` and the `map`) could use a unit test. I'll leave that to you, unless you'd like me to do it, which I don't mind.

```suggestion
                    if (editQueue.length > 0) {
                        const batchSize = 1000;
                        const textBatch = takeWhile(editQueue, (value, index) => {
                            return index <= batchSize && !value[1];
                        }).map(value => value[0]);
                        if (textBatch.length > 0) {
                            editQueue.splice(0, textBatch.length);
                            return append(textBatch.join('\n'));
                        } else {
                            return append.apply(null, editQueue.shift());
                        }
                    }
```

src/results-output/results-doc.ts
```diff
@@ -13,6 +13,7 @@ import * as replHistory from './repl-history';
 import * as docMirror from '../doc-mirror/index'
 import { PrintStackTraceCodelensProvider } from '../providers/codelense';
 import * as replSession from '../nrepl/repl-session';
+import { max } from 'lodash';
```
bpringe: This goes with my suggestion lower in the file.

```suggestion
import { max, takeWhile } from 'lodash';
```

bpringe (changes requested):

src/results-output/results-doc.ts
brdloush: Nice, it's much more concise now. I remember I was grepping for something like `takeWhile`, but didn't find it being used anywhere in the code, so sticked with that imperative code. Looks much better with `takeWhile` :+1: 

I'll try to find some time to add a few tests in the evening or tomorrow.

brdloush:
I've extracted the queue-splitting/batching code into its own utility function (which no longer mutates the received queue instance) and added few tests for it. Not sure where it's the best place to put it, as I believe it needs to be exported in order to be accessible in tests (or tested indirectly trough some other public fn, but that would probably be a bit painful).

bpringe (approved):
Nice refactor and extraction and great tests!

PR #1220, opened 2021-06-30
Workaround for paredit's inability to support multiline

JBlaz:
## What has Changed?
- When more than one line is selected use VSCodes default behavior for sexp navigation or selection. This will need to be removed when support is added.

## My Calva PR Checklist

I have:


PEZ:
Thanks for the PR!

I think there are two different cases here:

1. There is some very similar built-in command default bound to the same keys as Paredit's defaults
    * I think this only applies to *Expand Selection* and *Shink Selection* and only on Windows and Linux (on Mac the *Paredit Grow Selection* command is bound to different keys than the built-in command.
2. There is no very similar built-in command and/or it is not default bound to the same keys
    * This applies to all other Paredit Commands that I can think of
    * The PR includes a few of these, but not others.

For category 2 the workaround would disable the command when there are multiple cursors, instead of what happens today that the command applies to the first of the cursors. It's a matter of taste which is preferable.

For category 1, the new behaviour would make more sense. Calva's commands for growing and shrinking the selection work very similar to how VS Code's built-in commands do. It wouldn't be a disablement. The problem is, these commands are category 2 on Mac. We might be able to solve that by splitting the shortcut definition into separate definitions for Mac and Windows/Linux. But that adds complexity and maintenance...

Maybe it does most sense to leave it up to the user to work around this limitation? The Calva documentation could include information about it. Here would be a good place, I think: https://calva.io/paredit/#about-the-keyboard-shortcuts

JBlaz:
Fair enough. Documentation changed for your consideration.

Also, grow and shrink are the same key bindings at least on linux.

PEZ:
This is awesome. Sorry for the radio silence. I have to admit I just dropped the ball. 😢

PEZ:
Many thanks!

PR #1237, opened 2021-07-17
improve performance of rangeForDefun (cause of repl slowness) #942

brdloush:
## What has Changed?
- `rangeForDefun` got rewritten so that it no longer needs to navigate from the very beginning of the document (which caused slowness when appending to repl window)
- added way more tests for `rangeForDefun` (as there weren't any :-))

Greatly improves #942. Before the fix, outputting become sluggish at 6k lines and was nearly unusable at just 10k (freezes for 30-60s). Now even 100k lines is bearable : execution of repl expression takes about 5 seconds.

## My Calva PR Checklist

I have:

     - ~[ ] If I am fixing the issue, I have used [GitHub's fixes/closes syntax](https://help.github.com/en/articles/closing-issues-using-keywords)~

Ping @pez, @bpringe

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Changes to Calva.
 
 ## [Unreleased]
+- Performance improvement [REPL is Slow and Performance Degrades as the Output Grows](https://github.com/BetterThanTomorrow/calva/issues/942) 
 
 ## [2.0.205] - 2021-07-14
 - [Use new custom LSP method for server info command and print info in "Calva says" output channel ](https://github.com/BetterThanTomorrow/calva/issues/1211)
--- a/src/cursor-doc/paredit.ts
+++ b/src/cursor-doc/paredit.ts
@@ -92,7 +92,7 @@ export function selectOpenList(doc: EditableDocument) {
  */
 export function rangeForDefun(doc: EditableDocument, offset: number = doc.selection.active): [number, number] {
     const cursor = doc.getTokenCursor(offset);
-    return cursor.rangeForDefun2(offset);
+    return cursor.rangeForDefun(offset);
 }
 
 export function forwardSexpRange(doc: EditableDocument, offset = Math.max(doc.selection.anchor, doc.selection.active), goPastWhitespace = false): [number, number] {
--- a/src/cursor-doc/paredit.ts
+++ b/src/cursor-doc/paredit.ts
@@ -90,9 +90,9 @@ export function selectOpenList(doc: EditableDocument) {
  * Gets the range for the ”current” top level form
  * @see ListTokenCursor.rangeForDefun
  */
-export function rangeForDefun(doc: EditableDocument, offset: number = doc.selectionLeft, start: number = 0): [number, number] {
-    const cursor = doc.getTokenCursor(start);
-    return cursor.rangeForDefun(offset);
+export function rangeForDefun(doc: EditableDocument, offset: number = doc.selection.active): [number, number] {
+    const cursor = doc.getTokenCursor(offset);
+    return cursor.rangeForDefun2(offset);
 }
 
 export function forwardSexpRange(doc: EditableDocument, offset = Math.max(doc.selection.anchor, doc.selection.active), goPastWhitespace = false): [number, number] {
--- a/src/cursor-doc/token-cursor.ts
+++ b/src/cursor-doc/token-cursor.ts
@@ -635,7 +635,7 @@ export class LispTokenCursor extends TokenCursor {
     }
 
     rangeForDefun2(offset: number, commentCreatesTopLevel = true): [number, number] {
-        const cursor = this.clone();
+        const cursor = this.doc.getTokenCursor(offset);
         let lastCandidateRange: [number, number] = cursor.rangeForCurrentForm(offset);
         while (cursor.forwardList() && cursor.upList()) {
             const commentCursor = cursor.clone();
--- a/src/cursor-doc/token-cursor.ts
+++ b/src/cursor-doc/token-cursor.ts
@@ -613,6 +613,7 @@ export class LispTokenCursor extends TokenCursor {
      */
     rangeForDefun(offset: number, depth = 0, commentCreatesTopLevel = true): [number, number] {
         const cursor = this.clone();
+
         while (cursor.forwardSexp()) {
             if (cursor.offsetEnd >= offset) {
                 if (depth < 1 && cursor.getPrevToken().raw === ')') {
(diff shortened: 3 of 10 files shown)


brdloush:
I've rebased this PR to current dev, so that it VSIX for testing has those 2.0.205 changes included.

PEZ:
Awesome. This performance bottleneck is causing Calva to feel quite broken when I run in to it.

Can you add some pointers to what people should test when they get hold of the VSIX to the PR description? From the top of my mind:

* [ ] The debugger (maybe @bpringe can tell us how to confirm that the change hasn't broken things here)
* [ ] Rainbow brackets
* [ ] Evaluating top level form
* [ ] Evaluating top level form to cursor
* [ ] Custom REPL command variables involving the top level form
* [ ] Evaluating at the REPL window prompt

I am probably missing things, am in a bit of a hurry...

bpringe:
For the debugger, testing that breakpoints are found and that stepping works correctly should suffice.

bpringe:
I've quickly tested all of the points above and the only thing that I see an issue with is when I add an extra/unbalanced parenthesis, it's not highlighted in red like it is currently in Calva.

Currently in Calva:
image

With the VSIX from this PR:
image

brdloush:
Nicely spot @bpringe, I never noticed that these unbalanced parentheses gets highlighted in red, so I wasn't really missing this behavior.

@PEZ I've debugged current `dev` vs `defun-range-experiments` and found out that we slightly changed the behavior in new `rangeForDefun`. Old `dev` version returned at least `[offset, offset]` range "when nothing better was found". Our current version returns undefined in such case. This breaks mentioned ranbow brackets behavior in `updateRainbowBrackets` as this throws "NPE" (I'm not sure they call it like that in JS :smile:) :

```js
      endRange = endCursor.rangeForDefun(endOffset, false),
      rangeStart = startRange[0],
      rangeEnd = endRange[1];      // <-- NPE here as endRange is null
```

A simple fix might be adding two additional `!== undefined` checks when working with `lastCandidateRange`:

```js
    rangeForDefun(offset: number, commentCreatesTopLevel = true): [number, number] {
        const cursor = this.doc.getTokenCursor(offset);
        let lastCandidateRange: [number, number] = cursor.rangeForCurrentForm(offset);
        while (cursor.forwardList() && cursor.upList()) {
            const commentCursor = cursor.clone();
            commentCursor.backwardDownList();
            if (!commentCreatesTopLevel || commentCursor.getToken().raw !== ')' || commentCursor.getFunctionName() !== 'comment') {
                const currentFormRange = cursor.rangeForCurrentForm(cursor.offsetStart);
                if (currentFormRange !== undefined) {
                    lastCandidateRange = currentFormRange;
                }
            }
        }
        return (lastCandidateRange !== undefined)
            ? lastCandidateRange
            : [offset, offset];
    }
```

This alone helps in situation where the unbalanced bracket is the last form in editor viewport that gets examined in `updateRainbowBrackets`. But if the are some more forms after the unbalanced parenthesis, highlighting still doesn't work correctly. 

I suspect that it might somehow be related to yet another changed behavior:

```cljs
(ns a
  (:require [clojure.string :as s]))

(defn foo []
  (let [bar 1]
    "bar")) )  ;; <- last paren on this line is unbalanced
   
(println (s/upper-case "test output"))
|
```

When I evaluate this in old `dev` version, it does nothing.. basically no form after the unbalanced paren can get evaluated. In new version, it evals the `println` form.

Unfortunately the whole logic in `updateRainbowBrackets` is way too heavy for my brain. Perhaps you might have some idea how to fix it (if it's really the root problem of this highlighting issue) :crossed_fingers:

PEZ:
Great find, @bpringe, and awesome investigation, @brdloush!

As for new evaluation behaviour with unbalanced parens, I don't worry about that. Unbalance is unbalance. 😀

I'm thinking that instead of fixing the call site with that highlight miss, is it hard to fix the defun method to keep the behaviour it used to have in respect to what is returned in this particular case? I'm far afk, so can't check myself right now.

PEZ:
OK. So rereading @brdloush 's comment with a desktop browser, I see there is a suggestion to keep the old `undefined` behaviour.

Testing this I notice that clj-kondo has a check for unbalance that is better than Calva's. Hmmm...

brdloush:
@PEZ I have one more observation. Consider this code

```clj
|(ns a
  (:require [clojure.string :as s])) ;; <-- forwardUpSexp normally jumps after this paren, but not with unbalanced paren..

(defn foo []
  (let [bar 1]
    "bar")) )  ;; <-- this last paren is ubalanced - forwardUpSexp actually ends up here  

:last-form 
```

If there's this unbalanced paren on "bar" line and I have a cursor at the very beginning (`|`), then even at `dev` version, doing a `ctrl+alt+down` (`forwardUpSexp`) actually makes the cursor jump just after the unbalanced closing paren, instead of jumping to closing paren of `ns` form.

I believe this is a problem for that `updateRainbowBrackets` code, as in mentioned case, `startPaintingFrom` is assigned an offset that's behind the actual problematic unbalanced paren. When I manually made a hack to that variable and reassigned it to 0 (which is what `startPaintingFrom` for that block of code is assigned in `dev` version), than the unbalanced paren gets the red background.

@PEZ what do you think about that `forwardUpSexp` behavior in such case? Is it broken even in current `dev`, or is such behavior a "non-defined feature"? Perhaps if we find a way to fix it, our new `rangeForDefun` will work fine, including the highlighting.

PEZ:
@brdloush I think it is worth a try to fix the case you are pointing at. That said, it remains hard to define the ”correct” behaviour in the face of unbalance. Let's not do it in this PR.

Thinking about the highlight problem I started to think it makes sense that when we don't find a top level form, we return `undefined`. So I pushed a change where the highlighter call site deals with that instead of assuming a zero-width range. This should fix the issue that you found, @bpringe, can you test again with the same files?

brdloush:
@PEZ your commit fixes this case

image

but as mentioned, not the one when there are more forms after the imbalance:

image

PEZ:
@brdloush is the case that is not fixed caused by this PR or something we already have in the released version?

brdloush:
> @brdloush is the case that is not fixed caused by this PR or something we already have in the released version?

The highlighting itself works ok in released version. 

As I wrote in my previous comment, I think that the fact that highlighting is broken in our new branch is related to the fact that we now rely on `forwardUpSexp`, which - even in released version - is "broken":

Even in released version, `forwardUpSexp` jumps to end of line 5, when you use `ctrl+alt+down` at the very beginning of the document: (offset 0). Ideally, the cursor should end up at offset 0, not jump to end of line 5. It stays at offset 0 if there's not that imbalanced paren on line 5.

image

PEZ:
Ah, I'm starting to get it. =)

PEZ:
Though I suspect it is `forwardList` that is the problem:

image

PEZ:
Pushed a fix for `forwardList` now. In my testing it fixes the problem.

brdloush:
:+1: It's behaving very well now including the highlighting :clap: I'll install the VSIX later today or tomorrow and try to re-test everything more thoroughly.

@bpringe can you perhaps re-test your scenarios as well?

PEZ (approved):

bpringe:
I've been pretty busy but I'll try to test this in the next couple of days, if it's still needed.

PEZ:
Thanks for helping with analysing this problem and persisting with analysing it down to the root causes, @brdloush! I think that we managed to navigate this such that the right fixes where applied and also the tests improved a lot. Let's deal with any regressions and room for improvements in future PRs.

brdloush:
Awesome, looking forward to the next release :clap:

PR #1292, opened 2021-09-05
Add nrepl message log

### LINKED ISSUE #1198: Add nrepl message logging
This has been discussed before, and I believe Cider has this. It could help us debug some issues, such as #1153, but probably several others.

I think an output channel would be good for this. What do you think @PEZ?

bpringe:
## What has Changed?

- Adds a command for toggling nrepl message logging on and off - "Toggle nREPL Message Logging Enabled"
- When nrepl logging is enabled, and output channel called "nREPL Messages" is created and shown. nREPL messages are logged to this channel, and you can see the direction in which the message travelled.
- The command is enabled at all times, whether or not the repl is connected. This allows logging of messages as soon as a repl is connected (if logging is enabled before connecting a repl).
- When nrepl logging is disabled, the channel is hidden and disposed.

Closes #1198 

## My Calva PR Checklist

I have:

Ping @pez, @bpringe


src/nrepl/index.ts
```diff
@@ -77,7 +78,8 @@ export class NReplClient {
      * @param data
      */
     write(data: any) {
-        this.encoder.write(data)
+        this.encoder.write(data);
+        log(data, Direction.ClientToServer);
```
PEZ: Not sure about this, but to me it looks like we are always logging. Should we export a function that does the check about logging enabled or not and use it here to make it clearer what's going on?

src/nrepl/logging.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (channel) {
+        const formattedMessage = `${direction}\n${formatNreplMessage(message)}\n`;
+        channel.appendLine(formattedMessage);
+    }
+}
+
+const enum Direction {
+    ClientToServer = 'Client -> Server',
+    ServerToClient = 'Server -> Client'
+}
```
PEZ: I'm thinking maybe use something that is a valid map key for this? Then you can select it all, wrap it in square brackets and REPL with it. Given that the rest is in some valid format. Looks a bit like JSON, right? Maybe consider making it EDN so it is extra ready for the REPL even. (Maybe that last one is for later, though.)

PEZ (commented):
Awesome to finally see this happening! Haven't tried it yet, hopefully I get to do that during the day.

src/nrepl/index.ts
bpringe: Sure. I wasn't sure which way it should be done here. I opted for adding the check for logging being enabled in one place, and then calling code can just say "log here" and not worry about whether it's enabled. However, I feel it's not explicit so I'll change it.

src/nrepl/logging.ts
bpringe: So, with JSON I'm not sure if we can make the output immediately programmatically consumable because these objects are output one by one, and would need a comma separator. We could just put a comma after each and then the developer would just have to remove the last one and place the output in an array.

Of course, EDN makes this easier since it doesn't require comma separators 🤔 . I think I'd rather either leave it like it is for now and make it EDN later, or make it EDN with this PR.

PEZ: Maybe leave it as it is for now and we'll see if we or someone ever want to work with it as data? It just looked a lot like data to me so I thought I'd share the idea. That said, how about something like so?

```suggestion
const enum Direction {
    ClientToServer = '->',
    ServerToClient = '<-'
}
```

bpringe: Okay, I'll leave it for now.

With just the arrows, I don't think it would be as clear/self-explanatory in the output what each arrow means. "Does -> mean going to the server, or going to the client?"

Cider uses `-->` and `<--`, but I'd rather not do what it does if we can be more explicit.

PEZ: I think it is harder to read as it is now. Client, server, it stops me in my thinking. I think the arrows are super clear and easy to parse. But i we can amend them a bit:

```
-> sent
{ ... }

<- received
{ ... }
```

bpringe: Ah, I see. I suppose Client and Server could be a little less clear than "Calva sent this and received this." I like your above suggestion!

bpringe:
I think the code is ready now, I just need to add info to the docs about it. 

@PEZ Where should this go in the docs? We have multiple places for troubleshooting info about specific subjects on those particular pages. I'm not sure if this warrants a new, general troubleshooting page, or if we should fit it into some existing page.

PEZ:
For now we can slap it in here https://calva.io/connect/#troubleshooting maybe? Hmm, that is not a perfect fit at all. Maybe instead create a page `cider-nrepl` that goes just before the `Clojure LSP` page, and put the info about this trace there, and link to it from the connection troubeshooting section. Then add an issue to give the new page more love. WDYT? My rationale is to keep it incremental so that this awesome feature does not get locked up in documentation jail. 😄

bpringe:
That sounds good to me! I'll be busy this weekend, but will get to it soon.

PR #1293, opened 2021-09-07
Fixes #1024, paredit kill right

### LINKED ISSUE #1024: Paredit kill right
Cont'd from #380.  

Emacs paredit-kill effects in slightly different ways than either Calva's kill sexp or kill list.

In all examples that follow, killListForward outcome would be `(a|)`.  
Whereas paredit-kill has different behavior in each (it's more granular in its effects).

### Example 1 - s-expresions

```clojure
(a| b (c
       d) e)
```
\+ paredit-kill = (expressions starting on the current line past the cursor are killed)

```clojure
(a| e)
````

### Example 2 - comments

```clojure
(a| ;; comment
  e)
```

\+ paredit-kill = (comment killed)

```clojure
(a|
   e)
````

### Example 3 - newline

```clojure
(a|
  e)
```

\+ paredit-kill = (newline killed)

```clojure
(a| e)
````

### PR DESCRIPTION
**Pull request by xfthhxk**
## What has Changed?

- Implemented paredit's kill-right aka smartparens' sp-kill-hybrid-sexp for emacs
- Added `paredit.forwardHybridSexpRange`
- Added command `paredit.killHybridSexpForward`

Fixes #1024 

## My Calva PR Checklist

I have:

Ping @pez, @bpringe

### CODE REVIEW

File: src/cursor-doc/paredit.ts [hunk 1]
```diff
@@ -138,6 +138,49 @@ export function backwardListRange(doc: EditableDocument, start: number = doc.sel
...
+ * @returns [number, number]
+ */
+export function forwardHybridSexpRange(doc: EditableDocument, offset = Math.max(doc.selection.anchor, doc.selection.active), goPastWhitespace = false): [number, number] {
+    const cursor = doc.getTokenCursor(offset);
+    if (cursor.getToken().type === 'open') {
+        return forwardSexpRange(doc);
+    }
+    cursor.forwardList(); // move to the end of the current form
+    const text = doc.model.getText(offset, cursor.offsetStart);
+    const newlineIndex = text.indexOf("\n");
```
xfthhxk: Do we need to check for `\r` and `\r\n` also?

docs/site/paredit.md
```diff
@@ -101,6 +101,7 @@ Default keybinding                | Action | Description
  `ctrl+backspace`                    | **Kill List Backward** | Deletes everything from the cursor to the opening of the current enclosing form.<br> ![](images/paredit/kill-open-list.gif)
  `ctrl+alt+shift+delete`                 | **Splice Killing Forward** | Delete forward to end of the list, then Splice. <br> ![](images/paredit/splice-killing-forward.gif)
  `ctrl+alt+shift+backspace`              | **Splice Killing Backwards** | Delete backward to the start of the list, then Splice. <br> ![](images/paredit/splice-killing-backward.gif)
+ `ctrl+k` | **Kill Right** | Delete forward to the end of the current form or the first newline.<br> ![](images/paredit/kill-right.gif)
```
xfthhxk, line 104: Need an image for `kill-right.gif`, but not sure how to.

xfthhxk:
Hi @PEZ! I think this PR is in a good spot now and ready for your review. The major missing item is the lack of the animated gif for the new paredit functionality.  Perhaps I can get your help with that.  Thanks!

docs/site/paredit.md
PEZ: I use GIF Brewery. Let me know if you want me to record it. I think it I can set it up pretty easily.

xfthhxk: @PEZ If you wouldn't mind, I'd greatly appreciate it! Thank-you.

src/cursor-doc/paredit.ts
bpringe: If I'm understanding correctly, `\r\n` would also work with your logic here, since it includes `\n` at the end. You may be able to test this by changing the end of line sequence in a file before running the command for it. You can change the end of line sequence by clicking the button for that in the status bar.

bpringe: Oh, actually, what I said is probably not true about it working for `\r\n`. I suppose you want the index of the beginning of the line ending, so you may need to check for `\r\n` too. You still may be able to test it as I said above.

PEZ: Ideally we should test this with some unit tests. Though I don't know if we have the facility to chenge line endings setting for the documents we create there. I think this might be the first non-structural Paredit command we add. As for testing it manually: keep in mind that Calva does not handle changing line ending settings ”in flight”, iirc. You probably need to close and reopen the file. (And we should fix that some day 😄.)

PEZ:
Thanks for taking this on, @xfthhxk! I am going to be using it all the time!

I checked out the branch and gave it a spin. Have to start work4food here, but can report three findings:

1. This is nice!
2. I can't find the command in the command palette.
3. I found some issues

Let me suggest we make a file `test-data/paredit-sandbox.clj` and use it for manual testing. There is a Launch Task configured in the repo for launching with the `test-data` folder. Here's some sample content, with @clyfe's examples and some more:

```clojure
;; = Paredit sandbox

;; == Paredit kill-right
;; https://github.com/BetterThanTomorrow/calva/issues/1024

;; === Example 1 - s-expresions
;; Expressions starting on the current line past the cursor are killed
(a| b (c
       d) e)

;; Example 1b
(aa| (c (e
         f)) g)

;; === Example 2 - comments
;; Comment killed
(a| ;; comment
 e)

;; Example 3 - newline
;; newline killed
(a|
 e)

;; Example 4 - end of list
;; Don't kill past it

(a b (c |)
   e)
```

(Or course, all these should be covered by unit tests as well. This sandbox is mainly an quick way to try out things when we get error reports, and to get a feel for how commands work.)

The issues I have found are:
* Example 1b: With the cursor in, or adjacent, to `aa`, the structure breaks. (With the cursor just before `(c ...` it works.)
* Example 3: Too much is killed, i get `(a|)`

src/cursor-doc/paredit.ts
xfthhxk: I think @bpringe is correct.  My worry about `\r` is somewhat unfounded. Macs switched to `\n` in 2001 with the release of OS X. I'll try testing as you suggest.

xfthhxk:
@PEZ good finds!
* I've added the command to the command palette.
* I'll put this PR in Draft until I can resolve the issues noted in example 1b and 3.

src/cursor-doc/paredit.ts
bpringe: I saw this PR was moved to draft and then out of draft and a review was requested from us. I just want to check - was this tested on Windows, or in any case, with CLRF (`/r/n`) line endings?

bpringe (commented):
I just want to say this PR looks great! Pending more testing by @PEZ to see if those issues he mentioned are fixed, and making sure it works fine with `/r/n` line endings, it should be ready to be merged.

src/cursor-doc/paredit.ts
xfthhxk: Yes I tested by clicking on the `LF`/`CRLF` button on the bottom of VS Code, closing and reopening the file.  I will add some tests for this as @PEZ has suggested.  I don't have access to a Windows machine.

xfthhxk:
Just discovered another edge case that occurs while invoking kill right inside of Clojure comments ie `;` that I need to fix.

xfthhxk:
@PEZ Ready for you to take a look (again)  :)

src/cursor-doc/paredit.ts
xfthhxk: Added a test for this also.

PEZ:
This is awesome @xfthhxk ! Super clean and nice balance with commenting the tricky considerations. And such a sweet test suite as well!

Will try to record the gif now. In preparing for that, I tried the *Example 9* in the test-data file. It works as advertised. However I wonder what should happen in this case:

```clojure
[|#_#_ (comment
         (+ 2 3)) :foo]
```

I get

```clojure
[|:foo]
```
But it somehow ”feels” like I should get:

```clojure
[|]
```

PEZ:
This one doesn't work I think:

```clojure
;; Example 6 -- to the newline
"This| is a multiline
string. "
```

`ctrl+k`, `ctrl+k`

```clojure
;; Example 6 -- to the newline
"This|
string. "
```

I expect it to end up like so:

```clojure
;; Example 6 -- to the newline
"This|"
```

It is sort of a promise with Calva Paredit that strings behave like lists. Looking at the code i don't really see why it doesn't in this case, though...

PEZ:
A slightly modified *Example 8* seems to not work?

```clojure
;; Exmaple 8 -- map key value pairs
;; killing from :c includes the corresponding value
{|:a 
 1
 :b 2
 :c {:d 4
     :e 5}}
```

(It only kills the `:a` key.) Maybe that is how it is supposed to work? If so there is nothing special with these binding cases, right? If it is supposed to always kill in pairs, we can probably refactor some of the code in the commands for dragging sexpressions forward and backward, because they consider binding pairs. (Also, I think we can leave this room for improvement, so not a blocker for the PR, imo.)

PEZ:
Actually. 😄 Maybe the string example I tried is just that it behaves like a list and that we actually do not do the right thing with the lists:

```clojure
;; Example 3 - newline
;; newline killed
(a|
 e)
```

`ctrl+k`

Should give:

```clojure
;; Example 3 - newline
;; newline killed
(a| e)
```

But instead nothing happens.

PEZ:
I think that last thing makes it a bit of a wrong moment to record the GIF, since it is a **killer** feature (sorry, couldn't resist). 😄 

I'm saving some test forms here that I was planning to use for the GIF:

```clojure
(defn foo []
  
  (comment
    (a| b (c
           d) e)
    (a| ;; comment
     e)
    "This| is a string"
    #_[a b (c d
              e
              f) g]
    :a
    ))
```

xfthhxk:
@PEZ I double checked the items you mentioned in your most recent comments.  Everything is working as expected per my usage of paredit/smartparens.  `kill-right` has the following options:
* Delete up to the first closing delimiter to the right of the cursor on the current line if there is one
* If there isn't a closing delimiter on the current line, it travels to the newline and walks backwards by sexp until it finds the nearest open delimiter to the right of the cursor, then travels to the end of that form to determine how much to delete.
* if neither of the above cases hold, it deletes everything between the cursor and the newline
----
Regarding:
```clj
;; Example 6 -- to the newline
"This| is a multiline
string. "
```

The behavior exhibited is consistent with lists/vectors. ie to the newline
```clj
[1| 2 3 4
 5]
=>
[1 
5]
```

---
In the following example, only the `:a` is deleted and works as expected because the role newline plays as mentioned above.
```
;; Exmaple 8 -- map key value pairs
;; killing from :c includes the corresponding value
{|:a 
 1
 :b 2
 :c {:d 4
     :e 5}}
```
---
Example 3 from #1024 doesn't match my experience with paredit's kill functionality.

The only difference between this implementation and what I see in Emacs is with regard to killing the newline itself.  
For this case 
```clj
;; Example 3 - newline
;; newline killed
(a|
 e)
```
`ctrl+k` for me moves the cursor to be on `a`. If I hit `ctrl+k` again then the `a` itself is deleted with the `e` remaining where it is and moves the cursor to  the `(`. Finally `ctrl+k` will delete the whole expr.  I can try to replicate this behavior if we want high fidelity with the Emacs implementation.

PEZ:
I think you have Evil mode on. I tried in Emacs without evil mode:

F6E6BC2E-5FED-44A5-B9B3-FD0C0DF16FF9-22745-00011EBABD2A8FB8

What Emacs does wrong here is that it kills the closing paren (and the closing quote of the string), otherwise I think it behaves  nicely.

I think that if we implement it like #1024 suggests, then anyone using the VIM extension will also get the Evil mode behaviour of ctrl-k (which I can't see why you would _want_ to have, but anyway).

xfthhxk:
@PEZ Updated to handle the newline as #1024 suggests.  Thanks again for all the testing on this.  I will be happy to buy you some beers at the next conj :beers:

PEZ:
Awesome, will try to find some time for it today. Looking forward to those beers! 🍻

PEZ:
This is much better. I have yet one more wish, though. =) 

```clojure
    (a|
     e)
```

<kbd>ctrl</kbd>+<kbd>k</kbd>, <kbd>ctrl</kbd>+<kbd>k</kbd>

Expected:
```clojure
    (a| e)
```

Result:
```clojure
    (a|     e)
```

There is also something funny happening here:

```clojure
    #_|[a b (c d
              e
              f) g]
    :a
```

<kbd>ctrl</kbd>+<kbd>k</kbd>

Expected:
```clojure
    #_|
    :a
```

Actual:
```clojure
    #_|:a
```

I'll test a bit more and see if I find something more funny. 😄

xfthhxk:
> This is much better. I have yet one more wish, though. =)
> 
> ```clojure
>     (a|
>      e)
> ```
> 
> ctrl+k, ctrl+k
> 
> Expected:
> 
> ```clojure
>     (a| e)
> ```
> 
> Result:
> 
> ```clojure
>     (a|     e)
> ```

If the expression is aligned left, the result is what is expected.  However in this example, the expression has 6 or 7 spaces before the `e` which is why there are spaces between the `|` and `e`. Seeing the same behavior in Emacs without evil mode.


> There is also something funny happening here:
> 
> ```clojure
>     #_|[a b (c d
>               e
>               f) g]
>     :a
> ```
> 
> ctrl+k
> 
> Expected:
> 
> ```clojure
>     #_|
>     :a
> ```
> 
> Actual:
> 
> ```clojure
>     #_|:a
> ```
I'm unable to reproduce this.  Works as expected for me.

src/cursor-doc/paredit.ts
```diff
@@ -138,6 +138,90 @@ export function backwardListRange(doc: EditableDocument, start: number = doc.sel
...
+        hasNewline = true;
+        end = currentLineNewlineOffset;
+    }
+
+
+    // handle weird CRLF case first otherwise we'll not advance far enough
+    if (remainderLineText === '\n' && '\n' === doc.model.getText(currentLineNewlineOffset, currentLineNewlineOffset + 2)) {
+        end = currentLineNewlineOffset + 2;
+    } else if (remainderLineText === '' || remainderLineText === '\n') {
+        end = currentLineNewlineOffset + 1;
```
PEZ: The model actually knows about the the length of the new-line. You'll find it in `doc.model. lineEndingLength`. Not sure I am reading the code right, but it looks like that's what you need?

PEZ:
> I'm unable to reproduce this. Works as expected for me.

I'm too unable to reproduce it now. So that's good, I think. Maybe just add it as a second **Example 9** in the sandbox? Since it has been observed to misbehave at some point. Maybe even a unit test.

As for the indenting space not being killed properly, I can probably live with that. Even if I am not super interested in copying over the glitches from the Emacs implementation. 😄 Again, documenting the current behavior in some test might be an idea.

src/cursor-doc/paredit.ts
```diff
@@ -138,6 +138,90 @@ export function backwardListRange(doc: EditableDocument, start: number = doc.sel
...
+    } else if (hasNewline) {
+        // Try to find the first open token to the right of the document's cursor location if any
+        let nearestOpenTokenOffset = -1;
+
+        // Start at the newline.
+        // Work backwards to find the smallest open token offset
+        // greater than the document's cursor location if any
+        cursor = doc.getTokenCursor(currentLineNewlineOffset);
+        while(cursor.offsetStart > offset) {
+            // backwardSexp() returns false when it gets to an open token type
```
PEZ: I think this comment is not necessary now that you have understood this. It is a fundamental behavior of the cursor movements. It's documented on `backwardsSexp()`.

src/cursor-doc/paredit.ts
```diff
@@ -138,6 +138,90 @@ export function backwardListRange(doc: EditableDocument, start: number = doc.sel
...
+            while(cursor.backwardSexp()) {}
+            if (cursor.offsetStart > offset) {
+                nearestOpenTokenOffset = cursor.offsetStart;
+                cursor = doc.getTokenCursor(cursor.offsetStart - 1);
+            }
+        }
+
+        if (nearestOpenTokenOffset > 0) {
+            // get the end of the list for the nearest open token
```
PEZ: I think you have named the variables and such here super clearly, making this comment not necessary.

PEZ (commented):
Can you check if `doc.model. lineEndingLength` should be used? If not, I think things look good. Many thanks for this help! 👍 ❤️

src/cursor-doc/paredit.ts
xfthhxk: Yes. Thanks for the pointer. Made the property available and using it.

PEZ: Sweet!

PEZ:
Here's a gif I made for the Visual Guide:

683F29F2-06AA-4A82-9C0B-786325237354-44970-000447F74032646B

You can see that the heisenbug with killing a newline too much when using it after the `#_` showed up while recording. So something is not right there. I think we should try expose it with a test. We'll record a new gif one's we have this fixed.

xfthhxk:
>  I think we should try expose it with a test. 

Added a test for this but still not able to reproduce.

PEZ:
Found out why the heisenbug. It's the formatter:

```clojure
#_|
:a
```

<kbd>tab</kbd>

```clojure
#_|:a
```

That's why we do not catch it with the current test. Formatting happens later. Still strange you can repro the behaviour, but anyway. 😄 

This tells me two things:

1. This can't easily be fixed. Maybe it. shouldn't even be fixed
2. We don't need to worry this indicates a bug in `forwardHybridSexpRange`

So then that means that we are ready to merge. The glitch in the GIF is transparent, at least. 😄

PEZ:
🎉 

Thanks for doing this, @xfthhxk ! As I said before, I am going to abuse it! 😄

PR #1336, opened 2021-10-13
[LSP] Add support for semantic tokens

### LINKED ISSUE #1231: [LSP] Support semantic tokens feature
I'd like to suggest adding support for LSP semantic tokens :)

### What is it?
[LSP semantic tokens](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_semanticTokens) is a feature introduced on latest spec release 3.16 and clojure-lsp is one of the few servers that support it. It was called semantic highlighting in the past and basically needs the server to return tokens, saying which they are, like functions, enums, namespace, variable, keyword, etc and the client color it on the document following user theme, giving to the user a better and improved highlighting on their code :)

### Do I need it for Clojure?

[Here](https://emacs-lsp.github.io/lsp-mode/tutorials/clojure-guide/#semantic-tokens-experimental) is a comparation of it on emacs with cider(REPL) connected:
image
Remember that this would make highlight works without the need of a REPL ;)

This is a screenshot of Calva without semantic tokens as of now with default theme:
image

This is a screenshot of Calva without semantic tokens as of now with monokai theme:
image

Emacs(lsp-mode) with semantic tokens enabled:
image

You can see how tokens like `@db/db`, `string/upper-case`, and especially for Clojure, __macros__, are highlighted with a specific color, also some themes don't handle well clojure keywords and things like that, which with LSP probably would fix it 

### What Calva needs to do?

On server side, this feature is enabled by default already (`:semantic-tokens?`) as it exists for months already and it works great IMO. 
On client side, it needs to send during initialization that it supports the semantic tokens. The good news are, AFAICS there is no need for Calva to configure anything as the `vscode-languageclient` already handles that, but I could not make it work for calva for some reason and I suspect that is related to the version of that lib, maybe upgrading to `7.0.0` just like Dart-Code, which supports semantic-tokens as well, [did](https://github.com/Dart-Code/Dart-Code/commit/a11cba2dd6ca38d086a01619296f79f1fd1847b8#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R2260)?

I think this would be a nice addition for Clojure users, LMK what you think :)
c/c @PEZ @bpringe

ericdallo:
## What has Changed?

- Bump vscode-langaugeclient to 7.0.0 to get latest LSP 3.16 features like semantic tokens

Fixes #1231

## My Calva PR Checklist

I have:

- [X] Tested the VSIX built from the PR (so, after you've submitted the PR). You'll find the artifacts by clicking _Show all checks_ in the CI section of the PR page, and then _Details_ on the `ci/circleci: build` test. NB: *You need to sign in/up at Circle CI to find the _Artifacts_ tab.*
     - [X] If I am fixing just part of the issue, I have just referenced it w/o any of the "fixes” keywords.

Ping @pez, @bpringe

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Changes to Calva.
 
 ## [Unreleased]
 
+- [Add semantic tokens support from LSP](https://github.com/BetterThanTomorrow/calva/issues/1231)
+
 ## [2.0.216] - 2021-10-10
 - Fix: [Inline results display pushes the cursor away when evaluation at the end of the line](https://github.com/BetterThanTomorrow/calva/issues/1329)
 - Fix: [Inline evaluation display renders differently than the REPL display for empty spaces](https://github.com/BetterThanTomorrow/calva/issues/872)
--- a/package-lock.json
+++ b/package-lock.json
@@ -1897,8 +1897,7 @@
         "balanced-match": {
             "version": "1.0.0",
             "resolved": "https://registry.npmjs.org/balanced-match/-/balanced-match-1.0.0.tgz",
-            "integrity": "sha1-ibTRmasr7kneFk6gK4nORi1xt2c=",
-            "dev": true
+            "integrity": "sha1-ibTRmasr7kneFk6gK4nORi1xt2c="
         },
         "base": {
             "version": "0.11.2",
@@ -2124,7 +2123,6 @@
             "version": "1.1.11",
             "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.11.tgz",
             "integrity": "sha512-iCuPHDFgrHX7H2vEI/5xpz07zSHB00TpugqhmYtVmMO6518mCuRMoOYFldEBl0g187ufozdaHgWKcYFb61qGiA==",
-            "dev": true,
             "requires": {
                 "balanced-match": "^1.0.0",
                 "concat-map": "0.0.1"
@@ -2736,8 +2734,7 @@
         "concat-map": {
             "version": "0.0.1",
             "resolved": "https://registry.npmjs.org/concat-map/-/concat-map-0.0.1.tgz",
-            "integrity": "sha1-2Klr13/Wjfd5OnMDajug1UBdR3s=",
-            "dev": true
+            "integrity": "sha1-2Klr13/Wjfd5OnMDajug1UBdR3s="
         },
         "concurrently": {
             "version": "5.3.0",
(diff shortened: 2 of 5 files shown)


ericdallo:
We can improve/customize the tokens [as we want](https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide#custom-textmate-scope-mappings) on the client as well, via package.json with something like this if you think it makes sense:
```json
		"semanticTokenScopes": [
			{
				"language": "clojure",
				"scopes": {
					...
				}
			}
		]
```
I just think the default looks good.

Without semantic tokens:
image


With semantic tokens:
image

Probably the most remarkable change is related to macros as clojure-lsp knows what function calls are macros and what are not. Also, you can notice that aliases take a different color :)

bpringe:
Awesome! I thought I remembered having some issues when I tried to update `vscode-languageclient` in the past - at least some small things that needed changing, more than those imports, but maybe not...

Did you test this change much besides the semantic tokens usage?

Do we need to add a setting if we want to allow disabling/enabling semantic tokens?

ericdallo:
Not much @bpringe, only basic features.

clojure-lsp already has a flag for that `semantic-tokens?`, WDYT?

bpringe:
It looks a little odd to me by default, but as you said we can customize it, right? Mine doesn't seem to look as good as what you posted though. 🤔 

image

Namespaces/aliases and vars are the same color it seems. I tried other themes and saw the same thing.

ericdallo:
@bpringe I'll test your piece of code

ericdallo:
@bpringe confirmed the same behavior, it seems to be a corner case on clojure-lsp with macros where we don't color the macro aliases, I'll take a look on server-side later today.
Besides that, I'm almost sure clojure-lsp doesn't return tokens for var definitions like `a-test` in your case, but I'll confirm and check if we can include it as well

ericdallo:
@bpringe made some fixes on clojure-lsp to show the alias colors for macros as well:

image
WDYT?

bpringe:
Looks great! I'll give it a test soon.

ericdallo:
Some more improvements I made on clojure-lsp, print:

Namespace color (green):
image

def and defn var name color (yellow):
image

bpringe:
Seems good to me @ericdallo. 

@PEZ Since users can control this via clojure-lsp settings, such as by using `.lsp/config.edn`, should we add a Calva setting for this. Ease of use by not having to know how to configure clojure-lsp is one reason to add a setting. Is that reason enough? 

We don't have Calva settings for other clojure-lsp settings, but maybe this one would be a bit jarring enough of a change to warrant one?

PEZ:
I think we can start without a Calva setting and see if we get feedback that indicates that it is needed. There are a few other lsp settings that I think we have indications that suggests Calva settings could be needed. Or clearer guidance at least.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ Changes to Calva.
 
 ## [Unreleased]
 
+- [Add semantic tokens support from LSP](https://github.com/BetterThanTomorrow/calva/issues/1231)
+
```
PEZ, line 8: Did you read the Hack on Calva docs? 😄 The empty line between the header and the list breaks our CI. FYI. I'll fix this before releasing.

PEZ (approved):

PR #1358, opened 2021-10-22
Fixes #1356 kill right key conflicts on Windows

### LINKED ISSUE #1356: ctrl+k for Kill Right conflicts with several bindings on Windows
The **Kill Right** (ctrl+k or cmd+k) command overwrites the **single line comment chord** on Windows (ctrl+k ctrl+c)

xfthhxk:
## What has Changed?

- Change `ctrl+k` for kill right to `ctrl+k ctrl+k` for windows and linux
- Change `ctrl+shift+k` for select right to `ctrl+k ctrl+shift+k` for windows and linux
- Mac kill/select right stay the same

Someone should probably test on Mac/Windows.

Fixes #1356 

## My Calva PR Checklist

I have:

     - ~[ ] If I am fixing just part of the issue, I have just referenced it w/o any of the "fixes” keywords.~

Ping @pez, @bpringe

### DIFF
--- a/package.json
+++ b/package.json
@@ -1923,7 +1923,7 @@
                 "mac": "ctrl+k",
                 "win": "ctrl+k ctrl+k",
                 "linux": "ctrl+k",
-                "when": "calva:keybindingsEnabled && editorLangId == clojure && editorTextFocus && paredit:keyMap =~ /original|strict/"
+                "when": "calva:keybindingsEnabled && editorLangId == clojure && editorTextFocus && !selectionAnchorSet && paredit:keyMap =~ /original|strict/"
             },
             {
                 "command": "paredit.killSexpForward",
--- a/package.json
+++ b/package.json
@@ -1767,7 +1767,7 @@
                 "command": "paredit.selectRight",
                 "mac": "ctrl+shift+k",
                 "win": "ctrl+k ctrl+shift+k",
-                "linux": "ctrl+shift+k",
+                "linux": "ctrl+k ctrl+shift+k",
                 "when": "calva:keybindingsEnabled && editorLangId == clojure && editorTextFocus && !selectionAnchorSet && paredit:keyMap =~ /original|strict/"
             },
             {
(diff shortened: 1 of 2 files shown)


PEZ:
We should only change the shortcut on Windows, I think.

I don't know if there's a conflict on Linux. Maybe @bpringe knows?

xfthhxk:
OK. Sorry I don't know how to make it OS specific. Is there an example I missed somewhere?

PEZ:
Yes, you can check `paredit.forwardSexp`. There are quite a few like that.

Background: We used to try super hard to keep the shortucts cross platform, but it turned out to be that everybody lost in some cases. So a while ago we threw in that towel, even if we still want to have that as a default go-to. In this case I think it is not good to punish Mac-users for stupidness on the Windows platform. 😄

xfthhxk:
> Yes, you can check `paredit.forwardSexp`. There are quite a few like that.

cool thanks for the pointer!

bpringe:
> I don't know if there's a conflict on Linux. Maybe @bpringe knows?

Just FYI, I'm primarily on MacOS now, though I still have my Linux machine for now so I can test things there if needed.

package.json
```diff
@@ -1913,7 +1915,9 @@
             },
             {
                 "command": "paredit.killRight",
-                "key": "ctrl+k",
+                "mac": "ctrl+k",
+                "win": "ctrl+k ctrl+k",
+                "linux": "ctrl+k",
                 "when": "calva:keybindingsEnabled && editorLangId == clojure && editorTextFocus && paredit:keyMap =~ /original|strict/"
```
Kryptos-FR: `!selectionAnchorSet` might also needed here, otherwise anchor selection won't work (which is what `Ctrl+K, Ctrl+K` does by default, after the initial anchor with `Ctrl+K, Ctrl+B` has been set).

xfthhxk: @Kryptos-FR thanks for looking this over.  Added the `!selectionAnchorSet`.

bpringe:
I think there is a similar issue with conflicts on Linux. Maybe the same fix for Windows would be good for Linux.

image

xfthhxk:
@bpringe thanks for testing. Updated the linux bindings to match windows.

PEZ:
Thanks, @xfthhxk! 

@Kryptos-FR, @xfthhxk: Do you think we should be caring about the selection anchor status more with some other shortcuts? I haven't been aware about it at all, I realize.

Kryptos-FR:
@PEZ it is a rare feature that not a lot of people are aware about. And as far as I know, this `selectionAnchorSet` is only use for that combination of shortcuts (set the anchor, then select from the anchor up to the caret position).
image
Therefore, I don't think other shortcuts are affected.

PR #1363, opened 2021-10-24
Add ClojureDocs Command for examples to Rich Comments and to Output window

### LINKED ISSUE #689: Include material from clojuredocs.org in function documentation
The clojuredocs.org documentation for clojure core is really handy. The examples are invaluable given how terse clojure docs can be. 

It would be great to have access to the clojuredocs for a given  function - whether added to the 'show hover' doc output used in the current implementation, or via a new command.

PEZ:
## What has Changed?

Reach Clojuredocs examples quickly.

Adding:
* ClojureDocs examples in the hover-lookup of symbols. So that you can at a glance see neat ways and caveats about Clojure core-ish symbols. Each example have two buttons/links:
    * **To Rich Comment**, prints the example to a rich comment below the top level form you are editing/viewing
    * **To Output Window**, prints the example to the REPL/Output window
* A command for printing all clojuredocs examples for the symbol under the cusror to a Rich Comment, default bound to `ctrl+alt+r d` (to go well with the command for opening a new rich comment block).
* A command for printing all clojuredocs examples for the symbol under the cursor to the Output/REPL window, `ctrl+alt+o d`

Here's a demo of the RFC one:

https://user-images.githubusercontent.com/30010/138604983-3d460990-2060-4bc3-9949-8cc06f7577d1.mp4

Also showing examples in hovers, with button at each example to print it as a rich comment. (Print to REPL window TBD.)

https://user-images.githubusercontent.com/30010/138616480-77fdfbb6-c705-4004-b04d-023e47ea30a9.mp4

Please also consider retweeting:

* https://twitter.com/pappapez/status/1452325865933746185
* https://twitter.com/pappapez/status/1452409528511762444

There are some quirks with this:

- Currently this only works with Clojure REPLs, because I am using cider-nrepl to get the examples. I'm working on getting clojure-lsp to be able to serve these as well, and expect no ”Clojure only” issues there. See https://github.com/clojure-lsp/clojure-lsp/issues/598 and https://github.com/clojure-lsp/clojure-lsp/pull/612
- When the examples contains unbalanced brackets, this introduces unbalance into the file. Not sure if we can live with that or need to somehow balance up the code (and not sure how to do that either 😄).

Fixes #689

## My Calva PR Checklist

I have:

Ping @pez, @bpringe


bpringe:
> When the examples contains unbalanced brackets, this introduces unbalance into the file. Not sure if we can live with that or need to somehow balance up the code

I think we should not worry about that for a first version of this, or maybe ever, but we can see how it goes. This is great work!

src/lsp/main.ts
```diff
@@ -397,6 +398,18 @@ async function openLogFile(): Promise<void> {
     }
 }
 
+export async function getClojuredocs(symName: string, symNs: string): Promise<any> {
+    if (serverVersion > '2021.10.20-16.49.47') {
```
bpringe, line 402: How does this `>` operator work when checking against a string?

bpringe: It seems that even if this check worked as it looks it was intended, there's a chance that `null` would be returned from this function just because the version was not a desired/capable version, and then the message [here](https://github.com/BetterThanTomorrow/calva/blob/7d33d621265e9d9f56a0c1ba3f7ac0f307ceaabf/src/clojuredocs.ts#L190) would be misleading, because `clojureDocs` would be `null` but symbol maybe would exist. It would say `Only Clojure core (ish) symbols supported` when really the version was not a desired/capable version.

I may have missed something, though.

PEZ: It's ”dictionary” order, character by character. https://javascript.info/comparison#string-comparison This type of check works because clojure-lsp has the kind of version string that it has. For other schemes, like Calva's, we would rather bring in a semver library, I think.

As for the message: good catch! I originally wrote that for when there were only direct commands fetching clojuredocs. Now when it happens on hover, it makes no sense to show a message. And we don't actually, so probably the `message` field should be removed. Which actually puts the whole `NoDocsEntry` construct into question (that might have been a bad idea all the time, I don't know 😄).

bpringe: Ah, thanks for educating me on that comparison!

Issue #1269, opened 2021-08-20
Calva not detecting tests when trying to run a single tests.


jmayaalv:
Calva is unable to detect test when trying to run a single tests. Test is found if all the tests in the namespace are executed.   

I could reproduce it on windows and can say that works on cider, so probably not something related to `cider-nrepl` the output when running a single tests is always:

``` clojure
; Evaluating file: reports_test.clj
#'edge.test.reports-test/sb-switches-report-test
; Running test: _…
; No tests found. :scream:, ns: 0, vars: 0
``` 

The same behavior can be seen if `deftest` is aliased or refered, so doesn't seem to be related to https://github.com/BetterThanTomorrow/calva/issues/1086

bpringe:
Thanks for the report!

bpringe:
I cannot reproduce this on MacOS. I suspect it may be an issue only on Windows. Do you see any errors in the developer console after reproducing the issue? `Help -> Toggle Developer Tools -> Console tab`

@PEZ Maybe you could try this on your Windows machine?

@jmayaalv If you'd like to debug and fix this, Calva has a [pretty easy contribution process](https://github.com/BetterThanTomorrow/calva/wiki/How-to-Hack-on-Calva).

jmayaalv:
Thank you for looking at this @bpringe. I am not really a VS code user myself and the ecosystem is very foreign for me, i use Cider and on macOS as well. I reported this on behalf of a new team mate totally new to the Clojure ecosystem.

We check the  the console but there doesn't seem to be anything relevant there, screenshot below:

![image (21)](https://user-images.githubusercontent.com/758737/132198740-105e6bd0-dd56-4f08-89b9-1dd141864410.png)

bpringe:
Thanks, it seems a good next step would be for someone to run Calva in the extension development host on Windows and set breakpoints to see where things are going wrong. CC @PEZ

marcomorain:
Can you try to reproduce this again with the latest Calva please? We added better error reporting, and I would expect to see a better error message in the REPL window now.

jmayaalv:
Hello @marcomorain, 
indeed, now with the latest version we still get "no test found" and an error in a test fixture. But at at least now the error is visible. Once we removed the fixture the tests can be executed without problem. 

Thanks a lot for the help !

marcomorain:
Awesome!

PR #1365, opened 2021-10-27
Handle the `unknown-op` status from test commands.

marcomorain:
Calva currently fails silently when the test commands fail due to the
unknown-op status. Here we handle that error and show a warning message
when it occurs. This doesn't address the root cause of the issue, but it
will help users to understand the problem.

This will help to debug #1269

Ping @pez, @bpringe

### DIFF
--- a/src/nrepl/cider.ts
+++ b/src/nrepl/cider.ts
@@ -1,0 +1,39 @@
+
+// https://github.com/clojure-emacs/cider-nrepl/blob/a740583c3aa8b582f3097611787a276775131d32/src/cider/nrepl/middleware/test.clj#L45
+export interface TestSummary {
+    ns: number;
+    var: number;
+    test: number;
+    pass: number;
+    fail: number;
+    error: number;
+};
+
+// https://github.com/clojure-emacs/cider-nrepl/blob/a740583c3aa8b582f3097611787a276775131d32/src/cider/nrepl/middleware/test.clj#L97-L112
+export interface TestResult {
+    context: string;
+    index: number;
+    message: string;
+    ns: string;
+    type: string;
+    var: string;
+    expected?: string;
+    'gen-input'?: string;
+    actual?: string;
+    diffs?: unknown;
+    error?: unknown;
+    line?: number
+    file?: string;
+}
+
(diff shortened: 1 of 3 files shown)


marcomorain:
image

marcomorain:
This PR needs work before it can be merged – when the test command fails the new REPL prompt is not printed (as seen in the screenshot)

marcomorain:
Aha, the calls to `await session.test` will throw, which jumps over the call to `outputWindow.appendPrompt()`

marcomorain:
@PEZ this should be good to review now.

marcomorain:
image

I've moved the error message into the REPL output, rather than the VS Code warning pane.

PEZ:
Thanks!

LGTM. Have you had a look, @bpringe?

bpringe:
@marcomorain Thanks for this PR.

I installed the vsix and when I run this simple test using the `Run Current Test` command the repl seems to either freeze or stop printing things.

```clj
(ns test-lein.core-test
  (:require [clojure.test :as test]
            [test-lein.core :refer :all]))

(test/deftest a-test
  (test/testing "Some test"
    (test/is (= 1 1))))
```


I just get this message with nothing else:

```clj
; Running test: a-test…
```

Then I can't evaluate anything else, even if I run the command to interrupt evaluations. At least, I'm not seeing any results printed. Using the latest Calva version, this scenario works as expected.

marcomorain:
@bpringe thanks for that – I was only testing the negative case when I was running this, and never tested actually running tests 🤦 

Is there a way to run a test like yours in CI?

marcomorain:
This is ready for review again 🙏

bpringe:
> Is there a way to run a test like yours in CI?

What do you mean, exactly?

I tested the VSIX and it seems good to me!

marcomorain:
> What do you mean, exactly?

I mean, is there a way that we write an integration test of that starts Calva, connects to a REPL server, evaluates some source, and runs a test?

bpringe:
@marcomorain Oh, it may be possible. I'm not sure how easy it would be to involve a repl in integration tests, though.

PEZ:
It would be a super awesome addition to be able to run tests like those in CI! Please open an issue about it. And a PR is of course super duper welcome. We have an integration test setup that we run in CI. It only starts VS Code and checks that Calva loads. (Or some such). That's where we would hook this test in, most probably.

PR #1375, opened 2021-11-03
Remove inaccurate anti-parinfer propaganda ;)

cap10morgan:
In all seriousness, the statements about paredit in here that I removed / edited apply equally well to parinfer. It isn't non-structural editing, it's just more implicit by letting indentation specify structure rather than requiring explicit editor commands. I think your points in here are stronger when they aren't needlessly vilifying a straw-man parinfer. I think this is important for especially newcomers to Clojure / Lisps where there is already so much to learn and parinfer can be a really friendly way to get started without having to manually balance parens nor learn paredit. Thanks for considering!

### DIFF
--- a/docs/site/parinfer.md
+++ b/docs/site/parinfer.md
@@ -7,13 +7,11 @@ description: Yes, you can use Calva with with the Parinfer extension
 
 Yes, you can use Calva with with the Parinfer extension. The only conflict here is that Calva's auto-formatting (that which happens as you type) is going to risk clash with Parinfer's.
 
-However, Calva's [Paredit](paredit.md) is carefully crafted. We suggest you consider that Clojure is a LISP, and therefore structural. There is power in this structure that other languages can not let you wield. Take advantage of this and make Paredit your friend. It takes only some few minutes to learn two or three basic Paredit commands:
+However, Calva's [Paredit](paredit.md) is carefully crafted. If you don't already have an established preference for parinfer, we suggest you give it a try. Even parinfer veterans may find they prefer the explicit nature of paredit. It takes only some few minutes to learn two or three basic Paredit commands:
 
 * **Select form**
 * **Slurp**
 * **Barf**
-
-Besides letting you benefit from auto-formatting, working this way will quickly bring you in better contact with the structural nature of the Clojure code. You'll thank yourself later.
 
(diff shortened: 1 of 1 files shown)


PEZ:
I don't read in anti-parinfer propaganda there. 😄 It's my own experience that while I was using parinfer I didn't ”get” the thing about the structure. And when I pair program with Parinfer wielders I always notice how they do so much manual editing around when a ParEdit command would just fix it for them. This is even though they have ParEdit at hand as well.

Anyway, I'll consider your changes. Unfortunately I have been pushing and force pushing some stuff under your feet, so you will need to rebase you branch on latest `dev` and force push that up in order to clean this PR up. Sorry about the inconvenience!

cap10morgan:
@PEZ Makes sense. I went the other way and started with paredit before parinfer existed but then tried parinfer when it became available and didn't look back. Suddenly the editor was doing something for me based on input I was already giving it (indentation) that I used to have to do manually on top of the indentation formatting. It's like the best of both worlds of significant whitespace but without the ambiguity. Funny that you see more manual editing w/ parinfer users. For me it's the exact opposite!

I love both paredit and parinfer, to be clear. I'm trying to see if I'm ready to migrate from Cursive to Calva, but one of the long-time hindrances has been this page strongly implying I was a second-class citizen in Calva-land due to my parinfer preference using reasoning that didn't ring true to my own experience.

Now that you mention it, trying out a parinfer-first with paredit in a few corner cases approach might not be terrible. 🤔 

Anyway, no worries either way. I'll rebase on dev shortly.

Thanks for being open-minded about it!

cap10morgan:
This... got messier after rebasing on dev. I'm going to cherry pick the only relevant commit into a new dev-based branch and force push that.

cap10morgan:
OK cleaned up

PEZ:
Thanks! I'd like to take a step further now. 😄  Let's do what you suggest and not make it about parinfer vs paredit _preference_ at all. More be open about that there will be dragons and that disabling Calva's autoformatter might create a bad experience. So that's the basis for the suggestion to consider learning some few Paredit commands and skip Parinfer + Paredit strangeties.

But before we go and update the docs again... I actually don't know what the experience is, because I haven't tried in a while. Maybe we should both try it a bit and then revisit the docs?

PEZ:
I have spent some considerable time trying to add Parinfer natively to Calva. I really think it should be fully possible to use both Parinfer and Paredit together. My previous efforts have been about supporting that third mode of Parinfer that ”does the right thing” regardless if it is structure that needs to be healed of indentation that needs to be fixed. That really seems impossible to implement in VS Code (I am not the only one having given it a go).

If we narrow to requirements to let a Parinfer mode only deal with fixing structure based on indentation. Then Calva's formatter can take care of indenting things based on structure, same as it does today. IDK, maybe it could work!

Issue #1379, opened 2021-11-10
Version v2.0.224 causes problems on some machines (possibly Windows related)


PEZ:
* UPDATE, Nov 11: The latest 226 VSIX build seems to have fixed the problems. It's the one linked in this issue. Please keep testing it. I don't want to release another breaking release too soon. 😄 
* UPDATE 2, Nov 11: I've received report that the problem show up in 226 still, just in a different ”place”. **Paredit delete to end of list** was an example.
 
Symptoms. Editing files in CRLF mode, entering brackets instead inputs other characters. 

If you just need to work and do not have time for helping with fixing, use v2.0.225 which reverts the changes.

If you have a kind heart and want to help me bring Parinfer to Calva, know that there is a 226 dev build that can we can use for investigating this together. please read on:

Why it happens I don't know, but it has to do with a new method for auto-closing brackets, which was enabled by default in 224 and made impossible to code for the users where it didn’t work.

I can't reproduce on Mac, nor on Windows, so I really need help. If you can reproduce the problem and want to [run Calva in debug mode](https://github.com/BetterThanTomorrow/calva/wiki/How-to-Hack-on-Calva) to help pinpoint, that would be so awesome. ❤️ 

## How to test with dev 226

* Download and install [v2.0.226 VSIX](https://15952-125431277-gh.circle-artifacts.com/0/tmp/artifacts/calva-2.0.226-dev-ec9f3fcb.vsix)
* Read the [WIP docs](https://github.com/BetterThanTomorrow/calva/blob/dev/docs/site/parinfer.md) (Github doesn’t render it very well, but anyway)

This is what I currently want tested:

* Used with its defaults (Parinfer and Format Forward off) it should work without troubles.
* Used with Parinfer and Format Forward on, it should deliver the experimental features (mostly trouble free)
* Used with the new auto-closing it will keep causing problems in CRLF files, under some unknown circumstances. 226 can therefore be used to help figure out a way for me to reproduce the problem. (Which I can’t yet.)

### General, as-things-were, testing

To test that it doesn’t cause problems ”as is” run it without changing any Calva settings.

### Parinfer fans

To test the experimental features, use these settings:

```json
  "calva.fmt.formatForward": true,
  "calva.fmt.inferParensAsYouType": true,
  "[clojure]": {
      "editor.autoClosingBrackets": "never",
  }
```

### Bug hunters

To try provoke the problem, edit files in CRLF mode and have these setting:

```
[...]

PR #1381, opened 2021-11-12
Parinfer take 2

### LINKED ISSUE #253: Feature request: parinfer
Do you have any plan to support parinfer?  I prefer it to paredit and it would be great to have it as an integrated choice.

PEZ:
## What has Changed?

Employing a combination of [Parinfer](https://github.com/oakmac/parinfer) **indent mode** (infer brackets from indentation) and Calva's formatter to create an experience where, as-you-type, things keep being properly indented and where the user can restructure the code using indentation. NB: _It is **mode-less**, but it is **not** Parinfer SmartMode._

Here's a demo showing a bit what it looks like in action:

calva-parinfer-and-format-forward

Another change that is currently included here is that *Paredit Strict Mode bracket matching* is using Parinfer regardless if **Infer Parens** is enabled or not.

These are the settings added/affected:

* `calva.fmt.experimental.inferParensAsYouType`, default `false`
  * `calva.fmt.experimental.alertOnParinferProblems`. default `true`
* `calva.fmt.experimental.formatForward`, default `true`
* `calva.paredit.strictPreventUnmatchedClosingBracket`, default `true`, now uses Parinfer
* `calva.paredit.strictAutoClosingBrackets`, default `true`, now uses Parinfer

In coming iterations I'm considering making `formatForward` a part of `format as you type` and thus remove this setting. I'm also considering making bracket matching default included in Paredit strict mode, and remove those settings as well.

A status bar indicator/button is added which shows:

* **Structure and indentation healthy**. In this state the button toggles **Infer parens** on/off
* **Structure is broken**. In this state **infer parens** is disabled. Also shows an alert, which can be set to not show using `calva.fmt.experimental.alertOnParinferProblems`
* **Indentation needs to be fixed**. In this state **infer parens** is disabled and the button fixes the indentation. Also shows an alert, which can be set to not show using `calva.fmt.experimental.alertOnParinferProblems`

The command **Calva Parinfer: Fix Document Indentation** is added and enabled when the indentation of a document is not matching the structure.

### Design decisions

* I want it to be mode-less. So I will try to keep it like that until I have exhausted myself and then I might give up and go for modes. (Unlikely, this mode-less solution actually seems to work.)
* I want it to be non-intrusive and pop up dialogs and messages unless absolutely necessary.
* I want it to be as few settings and decisions for the user as possible, even if in this first iteration it is still in _experimental_ state so we will probably keep some more settings in order for users to be able to test things independent of each other.

### TODO:

There are some things that currently stops this from being released even as an experimental feature:

* [x] **Blocker**: Some changes in this PR makes structural editing as such break for some users (Windows and CRLF, and then some, because I can't reproduce on Windows). This is being tracked in #1379. This fully blocks releasing this.
* [x] **Blocker**: There is something wrong with how the extra edits work with the undo stack. Sometimes you can't undo back to a state when VS Code thinks the document is not ”dirty”. Worse, sometime the mirror document gets out of sync.
* [x] Parinfer indentMode will modify the structure even when the indentation is not correct, risking the file being incorrectly modified in places where the user might not even notice. I want to have this fixed before releasing, but _will release w/o if #1379 gets fixed_.
* [x] The paren inference will not work at all when the structure is broken, and currently there is often no clue about why it is not happening for the user.  I want to have this fixed before releasing, but _will release w/o if #1379 gets fixed_.
* [ ] **Blocker**: There are still problems with CRLF files: After editing such that a form is being ”pushed” forward, deleting the inserted text doesn't work.

And some nice-to-haves, that might or might not make it into this iteration.

* [x] Since Paredit strict mode bracket matching uses parinfer, it is affected by the issues above about indentation not being correct. We might need a fallback method or something...
* [ ] Currently Parinfer will often move brackets back when you try to move them down. It is very annoying, and I am figuring on ways to avoid it. It's extra annoying with Calva's [”open” Rich Comments](https://calva.io/rich-comments/#thinking-space-is-kept).
* [ ] ...

Fixes #253

See also:
* #1377
* #1379 

## My Calva PR Checklist

I have:

Ping @pez, @bpringe


svdo:
I found the following issue. Suppose you have the following content of a clojure file:

```clojure
(ns foo
  (:require [clojure.string :refer [blank?
                                    lower-case
                                    split
                                  trim]])) ;; note: incorrect indentation

(def foo)
```

Please note that the indentation of `trim` on line 5 is incorrect.

If you now type an open paren insed the `(def foo)`, the closing brackets of the `trim` line will be changed:

```clojure
(ns foo
  (:require [clojure.string :refer [blank?
                                    lower-case
                                    split]
                                  trim]))

(def foo())
```

The larger issue is: how to deal with incorrectly formatted files? Some possibilities:

- Ignore this issue and leave the behavior as-is (my nvim setup does it like this too for this specific case)
- Give a warning about incorrect indentation and offer to automatically change the indentation to match the braces before starting to edit
- Don't give that warning and simply fix it before editing (I have seen this as well in my nvim setup, not sure what determines whether this happens or the wrong thing)

Personally I don't like the current behavior, because it breaks my file in places that are completely unrelated to where I'm editing, that is very annoying, especially for larger files. I would be happy with either of the other two options.

PEZ:
I also think it is undesirable to break the structure. I will check if there is a way we can tell parinfer not to do that and instead fail. I think I saw something about it in the API.

As for the Ux... I'm not fond of just fixing it either. Some implementations I've used are naggy and I don't like that. We'll have to figure some. =)

svdo:
I just also had a case where incorrect indentation caused the parinfer to stop working entirely:

```clojure
(ns foo
  (:require [clojure.string :refer [blank?
                                    lower-case
                                    split
                                    trim]]))

(def foo [])

(defn some-fn [arg]
  (let [foo (map #(get-in % [:key1
                           :key2]) arg)] ;; note: incorrect indentation causes parinfer to stop working
    :foo))
```

svdo:
But please don't let yourself get discouraged because this is AWESOME!!!! :)

PEZ:
Thanks for including examples as text. You should tell @bpringe how much better that is than gifs and pngs. 😂

oakmac:
> The larger issue is: how to deal with incorrectly formatted files? 

A requirement for Parinfer is to run Paren Mode on a file before dropping into Indent or Smart Mode. See [Fixing Existing Files](https://shaunlebron.github.io/parinfer/#fixing-existing-files) for more details.

> Give a warning about incorrect indentation and offer to automatically change the indentation to match the braces before starting to edit

This is the approach taken by [atom-parinfer](https://github.com/oakmac/atom-parinfer) and [Sublime Text Parinfer](https://github.com/oakmac/sublime-text-parinfer).

Screenshot from Atom Parinfer when a file is opened that meets this criteria:

Screen Shot 2021-11-12 at 1 13 08 PM

PEZ:
Yes, so that is what I'd like to avoid, if I can. Naggy. We'll see if we will need to do it this way eventually.

PEZ:
It seems Parinfer insists on changing ”unrelated” parts of the file, even if I give it all the context I have around the edit. So that line of mitigations is off the table... Maybe if we consider inferring only within the current enclosing form, like the Calva formatter does, that we can accept the globalishness...

oakmac:
> Yes, so that is what I'd like to avoid, if I can. Naggy. We'll see if we will need to do it this way eventually.

If you can ensure that the entire structure of the file is balanced, then it may be safe to drop into Indent and/or Smart mode for the entire file. Otherwise it is possible that Parinfer will change the structure of the file, which may introduce bugs.

Alternatively, Parinfer may **find a bug** based on indentation. Here is a [real-world example](https://github.com/oakmac/atom-parinfer/commit/d4b49ec2636fd0530f3f2fbca9924db6c97d3a8f) of this happening as I was developing atom-parinfer.

> It seems Parinfer insists on changing ”unrelated” parts of the file, even if I give it all the context I have around the edit.

Indent and Smart mode are "aggressive" about creating structure based on indentation. The internal algorithm removes all closing parens that exist at the end of a line and replaces them based on indentation + opening parens.

You might find inspiration in the [parent expression hack](https://github.com/oakmac/sublime-text-parinfer/issues/23) as a way to only look at top-level forms (ie: in a normal Clojure namespace this is mostly `def`s and `defn`s) for applying Parinfer in a limited scope. atom-parinfer and Sublime Text Parinfer both use the "parent expression hack" for performance and to-date nobody (except Shaun in the linked thread) has ever commented on this being a problem for day-to-day development. Apparently, Lisp users tend to keep their parens balanced :stuck_out_tongue_winking_eye:

oakmac:
> But please don't let yourself get discouraged because this is AWESOME!!!! :)

I just want to second this. Having a solid Smart Mode Parinfer implementation for VS Code has been on my "TODO" stack for years now. I am super glad to see an effort to make this happen.

Please let me know if I can be of any help! Either in this thread or on a video call, etc.

Strong work @PEZ  :muscle: :nerd_face: :tada: 

cc @shaunlebron @aarandela @gauraklein @skershaw (plus many others who I know are excited about VS Code + Parinfer)

svdo:
> Yes, so that is what I'd like to avoid, if I can. Naggy. We'll see if we will need to do it this way eventually.

I can see why you feel that way, but for me it's actually the other way around. I recently renamed a lot of things with global search/replace, causing white space to be incorrect. I would actually consider it helpful when the editor suggests to fix it when I open the file. This is something that happens relatively rarely, right? At some point all files will be ok and you will not see the warnings anymore. 

There could even be a command to globally fix _all files_ in a workspace as a separate step, by the way; then you can commit that and be done. This is very common in JS world with tools like `standard`. Maybe something for the formatter?

PEZ:
@oakmac : FYI. This is not implementing SmartMode. I'm using a combo of IndentMode and Calva's formatter. There is some tiny little more context here: https://github.com/BetterThanTomorrow/calva/pull/1377

PEZ:
That parent expression hack sounds very similar to how Calva's formatter works, so I might explore that path a bit.

PEZ:
I've now updated the issue description to better guide the discussion here. 😄

oakmac:
> There could even be a command to globally fix all files in a workspace as a separate step, by the way; then you can commit that and be done. This is very common in JS world with tools like standard. Maybe something for the formatter?

FYI - [parlinter](https://github.com/shaunlebron/parlinter) was created for this purpose :+1:

svdo:
> FYI - [parlinter](https://github.com/shaunlebron/parlinter) was created for this purpose :+1: 

Oh wow that's the first time I heard of it, very nice!

PEZ:
Now updated with checks for if indentation and structure allows brackets inference or not. And some UI to relay this to the user and offer help to fix broken indentation. See PR description for some more details. Here's a VSIX with the changes as of this writing, for your convenience: https://16000-125431277-gh.circle-artifacts.com/0/tmp/artifacts/calva-2.0.226-pez-253-parinfer-9203d0ae.vsix

svdo:
This looks great @PEZ! I like the warning + ability to have it unobtrusively + command/button to quickly fix, well done!

PEZ:
New build, considering readonly documents before nagging about structure and indentation: https://16010-125431277-gh.circle-artifacts.com/0/tmp/artifacts/calva-2.0.226-pez-253-parinfer-4a1356b6.vsix (Good when navigating to library code.)

svdo:
Just ran into a case where this doesn't work yet:

```clojure
(ns foo)

(defn foo []
  (my-fn :foo {:bar (bar-fn)
               :baz (baz-fn)}))
    
```

When I put the cursor between `:foo` and `{:bar` and I type a space, the line below is not moved and instead the parens are modified, in other words: `:baz` and `(baz-fn)` are ejected from the map.

svdo:
Also, I didn't check if you already thougth of this, but just trying to help: the indentation check should probably also be done when the code changes for other reasons than just typing, e.g. cut, paste, live share, etc...

PEZ:
> When I put the cursor between :foo and {:bar and I type a space, the line below is not moved and instead the parens are modified, in other words: :baz and (baz-fn) are ejected from the map.

Not sure which build you where using, but in one of them I had a bug like that. Try with this one and let me know if it is still misbehaving:

https://16015-125431277-gh.circle-artifacts.com/0/tmp/artifacts/calva-2.0.226-pez-253-parinfer-fcc7baa3.vsix

PEZ:
> Also, I didn't check if you already thougth of this, but just trying to help: the indentation check should probably also be done when the code changes for other reasons than just typing, e.g. cut, paste, live share, etc...

It should be described better in the PR text, but anyway. Right now we run structure integrity checks whenever we open a new file and then every time it changes for whatever reason. Also indentation integrity is checked at the same time, but I don't think we can pick up the reason for the change so if bad indentation is introduced from ”underneath” I think we can't really guard against that, because we can't distinguish it from the user wanting to slurp or barf something.... We will infer parens and then check the integrity. The Live Share case is extra interesting. We should try that.

svdo:
Ok, that build indeed seems to fix the issue (I was using the one you linked before, from this morning). 

Regarding the checking: what you describe explains what I just noticed, namely: when I slurp something the mechanics of slurping apparently cause the indentation to briefly break, because I get the warning popup (even though right after it is ok again.)

PEZ:
> Regarding the checking: what you describe explains what I just noticed, namely: when I slurp something the mechanics of slurping apparently cause the indentation to briefly break, because I get the warning popup (even though right after it is ok again.)

Thanks. Do you mean when you do Paredit slurp?

PEZ:
I don't get that error. Regardless if I use paredit of parinfer for slurping...

PEZ:
Alright so the change from underneath is problematic. I just tested with having a branch where the indents are wrong for a file. When switching to that branch when the file was active. It was wrongly ”fixed” by parinfer... I don't think there is much I can do about it... OTH the change gets put on the undo stack so the user can undo it...

What is way worse. If I have some other file open. It gets totally wrecked. I think we are pretty far from releasing this! 😢

svdo:
Why was it wrongly fixed? Because it was using the represtation of the file from before the branch switch? Could you maybe add a hidden flag to a document to indicate whether it has been checked? So that if vscode changes the document from underneath you, you can detect that it hasn't been checked yet and treat it differently?

svdo:
Btw just had an exception in the extension host: on line https://github.com/BetterThanTomorrow/calva/blob/fcc7baa3a1414c53fe808c2ed62ff342200ca54a/src/providers/definition.ts#L16 it said it could not get `info` on a `null` object.

PEZ:
The `info` thing is probably unrelated to parinfer.

PEZ:
It was wrongly fixed because it had bad indentation.

PEZ:
And I can't distinguish the change from if the user has tried to get the parens to move around.

docs/site/parinfer.md
```diff
@@ -1,18 +1,39 @@
...
+    TL;DR: When enabling `calva.fmt.experimental.inferParensAsYouType`, _you need to also disable `editor.autoClosingBrackets`_.
+
+    For Parinfer's inference of the closing bracket when you type an opening one to work, VS Code's built-in `editor.autoClosingBrackets` setting needs to be **disabled**, and Calva currently **enables** this by default. This is because there are currently some troubles with Calva's backup auto-closing features (`calva.paredit.strictAutoClosingBrackets`).
+
+!!! Note "Multi-cursors not fully supported"
+    Calva only really considers the first cursor in a multi-cursor senario. Sometimes that's enough, often it is not.
+
+## Paredit is still there
+
+In Calva, Parinfer and [Paredit](paredit.md) are designed to coexist and both be there to let you edit the structure easily and efficiently. Since Pardit commands are always formatted, they leave the code in a state where Parinfer has what it needs to infer bracket placement as you either edit the indentation, or remove/add brackets.
```
oakmac:
```suggestion
In Calva, Parinfer and [Paredit](paredit.md) are designed to coexist and both be there to let you edit the structure easily and efficiently. Since Paredit commands are always formatted, they leave the code in a state where Parinfer has what it needs to infer bracket placement as you either edit the indentation, or remove/add brackets.
```

PEZ:
There is something funny going on with the undo stack in this branch. I've updated the PR about it. I'm starting to think I will not be able to make parinfer happen, unless I come up with a new idea of how to implement it, or re-write big parts of Calva's editing infrastructure. We'll see, maybe I'll figure this undo-thing out...

PEZ:
I fixed the undo bug. `git bisect` FTW. Might have fixed that CRLF bug in the same go. There is hope!

Here's a new convenience build link: https://16036-125431277-gh.circle-artifacts.com/0/tmp/artifacts/calva-2.0.226-pez-253-parinfer-610b4a2f.vsix

oakmac:
> Here's a new convenience build link: https://16036-125431277-gh.circle-artifacts.com/0/tmp/artifacts/calva-2.0.226-pez-253-parinfer-610b4a2f.vsix

I pulled down this package and played around some.

First impressions: this is really great! It feels very close to the "normal" Parinfer experience and will probably suffice for the majority of Parinfer users. Strong work @PEZ :muscle: 

From the perspective of a Parinfer user there were two behaviors I quickly noticed:

### Tab Keys

When I press <kbd>Tab</kbd> and <kbd>Shift</kbd>+<kbd>Tab</kbd> I expect it to indent and dedent the text. This is also true if multiple lines are selected.

As a super bonus / nice-to-have, the [parinfer.js API](https://github.com/oakmac/parinfer#api) returns `tabStops` which can be used to align text to the nearest open bracket point.

### Comments + cursor position

```clj
(defn foo []
  (let [a "a"
        b "b"]
    (bar)))
```

If I put my cursor before `(bar)))` and press `;`, I expect the cursor to be **after** the semicolon. Currently it is re-positioned before the semicolon.

<hr>

> NB: It is mode-less, but it is not Parinfer SmartMode.

I understand your intention here is not to implement Smart Mode verbatim. I will post more if I notice anything else :nerd_face:

PEZ:
Thanks, @oakmac! For the words of encouragement, for the help testing, and for the feedback! It means a great deal to get help from someone who's implemented and tested Parinfer implementation all over.

**About the tab key**. You are not alone in noting this expectations ”miss”. I'm not sure what to do about it yet. There are at least two aspects:
* **Expectations**: Calva users certainly expect that key to format the current form, Parinfer users will expect it to indent.
* **Implementation**: Just using VS Code's built-in Indent/dedent commands create an extra undo stop that I don't seem to be able to get rid off (I might figure it out, but right now...). Also it is a bit stupid about where to actually indent things. Implementing separate commands for it is a possibility and would address both. (Is this where I would use the tab stops info?)

For now this is probably best to deal with from the documentation side, informing that space and backspace are the closest equivalents. Maybe some good ideas will arrive to us on what to do.

**That cursors position**. Tricky. What I do now is take the new cursor position that parinfer gives me and add/subtract whatever character count has been added/subtracted before the cursor position by the inference operation. Despite it being a bit strange that I need to do the accounting, It seems to work for all cases but the comment. I think I am misusing the parinfer API somehow. When we find that error, I think this one will be fixed as well.

**The SmartMode note**. My intention is to set the expectations right. Some two years ago or so I spent a lot of time trying to get SmartMode implemented. For whatever reasons I just couldn't get it right. I then decided that Calva users will have to do without Parinfer. Then, since a while, I've started to think about it something like so: Calva has a pretty good auto-formatter. It is doing much the same as what `parenMode` does, and then some. What if I use that formatter to first indent things, and then infer parens? Could that create a mode-less experience that most often does what we want? In essence just take the _smart_ out of the equation and be stupid about it instead. I honestly don't know where it differs from SmartMode, but for someone who knows SmartMode well, I just want it to be clear that it is not what is going on.

PEZ:
**Bad news**: The CRLF bug is not fixed.

**Good news**: I can reproduce it now. And this is a Calva day for me, thanks to my employer, so maybe I can fix it?

PEZ:
CRLF bug is fixed. I have forgotten to update about that. 😄

oakmac:
> Thanks, @oakmac! For the words of encouragement, for the help testing, and for the feedback! It means a great deal to get help from someone who's implemented and tested Parinfer implementation all over.

:heart: 

> Calva users certainly expect that key to format the current form, Parinfer users will expect it to indent.

This may be solvable with a setting?

> Also it is a bit stupid about where to actually indent things. Implementing separate commands for it is a possibility and would address both. (Is this where I would use the tab stops info?)

I think implementing separate commands might work! Yes - the [tabStop info] from Parinfer should be able to tell you exactly where a user would expect the cursor to land on indent and dedent operations.

[tabStop info]:https://github.com/oakmac/parinfer/blob/master/test/cases/indent-mode.md#tab-stops

> What I do now is take the new cursor position that parinfer gives me and add/subtract whatever character count has been added/subtracted before the cursor position by the inference operation.

This sounds like a solvable edge case to me, but I have not dug into the Calva code personally yet.

> Some two years ago or so I spent a lot of time trying to get SmartMode implemented. For whatever reasons I just couldn't get it right.

I had this exact same experience :neutral_face:. I do not think the VS Code Buffer API can truly support what SmartMode needs.

> Calva has a pretty good auto-formatter. It is doing much the same as what parenMode does, and then some. What if I use that formatter to first indent things, and then infer parens? Could that create a mode-less experience that most often does what we want?

This is **brilliant** and from my testing last night it works great to match what Smart Mode is doing. This is very close to how the Parinfer algorithm works internally.

PEZ:
There are still some troubles with CRLF files. New troubles. I've added a note in the description. I don't know if it makes sense.

PEZ:
The new troubles in CRLF files were pretty horrible, even in the light of all crazy things that these Parinfer PRs have resulted in. An infinite loop this time. Caused by the previous bug fix. Now fixed. I hope w/o causing a new bug. Anyway, here's the convenience VSIX you should use: https://16066-125431277-gh.circle-artifacts.com/0/tmp/artifacts/calva-2.0.226-pez-253-parinfer-4143488c.vsix

PEZ:
Cross-linking to the Twitter mega thread about this journey. https://twitter.com/pappapez/status/1461018167002734592

svdo:
Still very happy with this :) I do think that the extension host crash that I mentioned above is related to this branch by the way; I'm getting it multiple times a day, whereas I have never seen it before I started testing this...

PEZ:
@oakmac I'm added your latest input to my TODO list for Calva Parinfer. Now I have a question for both you an @svdo:

I had an ambition to keep knowledge about the state of the indentation and structure health of the current document updated as-you.type. However, it seems I will need to do some refactoring before I can do that. I wonder if for now it would be enough to check the documents when they are opened and when they are activated? What do other implementations do here?

The reason I would want to go back to very basic check points is that currently the status is updated in a bit of non-reciprocal way. So a change can lead to us detecting problems, but reverting the change (w/o undo) will not restore the knowledge about the status back, so you have a good document, but parinfer is disabled...

PEZ:
@svdo can you run Calva in dev mode and see if you can catch where the crashes happen that way?

I'm not seeing any crashes, but it could be because I reload the extension host every time I've made a change to Calva. 😄

PEZ:
I'm glad you are both happy with how this works. Parinfer is pure danger for me, so I can't use it for my daytime job. 😄

svdo:
> I had an ambition to keep knowledge about the state of the indentation and structure health of the current document updated as-you.type. However, it seems I will need to do some refactoring before I can do that. I wonder if for now it would be enough to check the documents when they are opened and when they are activated?

In strict mode, Calva tries to keep the structure OK. Mostly that works, but there are a few cases in which it doesn't:

* Cut a part of code with unbalanced parens
* Paste a part of code with unbalanced parens
* Selecting a part of code and then deleting it or typing over it

I think at least in those cases the check will be necessary...

> can you run Calva in dev mode and see if you can catch where the crashes happen that way?

I will try go gather more information. I also want to better understand _when_ it happens. I'll let you know when there is news.

PEZ:
VS Code's API does not let me know the reason for an edit other than if it is an Undo or Redo, so I don't think I can check for paste/cut, unless we implement separate commands for those. We do that with delete/backspace, so not saying we can't.

OTOH, if the structure breaks, Parinfer won't work, and most often Calva will indicate unmatched parens, so I am not too worried about those situations. It is when the indentation is off that things start to get crazy.

PEZ:
I think I have fixed so that the indentation/structure health check is always updated correctly now. Please see if you can break this one: https://16081-125431277-gh.circle-artifacts.com/0/tmp/artifacts/calva-2.0.226-pez-253-parinfer-58313be7.vsix

Also not sure how expensive it is, haven't measured. And my Macbook Pro M1 Max, *cough*, might hide performance problems.

svdo:
Will do @PEZ, bit after the weekend probably. 

Btw, are you aware the the indentation is messed up when pasting multiple lines containing more than one form? The first for ends up correctly, but the ones after are incorrect ("too far to the left").

PEZ:
Not aware! But I can reproduce it. We might need to implement a special paste command after all...

svdo:
> Not aware! But I can reproduce it. We might need to implement a special paste command after all...

I wonder how other implementations solve this, I'd expect this is not specific to Calva/vscode?

svdo:
I was also thinking about the TAB key today. On the one hand it is _very_ convenient to have a super fast way of fixing the indentation, but on the other hand... Frankly, the fact that the standard TAB key behavior is absent has confused me from the beginning. I wonder if this would be a good time to reconsider and start using the TAB key in a more "standard" way, which the parinfer-tabstop satisfies probably. Maybe what's currently under the TAB key could be the standard reformat/reindent key combo?

svdo:
NICE!!!! I just noticed that multicursor editing now works. Suppose you have this, with three cursors at the beginning of each line: (the `|` indicates the cursor position)

```clojure
|:foo
|:bar
|:baz
```

Then you type `(` and you get:

```clojure
(:foo)
(:bar)
(:baz)
```

That's a long standing feature request I think :) :)

(Would only be nice if the multiple-cursors remain after the operation, now it collapses back into a single one.)

PEZ:
I was also trying a bit with multi-cursors. I label it **not supported**, but am happy it works a bit at least.

At some point we should take a good look at the editing infra in Calva and figure out how to make it multi-cursor aware. I have a feeling it is both very tricky and not so very tricky.

PEZ:
>  I wonder if this would be a good time to reconsider and start using the TAB key in a more "standard" way

Haha, I do not agree. But, I am entertaining the idea to make it part of enabling Parinfer, that the tab key would take on the more standard behaviour, so maybe we are on the same page anyway. 😄

PEZ:
>> Not aware! But I can reproduce it. We might need to implement a special paste command after all...

> I wonder how other implementations solve this, I'd expect this is not specific to Calva/vscode?

I think it might be. I have not verified, but what I think happens is this:

Say you have a form like this (the cursor is denoted with `|`):

```clojure
(foo
  (bar
    |)
```

And you paste form like this:

```clojure
(baz)
(gaz)
```

The first thing we will have in the buffer is:

```clojure
(foo
  (bar
    (baz)
(gaz)|)
```

Then several things wants to happen. These two amongst them:

* Format-on-paste
* Infer parens

I suspect that that *Infer parens* might be happening before *Formar-on-paste*. Then we will have:

```clojure
(foo
  (bar
   (baz))
(gaz)|
```

And then *Format-on-paste* will make it:

```clojure
(foo
  (bar
    (baz))
(gaz)
```

Whereas if it happened in the order of *Format-on-paste* -> *Infer parens* we would first have:

```clojure
(foo
  (bar
    (baz)
    (gaz)|)
```

And then *Infer parens* would have nothing to change.

I don't know if the order these two happens in is predictable, but I know I have no say in it. *Format on paste* is something VS Code does whenever it fancies.

PEZ:
I have now confirmed that ^ theory about the paste problem. Hammock time now.

eneroth:
Formatting multi-arity functions turns this,

```clojure 
(defn hello
  ([world]
   world)
  ([dog]
   dog))
```

Into this, currently:
```clojure
(defn hello
  ([world
    world])
  ([dog
    dog]))
```

When the `#".*" [[:inner 0]]` indentation rule is used.

PEZ:
@eneroth, can you describe a bit what the indentation rule does, and what you expect should happen and such?

Is this a follow up bug on the cljfmt bug?

PEZ:
@svdo: Paste should work now.

The change for making this work are rather fundamental so now there is a risk a lot of things have broken. My testing says we're good, but there are a lot of combinations...

eneroth:
@PEZ I've made a [screen recording](https://eneroth.com/dHqN2tBkkiqtRLXRCK4RZBvs/dHqN2tBkkiqtRLXRCK4RZBvs.mov) to try to explain how Parinfer, and formatting with this particular rule interacts.

The expectation of the rule is that it behaves the same as with `cljfmt`. I kind of lack the vocabulary to express that in words, but the screen recording above shows what `cljfmt` will do with it.

PEZ:
@eneroth , thanks, I'll check it out. Can you file it as a separate issue? I see how it is related to this PR, so please reference it, but I'd like to have it tracked separately, because it is a rather daunting task to fix it. Not sure if those discussions have been conducted here or in slack, but anyway, not a small task.

PEZ:
Here's the latest build. Fixing paste issues and splitting up the toolbar item into two separate items, one for toggling parinfer on/off and one for dealing with structural health checks. https://16136-125431277-gh.circle-artifacts.com/0/tmp/artifacts/calva-2.0.226-pez-253-parinfer-27186129.vsix

PEZ:
Last build has some weird behaviour. I've observed:

Typing `js` here `(|console.log)` => `(s|jconsole.log)`

Pasting soething like `(foo)` in something like this:
```clojure
(defn foo []
  |
  (bar))
```

Pastes `(foo)` which then ”runs away” to the right (I think with ever new indentation added...).

A bit too bad for me to be using it with my day work, so then it doesn't get tested as much as I had hoped...

PEZ:
Worse. I can't reproduce any of those right now...

svdo:
Installing...

svdo:
Not sure how to give you more info on this, but it's now twice that I had the vscode "hang up" on me. First it says the window is not responding and asks me what to do, and then (after waiting some more) it crashes with code 5. I had it just now after installing the latest build, but also with the build from this morning (89c146cf I believe).

Screenshot 2021-11-22 at 13 55 33
Screenshot 2021-11-22 at 13 56 42

PEZ:
@svdo You had this hang/crash before I made the **paste** changes, right?

PEZ:
@svdo I'm thinking I will add some debug messaging to the builds and then maybe we can get more clues.

svdo:
> @svdo You had this hang/crash before I made the **paste** changes, right?

Not sure, I think it was with commit 89c146c that I first experienced this.

By the way, after your "rather fundamental" changes I am experiencing a lot more glitches. The cursor makes weird jumps, brackets are no longer balanced (eg after cutting the last line of a hashmap), sometimes parinfer seems to be "one character behind" (I type something with indentation on a new line, and the closing bracket of the line above is moved after the _second_ keystroke instead of the _first_, but this is inconsistent). Do you recognize any of this?

PEZ:
I recognize similar things. Not running so much with parinfer on, but even with it disabled I am seeing problems. Figuring a bit about how to isolate the changes better to not affect non-parinfer use so much...

See if this one behaves better, please: https://16146-125431277-gh.circle-artifacts.com/0/tmp/artifacts/calva-2.0.226-pez-253-parinfer-723ff372.vsix I've allowed one glitch to happen in the hope that we will see fewer of the other glitches.

PEZ:
@svdo The VSIX I just linked does some debug logging. Hoping that it could help us with the crash issue you are seeing. Might of course help with other issues as well. I think I see clues to that thing you see with

> the closing bracket of the line above is moved after the _second_ keystroke instead of the _first_

The promises seem to be resolved in weird order.

svdo:
Installing... Where would I find the debug logging if the vscode window crashes? Is it stored on disk somewhere?

PEZ:
Not stored on disk. AFAIK. Didn't think of that...

PEZ:
I think something goes wrong with the paren inference after an **Undo**. Nopt sure about the repro yet, but anyway.

PEZ:
This one behaves better, judging from both my testing and from the debug log. Still not perfect, and ymmv.

https://16151-125431277-gh.circle-artifacts.com/0/tmp/artifacts/calva-2.0.226-pez-253-parinfer-7db12c50.vsix

src/doc-mirror/index.ts
```diff
@@ -175,64 +178,72 @@ export class MirroredDocument implements EditableDocument {
...
+                const performFormatAsYouType = formatAsYouTypeOn && event.reason != vscode.TextDocumentChangeReason.Undo && !mirroredDoc.rangeFormatted;
+                const performInferParens = parinferOn && event.reason != vscode.TextDocumentChangeReason.Undo && !mirroredDoc.parensInferred;
+                let performHealthCheck = !performFormatAsYouType;
+                        if (event.document === vscode.window.activeTextEditor?.document) {
+                    if (performFormatAsYouType) {
+                        console.count(`processChanges edits applied .then performFormatAsYouType: ${event.contentChanges[0].text}`)
+                        if (event.contentChanges.length === 1 && event.contentChanges[0].text.match(/[\[\](){}]/) && !cursor.withinString()) {
+                            const change = event.contentChanges[0];
+                            const start = event.document.offsetAt(change.range.start);
+                            console.count(`changeRange: ${[start, event.document.offsetAt(change.range.end)]}`)
```
bpringe, line 209: TIL about `console.count` 😄

PEZ:
With some wonderful help from a Pitch colleague I have today fully reverted on the change I made to control the order of changes using a promise chain and won in actual control by it. With some luck this is the most stable build so far:

https://16177-125431277-gh.circle-artifacts.com/0/tmp/artifacts/calva-2.0.226-pez-253-parinfer-39380a88.vsix

It includes a change that disables both the aggressive formatting and parinfer for people who do not opt in on experimenting with Parinfer. This should make it pretty safe to again try a launch on the marketplace that includes the experiment. Please help in testing both with Parinfer ON and OFF, so that we can find any quirks we don't want to unleash on our poor users. ❤️ 🙏

svdo:
Still testing @PEZ .

Regarding the crashes of the extension host: I think they are because of another vscode extension. I recently installed the "live sass compiler', this one seems to misbehave. So you can forget about that, sorry for the confusion.

PEZ:
Thanks @svdo! I've experienced some memory issues on my machine, and worried a bit that it is some variant of the crashes you have seen. I don't have any indications that it is to do with these builds, but anyway. I googled it a bit and seems to be something with MacOS Monterey that isn't tuned yet. Let's hope we don't have a memory leak introduced!

New build:

https://16192-125431277-gh.circle-artifacts.com/0/tmp/artifacts/calva-2.0.226-pez-253-parinfer-75db5c1c.vsix

I'm figuring if we are were we can launch the experiment. There are glitches still, these being the ones I know about:

1. Type fast when ”pushing” forms in front of the cursor forward, and inference sneaks in.
2. Some Paredit operations gets undone by parinfer.

I know how to fix 2. In this build I have fixed it for slurp forward, but not for barf forward.

No real idea how to fix 1, other than pretty big rewrites. Thinking that a wider experiment will tell us how big the problem is in real use.

PEZ:
I now found a rather significant issue:

```clojure
(foo| (bar)
      (baz))
```

<kbd>backspace</kbd>

**Expected**:

```clojure
(fo| (bar)
     (baz))
```

**Actual**
```clojure
(fo| (bar
      (baz)))
```

PEZ:
I consulted an experienced Parinfer user about ^that^ issue and he found it unsurprising and expected. He's an indent mode user mostly, but anyway, maybe not a stopper. The workaround here is for the user to disable Parinfer, edit the function name and then enable Parinfer again.

As I see it, Parinfer gets extra dangerous when it does things like this, but since we would launch this as an experiment I think I am fine with releasing with this flaw/bug. It should be noted that there are variants on this issue that are a bit worse. I had a situation where I spent quite some time to try to ”heal” the code after Parinfer had stomped on it today. Still, it is an experiment, and we can try guide the users with good documentation.

I do have an idea for how to fix this, but while I am still at a stage where I am ”taming” the basic mechanisms, I don't think it is the best idea to run away on this tangent.

svdo:
I have used this today quite a bit. The recent builds have become quite usable. There are still some glitches now and then, and they are not consistent. It's like some internal state gets borked and the later is ok again. The most annoying thing that I have seen today is that incorrectly-indented namespace forms were broken by putting the parens wrong instead of disabling parindent. Especially since that happened while I did not edit that part of the file. 

As an experiment that can be disabled I think this is good enough to release, but some more iterations are still needed I guess. I'm super grateful for all the work you're putting in this @PEZ!

oakmac:
> I consulted an experienced Parinfer user about ^that^ issue and he found it unsurprising and expected. He's an indent mode user mostly, but anyway, maybe not a stopper.

[The example above](https://github.com/BetterThanTomorrow/calva/pull/1381#issuecomment-980044946) is correct behavior for Indent Mode.

For Smart Mode, moving the cursor above or below the current line plays a role in determining behavior.

Examples from the [demo implementation](http://shaunlebron.github.io/parinfer/demo) below:

### Indent Mode

Nov-26-2021 13-04-32

### Smart Mode

parinfer-smartmode

> As I see it, Parinfer gets extra dangerous when it does things like this

I think this is mostly a matter of expectation. As a general rule, Parinfer users expect their text operations to perform "as is" against the raw text, and then the Parinfer algorithm generates the structure after their operation based on indentation (this is the core idea behind Indent Mode). Smart Mode is an enhancement to this and uses cursor + text change information to try and predict what the user's structural intention was via their text change. Smart Mode is not always 100% perfect in guessing what structure the user intended, but both Indent and Smart Modes honor the "text change first, structure follows" approach.

By contrast, Paredit philosophy is "apply an exact structural change, then the text will follow".

PEZ:
Yes, it is about expectations. Which is why I think it is dangerous. Especially for beginners. Let's hope we can make this implementation a bit safer than it currently is. =)

PEZ:
There is currently an issue with activating files having broken structure or bad indentation. So if you have three tabs, a fine file in one of them, one with broken structure and one with bad indents you should be able to observe the issue by watching the structure/indentation indicator while:

0. Have the fine file active.
1. Activate the file with broken structure -> the indicator flashes the error icon, then settles for checkmark
2. Activate the file with bad indentation -> the indicator flashes the warning icon, then settles for the checkmark

This makes editing of the file start out with the assumption that Parinfer operations are safe (for some definition of safe), when in fact it might be unsafe.

@svdo, could this explain the spurious errors you have experienced? If you are running with the alert setting on, it wouldn't explain, since then you would have been warned.

I don't know when this started to happen and when debugging it I ”see” it happen, but the root cause is still a mystery. Once we find it, this will be fixed. I'm treating it as a blocker for the experiment.

svdo:
Yes it might, I will try to confirm Monday, but I indeed have the warning pop-ups disabled.

PEZ:
I've fixed the indicator bug now. It was introduced when I added som updates of it for good measure, one of them was on VS Code onOpenDocument events, which I now realized is triggered for documents that are not visible to the user. Among those a document that is named `<path-to-visible-doc>.git`, which is not a Clojure document and was indicated as ”structure + indents OK”.

I've now stopped updating the status bar on these events, and also indicate non-clojure documents parinfer readiness with a dash instead of the checkmark.

I doubt this was the glitches you experienced, @svdo, because this was only the indicator, the actual health check that stops bracket inference is not made there. I've now added some more logging and maybe we can find the error you've seen that way.

Here is the latest build: https://16207-125431277-gh.circle-artifacts.com/0/tmp/artifacts/calva-2.0.226-pez-253-parinfer-b6933663.vsix

I won't work on this for some hours now. Family day here.

svdo:
Enjoy your weekend! I won't be able to try before Monday anyway :)

PEZ:
I found that the paste fix was making the formatting wrong in some cases. I've reverted that and we are back to when pasting multiple forms results in bad bracket inference.

PEZ:
I have a new take on the paste problem now. Latest build: https://16302-125431277-gh.circle-artifacts.com/0/tmp/artifacts/calva-2.0.227-pez-253-parinfer-30ef222c.vsix

PEZ:
I'm going to merge this now, btw. I think the quirks we have are safely guarded behind the settings. Maintaining a separate branch for this long stresses me up.

svdo:
Yes I know that feeling, good decision I think.

PEZ:
🎉 🤞

PR #1402, opened 2021-11-29
Extract error message handling into functions

marcomorain:
I want to re-use some of this infrastructure when adding Test Explorer
support, so I'd like to extract this code out into smaller, composable
functions. This will allow me to re-use them (and test them).

Changes:

- Extract the code for adding summaries together into `totalSummary`.
- Extract the summary message generation into `summaryMessage.
- Replace uses of _ with native TypeScript looping constructs.

Ping @pez, @bpringe

### DIFF
--- a/src/extension-test/unit/test-runner-test.ts
+++ b/src/extension-test/unit/test-runner-test.ts
@@ -1,0 +1,43 @@
+import * as expect from 'expect';
+import * as cider from '../../nrepl/cider';
+
+describe('test result processing', () => {
+    it('shows a summary', () => {
+
+        expect(cider.summaryMessage({
+            ns: 0, var: 0, test: 0, pass: 0, fail: 0, error: 0
+        })).toBe("No tests found. 😱, ns: 0, vars: 0");
+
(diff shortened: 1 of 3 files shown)


bpringe:
Thanks!

> Replace uses of _ with native TypeScript looping constructs

Can you give an example?

Btw, I'd prefer if we leave the PR template checklist on PRs. It helps us (me, at least), to remember to check things too. 😄

marcomorain:
> Btw, I'd prefer if we leave the PR template checklist on PRs. It helps us (me, at least), to remember to check things too. 😄

Will do 👍

marcomorain:
> Can you give an example?

The calls to 

```TypeScript
_.each(diagnostics, (errors, fileName) => {
```

became

```
for (const fileName in diagnostics) {
```

And similar changes for `_.each` being replaced with `for`.

bpringe:
Oh I see. Yeah I'm not sure that the lodash implementations for those operations provide much benefit.

PEZ:
Thanks!

Issue #1328, opened 2021-10-10
Run All Tests command doesn't run tests in .cljc file with reader conditional in ns


andriikovalov:
Steps to reproduce:

1. Create a new lein app `lein new app test-app`
2. Rename `core_test.clj` in to `core_test.cljc`
3. In `core_test.cljc` change `[clojure.test :refer :all]` to `#?(:clj [clojure.test :refer :all])`
4. Open project directory in VS Code, jack-in
5. Invoke Run All Tests command, expected: failing test in REPL, actual: "No tests found"

bpringe:
Do any of the other test commands work in this situation? I wonder if cider-nrepl (which is used for running tests), only checks `.clj` files.

andriikovalov:
Yes, I tried "Run tests for current namespace" and "Run current test", and they both work (I see output with test results in REPL).

Here are some other things I noticed with "Run all tests":

- `.cljc` file without reader conditionals: works
- `.cljc` with a reader conditional inside a test: works
- `.cljc` with a reader conditional in `ns :require` (as in the description): no tests found

bpringe:
Interesting, thanks for the added details.

bpringe:
So, Calva is running the deprecated `test-all` op of cider-nrepl to run all tests.

We can try using the not deprecated `test-var-query` op to run all tests and see if that fixes this issue. I noticed that running the tests for the namespace works, which uses `test-var-query`.

@marcomorain Since you've been working on test-related stuff, would you like to tackle this when you have time? If not, I can do it eventually.

marcomorain:
I'm taking a quick look now.

marcomorain:
👍 PR open which fixes this.

PR #1424, opened 2021-12-07
Use test-var-query for all test commands

marcomorain:
## What has Changed?

Remove use of deprecated Cider commands, in favour of the supported
test-var-query. This fixes an issue with cljc conditionals impacting the
ability to run tests.

Fixes: #1328

## My Calva PR Checklist

I have:

Ping @pez, @bpringe

### DIFF
--- a/src/nrepl/index.ts
+++ b/src/nrepl/index.ts
@@ -447,7 +447,13 @@ export class NReplSession {
         return new Promise<cider.TestResults>((resolve, reject) => {
             let id = this.client.nextId;
             this.messageHandlers[id] = resultHandler(resolve, reject);
-            this.client.write({ op: "test-all", id, session: this.sessionId, "load?": true });
+            this.client.write({
+                op: "test-var-query", id, session: this.sessionId, "var-query": {
+                    "ns-query": {
+                        'test?': true
+                    }
+                }
+            });
         })
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Changes to Calva.
 ## [Unreleased]
 - Fix: [Babashka Jack-In REPL doesn't show eval errors](https://github.com/BetterThanTomorrow/calva/issues/1413)
 - [Inform about conflict with the Clojure extension](https://github.com/BetterThanTomorrow/calva/issues/1427)
+- Fix: [Run All Tests command doesn't run tests in .cljc file with reader conditional in ns](https://github.com/BetterThanTomorrow/calva/issues/1328).
 
 
 ## [2.0.228] - 2021-12-02


bpringe:
Seems to work in my simple test 😃 . I'm not sure if you're done with the changes yet, but if so, once the changelog is updated and vsix is tested I think we can merge.

marcomorain:
This should be good merge @bpringe 👍

src/nrepl/index.ts
```diff
@@ -447,7 +447,13 @@ export class NReplSession {
         return new Promise<cider.TestResults>((resolve, reject) => {
             let id = this.client.nextId;
             this.messageHandlers[id] = resultHandler(resolve, reject);
-            this.client.write({ op: "test-all", id, session: this.sessionId, "load?": true });
+            this.client.write({
```
marcomorain, line 450: Hrm, actually this might be a behaviour change - since we are no longer passing `load?` true. Let me confirm...

marcomorain:
I'm not sure about how to move forward with this – this is a behaviour change - the `testAll` command currently loads all namespace, but this PR changes that behaviour.

PEZ:
Is it that the new op does not accept the `load?` option?

marcomorain:
Yup, that’s exactly it.

PEZ:
I say, go for this. Your activities around this whole test story is improving things majorly and some behavior changes are inevitable. We can't afford to hang on to deprecated functionality. Maybe consider adding something to calva.io about that loading is a user responsibility?

bpringe:
I'm not sure this actually changes behavior. For example, in version 2.0.228 of Calva,I have the following test code:

```clojure
(ns test-lein.core-test
  (:require [clojure.test :as test :refer [deftest is]]))

(deftest test-1
  (is true))

(deftest test-2
  (is true))
```

and I do the following:

1. Run all tests, see they all pass
2. Change `true` to `false` in `test-2`
3. Run all tests again

Whether I save the file or not at step 2, all tests continue to pass, even though I changed `test-2` to make it fail. Furthermore, even if I run `Refresh All Namespaces` or `Refresh Changed Namespaces` then run all tests again, the change _still_ isn't picked up.

That last issue may be a problem with the refresh commands. I don't normally use them, but they don't seem to work as I expect in this case.

My overall point is that currently it doesn't seem that changes are loaded before tests are run anyway, at least not in this lein project I'm testing with. I have to run the command to load the test file in order for changes in tests to be picked up. I thought that I've observed otherwise in the past though 🤔 . I tested also with a deps.edn project and observed the same behavior.

marcomorain:
Sounds like we should merge away then?

src/nrepl/index.ts
bpringe: There doesn't seem to be a change in behavior, as described further in the main PR thread.

PR #1443, opened 2021-12-21
Fix link

dr0i:
This PR is directed at the `published` branch since it just fixes a link in the documentation.

### DIFF
--- a/docs/site/debugger.md
+++ b/docs/site/debugger.md
@@ -35,7 +35,7 @@ Calva comes with a powerful expression-based debugger, inspired by [Cider](https
 
 ## Dependencies
 
-The debugger itself relies pretty heavily on [cider-nrepl](https://github.com/clojure-emacs/cider-nrepl), as do other parts of Calva. This library is loaded as a dependency when you use Calva Jack-in. If you are not using Calva Jack-in, you can add these dependencies in your project definition or user profile. See the [Calva Jack-in guide](/jack-in-guide) for more information.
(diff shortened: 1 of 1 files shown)


dr0i:
No clue why this could fail a test-ts-unit.

dr0i:
I close this issue for the links are properly working when building the site at https://calva.io/debugger/ .

PEZ:
Hello! Thanks for caring about the documentation!

Did the links work properly in the local/dev site when you had fixed them? If so I think we should merge your fixes, because the current ones do not work on github/when browsing the Calva repo.

The ts-unit tests are a bit flaky. There is a generative test that sometimes produces failing tests. I haven't investigated so I don't yet know if it is the generator that is doing something wrong or that it actually is finding a bug. Anyway, your fix has nothing to do with it.

dr0i:
> Did the links work properly in the local/dev site when you had fixed them? If so I think we should merge your fixes, because the current ones do not work on github/when browsing the Calva repo.

Yes, after the fix they work at github, see https://github.com/dr0i/calva/blob/fixLink/docs/site/debugger.md which references https://github.com/dr0i/calva/blob/fixLink/docs/site/jack-in-guide.md and https://github.com/dr0i/calva/blob/fixLink/docs/site/getting-started.md.

PEZ:
Sorry for being unclear. I did check that your links worked on github and that the current links don't. That's why I reopned. 😄 

However, right now I can't remember if that also means it works on the docs site. The way to test it is to build the docs locally.

dr0i:
Makes sense. Sorry for being a bit dumb :) With your hint, I did `mkdocs serve` in the `fixLink` branch and visited http://127.0.0.1:8000/debugger/ - and the links are there good, too ! So the mkdocs seem to strip the `md` suffix of relative links in the markdown files. So indeed this PR should work for both: github and the generated docs.

PEZ:
That's awesome. Thanks again!

Issue #1442, opened 2021-12-20
Add custom commands from libraries


Cyrik:
Add custom snippets through code or a special file in a lib, similar to the new clj-kondo auto-include config. 
This would make it easier for tooling libraries to provide commands for ease of use. Currently to use something like debux, scope capture, omni-trace or even zprint I have to require it during every repl session, in every namespace. 
If the tools could provide custom snippets this would make them a lot easier to use, especially for beginners.

Using omni-trace as an example you could have the following snippet to trace code in the repl:
```clojure
((requiring-resolve 'cyrik.omni-trace/run-traced) '$current-form)
```
By combing this with #1441 you could now also use libraries like debux or scope-capture to wrap existing code during a debug session.

I'm not set on a way to register those snippets. 
The most general solution would probably be a way to call a calva registration function through the running repl, probably by using a custom nrepl middleware function, but that's also a hard dependency on nrepl and makes it harder for library authors to do.
A more user-friendly way would be a file, let's say calva-commands.edn that has to exist in the jar. I'm not sure if calva is aware of all loaded jars or if they can be scanned, so this might be impossible.
The easiest solution would probably be a .calva/ folder in the project root where any library can create it's own .edn file or subfolder and those would be read automatically, like clj-kondo is doing now: [https://github.com/clj-kondo/clj-kondo/issues/1483](https://github.com/clj-kondo/clj-kondo/issues/1483)

PEZ:
Let me see if I understand the use case. A library can add snippets making it convenient for Calva users to use the library?

Cyrik:
Yes. It's mostly helpful for dev-time libraries, though.
I don't want to over-promote my own stuff, but it's the reason I thought of this so here goes. When I want to debug something I'm working on I have [omni-trace](https://github.com/Cyrik/omni-trace) in my deps, so then to use it I have to do:

```clojure
(require '[cyrik.omni-trace :as o])

(o/run-traced 'cyrik.omni-trace.testing-ns/run-machine)

(tap> (o/rooted-flamegraph 'cyrik.omni-trace.testing-ns/run-machine))
```

I have that in my personal snippets so it's easy to just hit a key and it runs, but all users would have to create that snippet themselves. 
The same thing applies to the something like
```clojure 
((requiring-resolve 'clojure.tools.deps.alpha.repl/add-libs) '$current-form)
```

[...]

PR #1459, opened 2022-01-09
Add custom commands from libraries

Cyrik:
## What has Changed?

- Load custom snippets from .calva/config.edn

Addressing: #1442

## My Calva PR Checklist

I have:

Ping @pez, @bpringe

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -4,6 +4,7 @@ import { ReplConnectSequence } from './nrepl/connectSequence';
 import { PrettyPrintingOptions } from './printer';
 import { parseEdn } from '../out/cljs-lib/cljs-lib';
 import * as state from './state';
+const { fromJS } = require('immutable');
 
 const REPL_FILE_EXT = 'calva-repl';
 const KEYBINDINGS_ENABLED_CONFIG_KEY = 'calva.keybindingsEnabled';
@@ -30,12 +31,26 @@ const documentSelector = [
     return name.replace(/^[\s,:]*/, "").replace(/[\s,:]*$/, "")
 }
 
-async function readEdnConfig() {    
+async function readEdnWorkspaceConfig() {
     try {
         const file = vscode.Uri.file(state.resolvePath('.calva/config.edn'));
         const data = await vscode.workspace.fs.readFile(file);
-        const parsed = parseEdn(new TextDecoder("utf-8").decode(data));
-        state.setProjectConfig(parsed);
+        return addEdnConfig(new TextDecoder("utf-8").decode(data));
+    } catch (error) {
+        return error;
+    }
+}
+
+async function addEdnConfig(data:string) {    
+    try {
+        const parsed = parseEdn(data);
+        const old = state.getProjectConfig();
+        if (old && old.customREPLCommandSnippets) {
+            state.setProjectConfig({customREPLCommandSnippets: old.customREPLCommandSnippets.concat(parsed?.customREPLCommandSnippets ?? [])});
+        } else {
+            state.setProjectConfig({customREPLCommandSnippets: (parsed?.customREPLCommandSnippets ?? [])});
+        }
+        
     } catch (error) {
         return error;
     }
@@ -43,7 +58,7 @@ async function readEdnConfig() {
 var watcher = vscode.workspace.createFileSystemWatcher("**/.calva/**/config.edn", false, false, false);
 
 watcher.onDidChange(() => {
-    readEdnConfig(); 
+    readEdnWorkspaceConfig(); 
 });
 
 // TODO find a way to validate the configs
@@ -85,7 +100,8 @@ function getConfig() {
 }
 
 export {
-    readEdnConfig,
+    readEdnWorkspaceConfig,
+    addEdnConfig,
     REPL_FILE_EXT,
     KEYBINDINGS_ENABLED_CONFIG_KEY,
     KEYBINDINGS_ENABLED_CONTEXT_KEY,
--- a/src/config.ts
+++ b/src/config.ts
@@ -2,6 +2,8 @@ import * as vscode from 'vscode';
 import { customREPLCommandSnippet } from './evaluate';
 import { ReplConnectSequence } from './nrepl/connectSequence';
 import { PrettyPrintingOptions } from './printer';
+import { parseEdn } from '../out/cljs-lib/cljs-lib';
+import * as state from './state';
 
 const REPL_FILE_EXT = 'calva-repl';
 const KEYBINDINGS_ENABLED_CONFIG_KEY = 'calva.keybindingsEnabled';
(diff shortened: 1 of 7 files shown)


Cyrik:
This is just the most basic underpinning to start with.
Not sure if it makes sense to address the actual issue on top of this, since having config.edn on its own would already be useful.

There is still a lot missing since it currently only uses `:customREPLCommandSnippets` from inside the config.edn file. We should probably discuss how many of the other calva options should be supported here.
I'm also thinking about having the snippets as actual sci code, instead of strings, inside the edn file, but I'm not sure that's possible.

Cyrik:
I've realized that importing `.calva/config.edn` from the jars makes no sense, so I'll follow clj-kondos example and will import files in "calva.exports/" as described in clj-kondos docs: https://github.com/clj-kondo/clj-kondo/blob/master/doc/config.md#exporting.
For now, I'll import the config.edn directly in "calva.exports/config.edn", we can discuss if it makes sense to have subfolders based on org/lib-name

PEZ:
I think it makes sense to make it about custom command snippets only. So `custom-commands.edn`, maybe?

And, yes, a big win here would be if could write the clojure code as clojure code here, and not as strings.

Will try to understand how clj-kondo does this ”from libraries” thing.

src/config.ts
```diff
@@ -49,7 +71,7 @@ function getConfig() {
         asyncOutputDestination: configOptions.get("sendAsyncOutputTo") as string,
         customREPLCommandSnippets: configOptions.get("customREPLCommandSnippets", []),
         customREPLCommandSnippetsGlobal: configOptions.inspect("customREPLCommandSnippets").globalValue as customREPLCommandSnippet[],
-        customREPLCommandSnippetsWorkspace: configOptions.inspect("customREPLCommandSnippets").workspaceValue as customREPLCommandSnippet[],
+        customREPLCommandSnippetsWorkspace: commands,
```
PEZ, line 93: We might want to consider existing configs.

Cyrik:
I think custom commands are just the first step, we could also add custom tooltips, inline eval results and other things. But I'm open to having multiple edn files.
clj-kondo doesn't actually have the code in the edn file, it kind of links to the code in a round-about way and has the code in .clj files in subdirectories. We could do the same thing.

src/config.ts
Cyrik: Those are merged earlier, but the merge logic could be extended. Currently, I'm just throwing in everything. We might merge on snippet names.

src/config.ts
```diff
@@ -49,7 +71,7 @@ function getConfig() {
         asyncOutputDestination: configOptions.get("sendAsyncOutputTo") as string,
         customREPLCommandSnippets: configOptions.get("customREPLCommandSnippets", []),
         customREPLCommandSnippetsGlobal: configOptions.inspect("customREPLCommandSnippets").globalValue as customREPLCommandSnippet[],
```
PEZ, line 92: Are you thinking that the config format should stay so similar that we can use both regular VS Code settings and EDN files for this? If so maybe we can keep both global and workspace config in settings and merge the EDN configs. (See my other comment about existing configs.)

PEZ:
> we could also add custom tooltips, inline eval results and other things

What is a custom tooltip, and what is an inline eval result in this context?

Also: is the context ”things that libraries can add” or something else?

Generally I think Calva settings should be regular VS Code settings. What warrants EDN config is stuff like Clojure code, and when we share config with some library we use (like we do with `cljfmt`).

Cyrik:
Ah sorry for the out of context feature drop. We had discussed custom tooltips in slack some time ago.

> What is a custom tooltip, and what is an inline eval result in this context?
>Also: is the context ”things that libraries can add” or something else?

Those would be things libraries and users can add.
Custom tooltips would be similar to custom snippets. So if my library has a some special thing it can display in a hover tooltip at some var/symbol/etc it can define a custom tooltip and that would be displayed in addition to the normal things.
Custom inline eval results are the same thing, just for the inline eval display.
Other things might be access to the new nrepl side-loader so that you add middleware without the user having to paste anything into their deps.edn or project.clj file.

All of these are things that should make it easier for tools authors to integrate better into calva without users having to do more than include in deps. Ideally at some point tools.deps.alpha will be done so that we can even add those tools into a running repl without any configs at all.

src/config.ts
Cyrik: Yes, after playing with it a bit more that makes sense to me.
The local .calva/config.edn would have the same calva settings as .vscode/settings.json. The something.jar/calva.export/config.edn would have dependency-specific settings like custom snippets.

PEZ:
Thanks for clarifications. TBH, I didn't really follow the first time. Also, good for other readers to have context.

I realize tooltips, inline results, and nrepl middleware are mere examples, but anyway:

* tooltips: Is there a use case where the user would need to add some custom things the hover for some symbol? I'm striving to keep Calva's settings down to a minimum, and this seems like an outlier need. As for library developers, I guess if they can add behaviour to Calva this way, it might make sense, but otherwise Calva already shows the doc string.
* inline eval results: I still don't get it. 😄 Can you give an example?
* nrepl middleware: Yesterday I saw something about an nrepl.edn file. Is that related, you think? (I know nothing about this file, not even which tool is reading it, you just made me recall.)

Cyrik:
### Tooltips
I'm not sure if there's any use-case for users. It would probably be mostly a thing for libraries and for testing those without having to compile. 
Docs make sense to me since docstrings are not very flexible. If we had custom tooltips the "ClojureDocs example" functionality could have been written as a library instead of adding it directly into Calva. Now when some alternative to ClojureDocs shows up we have to also bake this into Calva and provide an option to switch. As a library with custom tooltips users can choose which one they use.
My use-case here is displaying traced calls in the tooltips as if you had eval'd all the calls inline. So if you do `(run-traced (some-func 5))` I can show you the result in the `some-func` tooltip and do the same for anything it might have called.
Another library that would benefit from this is spec/malli where you can show the spec for a function in the tooltip of the function.

### Inline eval results
Same tracing example, but some other tool might want to display something else there. A linter might show a warning or something.

### nrepl middleware
Yes, currently you can put something like `{:middleware [my/middleware]}` into `.nrepl.edn` and it will install that middleware for you. This works fine, but it's one more step the user has to do to use my library and a lot of beginners will be confused by the interaction. It would be great if the user didn't have to think about it. Then even the "Getting started repl" could include tooling libraries to showcase them. Although this example probably should be built into nrepl instead of calva, but it was the first thing that came to mind.

### General

The main idea is to provide extension points into Calva so that tools authors can extend it without having to touch the code and without users having to copy-paste configuration to their setting.json file. An added benefit would be that some of the Calva functionality could be pulled out into (optional) libraries written in Clojure.
A nice side benefit would be that fixing cljs <-> typescript compatibility wouldn't be as important anymore, although that might be a bit too much to hope for.

Cyrik:
@PEZ So after our talk I've tried to make it possible to have snippets that quoted cljs, but didn't get that working yet. Also adding logging to the config code is problematic since it gets called before the log is initialized :)
So I think for now I'm happy for you to merge this and we can build the other things on top. Even adding "snippets as quoted code" should be backward compatible since we can check the type of the returned value.

If there's anything more you want me to do here let me know, otherwise I'll start looking at tooltips for a new PR. Hope they aren't as bad as we think 😃

src/connector.ts
```diff
@@ -17,6 +17,34 @@ import * as calvaDebug from './debugger/calva-debug';
 import { setStateValue, getStateValue } from '../out/cljs-lib/cljs-lib';
 import * as replSession from './nrepl/repl-session';
 import * as clojureDocs from './clojuredocs';
+import * as jszip from 'jszip';
+import { addEdnConfig } from './config';
+
+async function readJarContent(uri: string) {
+    console.log(uri);
```
bpringe: There's a `getJarContents` function in utilities.ts. I'm not sure if you knew that. Maybe it could be reused here?

PEZ: Let us know if you think this can be achieved, @Cyrik. Other than that I think we are good to go.

Cyrik: ah great, I didn't see that. I'll work it in later today, thanks for the feedback

bpringe:
I think some documentation should be added to this PR. @Cyrik After that's done and the VSIX is tested, I think it'll be good to go.

Cyrik:
Sounds good @bpringe.
I'd extend the custom commands doc and create a new one for the integration conf, ok? 
Do you want the integration conf doc under etcetera or features? I'd probably throw it under etcetera for now and move it over once the feature is a little bigger.

PEZ:
etcetera sounds good for now.

I think we can merge this w/o the docs, and you add them in a new PR. Thinking that we should not hold up on your ability to build on the new functionality.

PEZ:
There. 😄 

Many thanks for contributing again, @Cyrik!

PR #1466, opened 2022-01-12
Fix currentTopLevelFunction function for metadata skipping

### LINKED ISSUE #1463: Snippet in custom command doesn't work with function metadata
`$top-level-defined-symbol` does not return the function symbol when there is metadata declared on the function.

Sample custom command definition:
```
    {
      "name": "print function name",
      "key": "printFunc",
      "snippet": "(println '$top-level-defined-symbol)"
    },
```

# Failing Scenario (when metadata is present)

```
(defn ^{:foo "bar"} baz
  []
  true)
```

Ouput for custom command when cursor is in `baz` function body:
```
(println ')
; Syntax error reading source at (REPL:1293:12).
; Unmatched delimiter: )
```

# Functional Scenario (when no metadata is present)
```
(defn baz
  []
  true)
```

Ouput for custom command when cursor is in `baz` function body:
```
(println 'baz)
baz
nil
```

Calva version: v2.0.233

sashton:
## What has Changed?

~- This PR updates the forwardSexp function for the case where there exists metadata between two `id` nodes and `skipMetadata==true`. Previously, this function would move forward through whitespace and metadata, finally landing on an `id` node. However, in the `switch (token.type) ... case 'id'` code there would be one final call to `this.next()`, thereby moving the cursor past the correct s-expr.~
~- Now, the `switch (token.type) ... case 'id'` scenario has been updated to differentiate between:~
~- the initial iteration of the while loop, where the cursor is still on the original `id` node, and it _is_ correct to move past this original `id` node~
~- any subsequent iteration of the loop, where it is not correct to move the cursor past the current `id` node~

Update:
- The original solution of updating fowardSexp caused unintended consequences with other cursor-motion scenarios. I have reverted my changes there, and instead made the change in `currentTopLevelFunction`. 
- This change updates `currentTopLevelFunction` by checking whether the cursor is on metadata, and, if so, moves forward via `fowardSexp` until a non-metadata node is found. It is a simpler, more specific solution than the first attempt.

Fixes #1463

## My Calva PR Checklist

I have:

     ~- [ ] If I am fixing just part of the issue, I have just referenced it w/o any of the "fixes” keywords.~

Ping @pez, @bpringe

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Changes to Calva.
 ## [Unreleased]
 
 - [Improve LSP startup feedback on status bar](https://github.com/BetterThanTomorrow/calva/pull/1454)
+- Fix: [Snippet in custom command doesn't work with function metadata] (https://github.com/BetterThanTomorrow/calva/issues/1463)
 
 ## [2.0.233] - 2022-01-07
 - [Add experimental support for Test Explorer](https://github.com/BetterThanTomorrow/calva/issues/953)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ Changes to Calva.
 ## [Unreleased]
 
 - [Improve LSP startup feedback on status bar](https://github.com/BetterThanTomorrow/calva/pull/1454)
+- Fix: [Snippet in custom command doesn't work with function metadata] (https://github.com/BetterThanTomorrow/calva/issues/1463)
+
 
 ## [2.0.233] - 2022-01-07
 - [Add experimental support for Test Explorer](https://github.com/BetterThanTomorrow/calva/issues/953)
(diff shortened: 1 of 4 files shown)


PEZ:
Thanks!

Some tests are failing now. Do you know how to run the tests locally? Let me know otherwise and I'll show you.

sashton:
@PEZ I see the failures. I'm investigating. I may need to solve this a different way. From reading the failed tests, I think I had an incorrect mental model of how forwardSexp was designed to operate over metadata and reader tags.

PEZ:
Those parts are quite messy. We should clean it up some day.

sashton:
@PEZ I've reverted the original change and instead made the change in `currentTopLevelFunction`. All tests now pass, including the new one I added to cover the scenario in the bug report.

PEZ:
I'm having a look and a think about this. You have probably found the right place to do this, but at the same time this function is also very central to a lot of things. And the current tests do pass. I just need to experiment and think a bit. Please don't hesitate to bump this if I go silent for more than a few days.

PEZ:
Right so I did go silent... But now have had a look and a think. This seems like the right way to solve the problem so I am going to go ahead and merge. Many thanks for helping in improving Calva!

PR #1473, opened 2022-01-14
Update Jack-in Dependency Versions docs

MatElGran:
## What has Changed?

This removes specific default versions for Jack-in dependencies from documentation so that it is not necessary to maintain them anymore. These are replaced with an example configuration and a pointer to the location of the current ones in the code. 

Happy to discuss those choices and make any change deemed necessary. 

PS: thanks for this project, it's awesome and has really helped me discovering Clojure

## My Calva PR Checklist

I have:

- [ ] ~Directed this pull request at the `dev` branch. (Or have specific reasons to target some other branch.)~
- [ ] ~Made sure I have changed the PR base branch, so that it is not `published`. (Sorry for the nagging.)~
- [ ] ~Updated the `[Unreleased]` entry in `CHANGELOG.md`, linking the issue(s) that the PR is addressing.~
- [ ] ~Figured if **anything** about the fix warrants tests on Mac/Linux/Windows/Remote/Whatever, and either tested it there if so, or mentioned it in the PR.~
- [ ] ~Added to or updated docs in this branch, if appropriate~
- [ ] ~Tested the VSIX built from the PR (so, after you've submitted the PR). You'll find the artifacts by clicking _Show all checks_ in the CI section of the PR page, and then _Details_ on the `ci/circleci: build` test.~
     - [ ] ~Tested the particular change~
     - [ ] ~Figured if the change might have some side effects and tested those as well.~
     - [ ] ~Smoke tested the extension as such.~
- [ ] ~Referenced the issue I am fixing/addressing _in a commit message for the pull request_.~
     - [ ] ~If I am fixing the issue, I have used [GitHub's fixes/closes syntax](https://help.github.com/en/articles/closing-issues-using-keywords)~
     - [ ] ~If I am fixing just part of the issue, I have just referenced it w/o any of the "fixes” keywords.~
- [ ] ~Created the issue I am fixing/addressing, if it was not present.~

Ping @pez, @bpringe

### DIFF
--- a/docs/site/customizing.md
+++ b/docs/site/customizing.md
@@ -92,13 +92,25 @@ Jack-in and Connect are very customizable through [Custom Connect Sequences](con
 
 ## Jack-in Dependency Versions
 
-The versions of the dependencies [Calva Jack-in](jack-in-guide.md) injects in order for the REPL session to support IDE features are configurable via the VS Code settings `calva.jackInDependencyVersions`. At the time of this writing the default versions are:
+[Calva Jack-in](jack-in-guide.md) injects the following dependencies in order for the REPL session to support IDE features
 
-Dependency | Version | Description
----------- | ------- | -----------
-[nrepl](https://github.com/nrepl/nrepl) | 0.8.3 | nREPL is the wonderful piece of software that gives Calva a structured and extensible connection to the REPL in your Clojure and ClojureScript projects.
-[cider-nrepl](https://github.com/clojure-emacs/cider-nrepl) | 0.25.8 | cider-nrepl is middleware that extends the nREPL connection with all sorts of nice stuff that Calva uses to give you a delightful IDE experience.
-[cider/piggieback](https://github.com/nrepl/piggieback) | 0.5.2 | Piggieback is used to create nREPL sessions in ClojureScript projects. (Not with [shadow-cljs](http://shadow-cljs.org) projects though, which provides its own middleware for this.)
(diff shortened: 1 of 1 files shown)


docs/site/customizing.md
```diff
@@ -92,13 +92,25 @@ Jack-in and Connect are very customizable through [Custom Connect Sequences](con
...
+"calva.jackInDependencyVersions": {
+    "nrepl": "0.9.0",
+    "cider-nrepl": "0.27.4",
+    "cider-piggieback": "0.5.3"
+}
+```
+
+Current default versions can be found in [package.json](https://github.com/BetterThanTomorrow/calva/blob/published/package.json) at the following path: 
+
+`$['contributes']['configuration']['properties']['calva.jackInDependencyVersions']['defaults']`
```
PEZ, line 113: Maybe we don't even need these?

PEZ: Maybe we don't need the example and where to find the defaults?

PEZ:
Thanks for contributing!

docs/site/customizing.md
MatElGran: I _think_ I agree, given that this seems to be some advanced configuration and someone who needs this would probably be able to correctly guess all this by themselves. I'll make the changes and let you decide.

PEZ: It's a tricky balance, but generally I try to keep the docs free from too much VS Code guidance. Thinking that kind of information is the responsibility of other sources. I often break this ”rule” though. 😄 In this case, I'm banking on that people know to just open settings and look at the defaults there.

PR #1479, opened 2022-01-18
[LSP] Exclude repl output from lsp

### LINKED ISSUE #1250: Suspicion: when Calva's REPL window has a lot of text, it is slowing down clojure-lsp
As VS Code is sending all Clojure files to clojure-lsp, and the REPL/output window is a Clojure file, it is analysed and could potentially be causing performance problems affecting the whole project.

ericdallo:
## What has Changed?

- Never send repl content to LSP analysis 

Fixes #1250

## My Calva PR Checklist

I have:

- [X] Tests

Ping @pez, @bpringe

### DIFF
--- a/src/lsp/main.ts
+++ b/src/lsp/main.ts
@@ -13,7 +13,7 @@ import * as path from 'path';
 import * as state from '../state';
 import { provideHover } from '../providers/hover';
 import { provideSignatureHelp } from '../providers/signature';
-import * as cider from '../nrepl/cider'
+import { isResultsDoc } from '../results-output/results-doc';
 
 const LSP_CLIENT_KEY = 'lspClient';
 const RESOLVE_MACRO_AS_COMMAND = 'resolve-macro-as';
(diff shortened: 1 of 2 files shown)


src/lsp/main.ts
```diff
@@ -64,6 +64,24 @@ function createClient(clojureLspPath: string): LanguageClient {
...
+                return next(document);
+            },
+            didSave: async (document, next) => {
+                if (document.uri.path.endsWith(config.REPL_FILE_EXT)) {
+                    return;
+                }
+                return next(document);
+            },
+            didChange: async (change, next) => {
+                if (change.document.uri.path.endsWith(config.REPL_FILE_EXT)) {
```
PEZ: We should probably use `isResultsDoc(doc)` from `results-doc.ts` here.

I noticed that there was prior art in this file doing it this way, so we should fix that one too.

A reason to check the extension is if we consider that Calva would have more than one output/REPL window, but if that happens we will need to revisit a lot of code, niot just this one.

PEZ:
Thanks!

PEZ:
There seems to be some unrelated commits in the PR branch. Can you clean that up? (I'm not sure this is necessary, but it is generally easier to deal with ”clean” PRs :smile:.)

PEZ:
@bpringe Are we relying on clojure-lsp in a way that we need to consider before merging this? I seem to recall something about the debugger using clojure-lsp analysis.

src/lsp/main.ts
ericdallo: Done, I found no way to fix the one that was already being used on this file, we would need to convert a vscode.Uri to a vscode.TextDocument

ericdallo:
Thanks @PEZ ! Done, LMK if any more issues

src/lsp/main.ts
PEZ: I see. Might be the reason it was not done like that. 😄

bpringe:
@PEZ The debugging code just uses clojure-lsp for decorations (for the instrumented functions). It doesn't seem to even add decorations in the repl window currently, and I'm unsure if that's intentional, but either way, I don't think it matters if this disallows that.

bpringe:
Thanks @ericdallo! Let's merge. 🚀

bpringe:
By the way, I removed that newline between changes in the unreleased section of the changelog (on dev). It might have caused an issue with our publish script.

PR #1510, opened 2022-01-27
Update output queueing and batching logic

### LINKED ISSUE #1189: REPL stuck and requires window reload occasionally
Hello, thanks again for a wonderful tool.

For a while I've had the intermittent (but basically daily) issue that when I try to jack-in to a Clojure REPL Calva will hang at this step in the repl output window:

```
; This is the Calva evaluation results output window.
; TIPS: The keyboard shortcut `ctrl+alt+c o` shows and focuses this window
;   when connected to a REPL session.
; Please see https://calva.io/output/ for more info.
; Happy coding! ♥️
```

The session never hooks up and I can never interact. The status bar shows `nREPL cljc/clj` as normal, and attempting to jack in again does nothing to fix it.

To resolve this I have to reload the VSCode window

I can see this stacktrace pertaining to Calva in the VSCode devtools but I don't know if it's related:

```
 ERR Cannot read property 'fsPath' of null: TypeError: Cannot read property 'fsPath' of null
    at outputFileDir (/Users/charlie.briggs/.vscode/extensions/betterthantomorrow.calva-2.0.198/src/results-output/results-doc.ts:45:54)
    at DOC_URI (/Users/charlie.briggs/.vscode/extensions/betterthantomorrow.calva-2.0.198/src/results-output/results-doc.ts:50:32)
    at append (/Users/charlie.briggs/.vscode/extensions/betterthantomorrow.calva-2.0.198/src/results-output/results-doc.ts:254:43)
    at Object.appendPrompt (/Users/charlie.briggs/.vscode/extensions/betterthantomorrow.calva-2.0.198/src/results-output/results-doc.ts:362:5)
    at /Users/charlie.briggs/.vscode/extensions/betterthantomorrow.calva-2.0.198/src/extension.ts:51:26
    at Generator.next (<anonymous>)
    at o (/Users/charlie.briggs/.vscode/extensions/betterthantomorrow.calva-2.0.198/out/extension.js:2:982375)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

### Versions

Mac OS: Big Sur 11.2.3
VSCode:
```
Version: 1.56.2
Commit: 054a9295330880ed74ceaedda236253b4f39a335
Date: 2021-05-12T17:44:30.902Z (2 wks ago)
Electron: 12.0.4
Chrome: 89.0.4389.114
Node.js: 14.16.0
V8: 8.9.255.24-electron.0
OS: Darwin x64 20.3.0
```
Calva: 
```
Name: Calva: Clojure & ClojureScript Interactive Programming
Id: betterthantomorrow.calva
Description: Integrated REPL, formatter, Paredit, and more. Powered by cider-nrepl and clojure-lsp.
Version: 2.0.198
Publisher: Better Than Tomorrow
VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=betterthantomorrow.calva
```

### LINKED ISSUE #1509: Output window gets stuck constantly appending to queue
I've identified a long running issue where the output window gets into a bad state and no longer updates with new output. What I found is that the current logic which uses an `isUpdating` flag encounters some edge condition that causes `isUpdating` to never get cleared, causing the queue to grow indefinitely and require a vscode restart.

stefan-toubia:
## What has Changed?

This refactors the logic for queueing and writing to the output window. #1509 points out that certain conditions can cause the output window to get stuck in `isEditing` mode and cause the queue to grow indefinitely.

The main change in this PR is to queue **all** appended output to the new `EditQueue` to try and eliminate some of these edge conditions causing the output to break.

This is still a draft because I still have some questions around the previous batching logic and `OnAppendedCallback`

Fixes #1189
Fixes #1509

## My Calva PR Checklist

I have:

- [x] Tests

Ping @pez, @bpringe

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Changes to Calva.
 
 ## [Unreleased]
 
+- Fix: [Results doc gets in a bad state and does not update](https://github.com/BetterThanTomorrow/calva/issues/1509)
+
 ## [2.0.259] - 2022-03-26
 - [Add setting for enabling LiveShare support](https://github.com/BetterThanTomorrow/calva/issues/1629)
 
--- a/src/results-output/results-doc.ts
+++ b/src/results-output/results-doc.ts
@@ -27,7 +27,7 @@ const START_GREETINGS =
 ; Happy coding! ♥️';
 
 export const CLJ_CONNECT_GREETINGS =
-  '; TIPS: \n\
+  '; TIPS:\n\
 ;   - You can edit the contents here. Use it as a REPL if you like.\n\
 ;   - `alt+enter` evaluates the current top level form.\n\
 ;   - `ctrl+enter` evaluates the current form.\n\
@@ -295,7 +295,14 @@ async function writeToResultsDoc({ text, onAppended }: ResultsBufferEntry): Prom
   const doc = await vscode.workspace.openTextDocument(docUri);
   const insertPosition = doc.positionAt(Infinity);
   const edit = new vscode.WorkspaceEdit();
-  edit.insert(docUri, insertPosition, `${lastLineIsEmpty(doc) ? '' : '\n'}${text}`);
+  let editText = text;
+  if (!lastLineIsEmpty(doc)) {
+    editText = '\n' + editText;
+  }
+  if (!editText.endsWith('\n')) {
+    editText += '\n';
+  }
+  edit.insert(docUri, insertPosition, editText);
   if (!((await vscode.workspace.applyEdit(edit)) && (await doc.save()))) {
     return;
   }
--- a/src/results-output/results-doc.ts
+++ b/src/results-output/results-doc.ts
@@ -283,7 +283,7 @@ function registerResultDocSubscriptions() {
   let currentResultDocs = visibleResultsEditors();
   const subOpen = vscode.window.onDidChangeVisibleTextEditors((editors) => {
     const current = editors.filter((editor) => isResultsDoc(editor.document));
-    const opened = current.filter((editor) => currentResultDocs.indexOf(editor) < 0);
+    const opened = current.filter((editor) => currentResultDocs.includes(editor));
     currentResultDocs = current;
     opened.forEach(handleResultDocEditorDidOpen);
   });
--- a/src/results-output/results-doc.ts
+++ b/src/results-output/results-doc.ts
@@ -147,6 +147,8 @@ export async function initResultsDoc(): Promise<vscode.TextDocument> {
   await vscode.workspace.applyEdit(edit);
   void resultsDoc.save();
 
+  registerResultDocSubscriptions();
+
   // For some reason onDidChangeTextEditorViewColumn won't fire
   state.extensionContext.subscriptions.push(
     vscode.window.onDidChangeActiveTextEditor((event) => {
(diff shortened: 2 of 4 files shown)


src/results-output/results-doc.ts
```diff
@@ -236,85 +236,147 @@ export function appendCurrentTopLevelForm() {
...
-    };
+    }
+
+    const onAppendedCallbacks = items
+        .map((item) => item.onAppended)
+        .filter(identity);
+
+    // Note that these onAppended callbacks are batched and called after all
+    // current batched entries have been written.
+    onAppendedCallbacks.forEach((onAppended) =>
```
stefan-toubia: It wasn't immediately apparent to me what the intent was behind the previous batching behavior, so I held off on addressing it until I could clarify. Previously the batches had extra partitioning logic that ensured at most one `onAppendedCallback` existed in a batch. I could see two reasons for this -

1. It's important to call `onAppended` immediately after that specific text was written and before any next entries are written
2. It's important to provide the exact `Location` for that text entry

It looks like these callbacks are (exclusively?) used by `evaluate` to add selection decoration, so I'm under the impression that it's really (2) that we want here. Regardless, I think both cases do not require special batching logic, instead I can rearrange the contents of `handleNextOutputQueueBatch` to call each queued `onAppended` item immediately after writing it instead of after writing the whole batch.

src/results-output/results-doc.ts
```diff
@@ -236,85 +236,147 @@ export function appendCurrentTopLevelForm() {
...
+  setTimeout(async () => {
+    while (editQueue.size() > 0) {
+      try {
+        await handleNextOutputQueueBatch();
+      } catch (err) {
+        console.error("Error writing output: ", err);
+      }
+    }
+  }, 100);
+});
```
stefan-toubia: There's some room for improvement on how the queue gets flushed but this gets the job done.

src/results-output/results-doc.ts
PEZ: My brain is fried right now so I will refrain from having any input on this right now. And I think @bpringe is the one who knows most about this. anyway. Also, iirc, @brdloush has been thinking (and also doing) things about this whole batching in general.

PEZ:
Thanks for doing this!

stefan-toubia:
I'm just happy to have finally identified the cause of this issue, it's been an almost daily issue for me for the past year or so!

src/results-output/results-doc.ts
bpringe: Yeah, I believe the batching was added by @brdloush, so let's see what he has to say about this.

stefan-toubia: No response from @brdloush, I'd like to squash this bug. I can rework this to maintain the existing batching logic. The important part of this PR is removing `applyingEdit`. WDYT @PEZ and @bpringe, do you see any blockers?

src/results-output/results-doc.ts
```diff
@@ -236,85 +236,143 @@ export function appendCurrentTopLevelForm() {
...
-
-   If something must be done after a particular edit, use the onResultAppended callback. */
+
+type EditQueueItem = {
+  line: string ;
+  onAppended?: OnAppendedCallback;
+};
+type QueuedCallback = (item: EditQueueItem) => void;
+export class EditQueue {
+  #queue: EditQueueItem[] = [];
```
bpringe, line 256: Can this be done more functionally instead? I like to avoid classes, especially ones that mutate internal state, as they seem to add complexity with no real benefit (at least in cases like this). If you disagree, I'd be happy to hear you out and discuss it further.

bpringe (changes requested):

src/results-output/results-doc.ts
bpringe: I'd like to test this and review this more thoroughly when I have more time, unless @PEZ can get to it before me.

bpringe:
I also want to say thanks for spending time on this!

src/results-output/results-doc.ts
```diff
@@ -236,85 +236,143 @@ export function appendCurrentTopLevelForm() {
...
-   apply edits one after another without issues.
-
-   If something must be done after a particular edit, use the onResultAppended callback. */
+
+type EditQueueItem = {
+  line: string ;
+  onAppended?: OnAppendedCallback;
+};
+type QueuedCallback = (item: EditQueueItem) => void;
+export class EditQueue {
```
bpringe: Would you mind refactoring this code to be more functional (similar to how it was, but with the new functionality you've applied)? I'd prefer we not use a class for this. I think classes tend to increase complexity and increase the likelihood of bugs and/or hard to maintain code.

I could help with this if you want.

bpringe (changes requested):

src/results-output/results-doc.ts
bpringe: Actually, maybe hold off on that. I left a comment on the issue about a repro case: https://github.com/BetterThanTomorrow/calva/issues/1509#issuecomment-1052792339. I want to make sure we understand the cause better before we make changes.

stefan-toubia:
I'll be away for the next week or so, but I'll revisit this when I'm back.

stefan-toubia:
@PEZ @bpringe 

I revisited this and cleaned it up for review. I kept the original batching logic for the most part, but cleaned up the `append` function by breaking it down into separate parts. The most important change here that will fix the issue of the output getting into a bad state is the addition of the `try/catch` to protect the `outputPending` flag.

I'm also pretty sure I was able to figure out how to reproduce this issue, described here #1641.

PEZ (commented):
Thanks for doing this! ❤️ 

The changes look good to me. I have tested the branch and things behave as I expect.

Please update the PR protocol adding a changelog entry, etc, and we're good to go, imo. 🙏

stefan-toubia:
@PEZ I've updated the changelog and tested the vsix that I built locally, but I was unsure how to use the artifact from `ci/circleci: build`. This downloaded as a zip file that extracts into a directory, I'm not sure how to convert that into a vsix file.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ Changes to Calva.
 
 ## [Unreleased]
 
+- Fix: [Results doc gets in a bad state and does not update](https://github.com/BetterThanTomorrow/calva/issues/1509)
+
```
PEZ: Because reasons, we can't have an empty row between the header and the list

```suggestion
## [Unreleased]
- Fix: [Results doc gets in a bad state and does not update](https://github.com/BetterThanTomorrow/calva/issues/1509)

```

PEZ (approved):
Thanks again. This is good to be merged now. I can take care of the changelog thing after merge.

PEZ:
@stefan-toubia we should update the PR template about that testing the VSIX is a bit optional and more necessary when changing in the CLJS code. As for the `.zip` file ending. Sometimes it downloads like that, I have no clue why. But a `.vsix` file is really just a zip-file with some special content in it. so changing the file extension is enough.

PEZ:
🎉 ❤️ 🙏

bpringe (commented):
Good stuff!

PR #1517, opened 2022-02-05
Fix running ns tests may print error

### LINKED ISSUE #1516: TypeError when "Run Tests for Current Namespace" in non "-test" namespace
I got error:
```
TypeError: Cannot read property 'test' of undefined
```
when trying to **"Run Tests for Current Namespace"** in a non `-test` namespaces. [This command's official docs](https://calva.io/test-runner/#test-commands) state that:

> Runs the tests for the current namespace. If not a -test namespace, tests for the current namespace plus its corresponding <current-namespace>-test namespace will be run.

so it seems that non `-test` namespaces are indeed supported so I would expect the command worked instead of giving an error, especially if individual tests can be run using command **"Run Current Test"**.

For instance, in the attached image you can see 2 test files `core-test.clj` & `test-core.clj` and the output in REPL for each file  when first each test is run individually using **"Run Current Test"** and then all tests in namespace are run using **"Run Tests for Current Namespace"**. As you can see, everything works fine for `core-test` namespace but we got the error in this issue when running "Run Tests for Current Namespace" in namespace `test-core`.

I understand that the convention is to write tests inside `-test` namespaces but based on the official docs and the fact that the tests in the namespace can be successfully run by evaluating `(clojure.test/run-tests 'example.test-core)` in the REPL,  I do believe this is an issue.

Screen Shot 2022-02-03 at 13 37 50

bpringe:
## What has Changed?

- Filters test results to get only actual results. (Filters out nrepl messages saying a namespace was not found when tests were run for a non-existing namespace.)

Fixes #1516 

## My Calva PR Checklist

I have:

- [x] Tests

Ping @pez, @bpringe

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Changes to Calva.
 
 ## [Unreleased]
 - Fix: ['Add to history' not working on eval in repl](https://github.com/BetterThanTomorrow/calva/issues/1594)
+- Fix: [TypeError when "Run Tests for Current Namespace" in non "-test" namespace](https://github.com/BetterThanTomorrow/calva/issues/1516)
 
 ## [2.0.237] - 2022-01-30
 - Fix: ['Show Previous REPL History Entry' command not working on v.2.0.236](https://github.com/BetterThanTomorrow/calva/issues/1594)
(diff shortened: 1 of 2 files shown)


marcomorain:
thanks for taking care of this 🙏

bpringe:
No problem!

PR #1562, opened 2022-02-24
Fix bug with clojure lsp format config race condition

### LINKED ISSUE #1561: Format config from clojure-lsp broken
The format on save works when I save a file, but apparently, the format doesn't correspond to the `clojure-lsp` config. Also, an error is showing when a open my project.

versions:
```
calva v2.0.245
clojure-lsp version used: 2022.02.23-12.12.12
clj-kondo version used: 2022.02.09
```
settings:
```
"calva.fmt.configPath": "CLOJURE-LSP",
"calva.clojureLspPath": "/usr/local/bin/clojure-lsp",
"editor.formatOnSave": true,
```
error:
```
Fetching formatting settings from clojure-lsp failed. Check that you are running a version of clojure-lsp that provides "cljfmt-raw" in serverInfo.
```

Running the Clojure-lsp Server Info I see the `cljfmt-raw` field.
```
  "cljfmt-raw": "{:indents {assoc-in-some [[:block 1] [:block 1] [:block 1] [:block 1]], facts [[:block 1] [:block 1] [:block 1] [:block 1]], with-responses [[:block 0] [:block 0] [:block 0] [:block 0]], provided [[:inner 0] [:inner 0] [:inner 0] [:inner 0]], as-delegate [[:block 2] [:block 2] [:block 2] [:block 2]], lambda-expr [[:inner 0]], |fn [[:inner 0]], assoc-if [[:block 1] [:block 1] [:block 1] [:block 1]], defglobal-native [[:inner 0]], profile [[:block 2] [:block 2] [:block 2] [:block 2]], with-scopes [[:block 1] [:block 1] [:block 1] [:block 1]], consume! [[:block 0] [:block 0] [:block 0] [:block 0]], defint [[:block 0] [:block 0] [:block 0] [:block 0]], as-delegate-with-lift [[:block 2] [:block 2] [:block 2] [:block 2]], defglobal-library [[:inner 0]], |if-not [[:block 1]], |when [[:block 1]], let-entities [[:block 2] [:block 2] [:block 2] [:block 2]], |with-state [[:block 1]], constraint-fn [[:block 2] [:block 2] [:block 2] [:block 2]], as-customer [[:block 1] [:block 1] [:block 1] [:block 1]], for-all [[:block 1] [:block 1] [:block 1] [:block 1]], flow [[:block 1] [:block 1] [:block 1] [:block 1]], against-background [[:inner 0] [:inner 0] [:inner 0] [:inner 0]], quick-check [[:block 1] [:block 1] [:block 1] [:block 1]], against-backgroud [[:block 0]], logged-call [[:block 1] [:block 1] [:block 1] [:block 1]], run-recipe [[:inner 0] [:inner 0] [:inner 0] [:inner 0]], log-messages [[:block 0] [:block 0] [:block 0] [:block 0]], profiled [[:block 1] [:block 1] [:block 1] [:block 1]], |def-expr [[:inner 0]], as-admin [[:block 1] [:block 1] [:block 1] [:block 1]], try-type [[:block 0] [:block 0] [:block 0] [:block 0]], |if [[:block 1]], future-facts [[:block 1] [:block 1] [:block 1] [:block 1]], some-with-open [[:block 1] [:block 1] [:block 1] [:block 1]], not-join [[:block 1]], fact [[:block 1] [:block 1] [:block 1] [:block 1]], expand-routes [[:inner 1] [:inner 1] [:inner 1] [:inner 1]], flow-with-defaults [[:block 1] [:block 1] [:block 1]], if-expr [[:block 1]], |let [[:block 1]], idempotent-start [[:block 1] [:block 1] [:block 1] [:block 1]], idempotent-stop [[:block 1] [:block 1] [:block 1] [:block 1]], as-of [[:block 1] [:block 1] [:block 1] [:block 1]], with-fn-validation [[:block 0] [:block 0] [:block 0] [:block 0]], defuf [[:block 1] [:block 1] [:block 1] [:block 1]], rethrow [[:block 1] [:block 1] [:block 1] [:block 1]], flow-as-of [[:block 1] [:block 1] [:block 1] [:block 1]], assoc-some [[:block 1] [:block 1] [:block 1] [:block 1]], future-fact [[:block 1] [:block 1] [:block 1] [:block 1]], tabular [[:inner 0] [:inner 0] [:inner 0] [:inner 0]], as-customer-with-lift [[:block 1] [:block 1] [:block 1] [:block 1]], non-test-action [[:inner 0] [:inner 0] [:inner 0] [:inner 0]], data-fn [[:block 1] [:block 1] [:block 1] [:block 1]], let-expr [[:block 1]], assoc [[:block 1] [:block 1] [:block 1] [:block 1]], assoc-in-if [[:block 1] [:block 1] [:block 1] [:block 1]], flow-without-validation [[:block 1] [:block 1] [:block 1] [:block 1]], ignore-conflict [[:inner 0] [:inner 0] [:inner 0] [:inner 0]], with-cid [[:block 1] [:block 1] [:block 1] [:block 1]], let-flow [[:block 1] [:block 1] [:block 1] [:block 1]], embeds [[:block 0] [:block 0] [:block 0] [:block 0]], system-map [[:block 0] [:block 0] [:block 0] [:block 0]], defhandler [[:block 2] [:block 2] [:block 2] [:block 2]]}}"
 ```

### PR DESCRIPTION
**Pull request by PEZ**
## What has Changed?

The new feature with fetching cljfmt config from clojure-lsp was a bit too naïvely implemented. It worked when I tested it while developing, but of course the order in witch things initialize is not reliable like that...

I've now added a function in lsp/main.ts that lets's you await the lsp client before you query it. It's a bit brute, I think, but it does the job, afaikt. Might come in handy for other things.

I've also made the `startClient` function abort if the client fails to start. Less "Promise reject was not handled” messages in the console.

Can you have a look at the changes, @bpringe ? I'm always a bit confused around promises. 😄 

I'm a bit worried that formatting operations will now hang if the CLOJURE-LSP option is used and clojure-lsp takes time to initialize....

Also, @corasaurus-hex : I know you removed some async/await constructs because they weren't needed. Here I am introducing such things again. If you see an opportunity for me to learn a bit about this, please go ahead. 😄 

Fixes #1561

## My Calva PR Checklist

I have:

- [ ] Tests

Ping @pez, @bpringe

### DIFF
--- a/src/calva-fmt/src/config.ts
+++ b/src/calva-fmt/src/config.ts
@@ -36,15 +36,12 @@ async function readConfiguration(): Promise<{
     const workspaceConfig = vscode.workspace.getConfiguration('calva.fmt');
     const configPath: string = workspaceConfig.get('configPath');
     if (configPath === LSP_CONFIG_KEY && !lspFormatConfigFetched) {
-        lspFormatConfig = await lsp
-            .getCljFmtConfig()
-            .then((c) => {
-                lspFormatConfigFetched = true;
-                return c;
-            })
-            .catch((e) => {
-                lspFormatConfigFetched = true;
-            });
+      try {
+        lspFormatConfig = await lsp.getCljFmtConfig(); 
+        lspFormatConfigFetched = true;
+      } catch (_) {
+        lspFormatConfigFetched = true;
+      }
     }
     if (configPath === LSP_CONFIG_KEY && !lspFormatConfig) {
         vscode.window.showErrorMessage(
--- a/src/calva-fmt/src/config.ts
+++ b/src/calva-fmt/src/config.ts
@@ -36,12 +36,12 @@ async function readConfiguration(): Promise<{
     const workspaceConfig = vscode.workspace.getConfiguration('calva.fmt');
     const configPath: string = workspaceConfig.get('configPath');
     if (configPath === LSP_CONFIG_KEY && !lspFormatConfigFetched) {
-      try {
-        lspFormatConfig = await lsp.getCljFmtConfig(); 
-        lspFormatConfigFetched = true;
-      } catch (_) {
-        lspFormatConfigFetched = true;
-      }
+        try {
+            lspFormatConfig = await lsp.getCljFmtConfig();
+            lspFormatConfigFetched = true;
+        } catch (_) {
+            lspFormatConfigFetched = true;
+        }
     }
     if (configPath === LSP_CONFIG_KEY && !lspFormatConfig) {
         vscode.window.showErrorMessage(
--- a/src/calva-fmt/src/config.ts
+++ b/src/calva-fmt/src/config.ts
@@ -72,7 +72,7 @@ async function readConfiguration(): Promise<{
     }
 }
 
-export function getConfig() {
-    const config = readConfiguration();
+export async function getConfig() {
+    const config = await readConfiguration();
     return config;
 }
--- a/src/calva-fmt/src/config.ts
+++ b/src/calva-fmt/src/config.ts
@@ -1,7 +1,7 @@
 import * as vscode from 'vscode';
 import * as filesCache from '../../files-cache';
 import * as cljsLib from '../../../out/cljs-lib/cljs-lib.js';
-
+import * as lsp from '../../lsp/main';
 const defaultCljfmtContent =
     '\
 {:remove-surrounding-whitespace? true\n\
@@ -12,6 +12,7 @@ const defaultCljfmtContent =
 
 const LSP_CONFIG_KEY = 'CLOJURE-LSP';
 let lspFormatConfig: string;
+let lspFormatConfigFetched = false;
 
 function configuration(
     workspaceConfig: vscode.WorkspaceConfiguration,
@@ -26,9 +27,17 @@ function configuration(
     };
 }
 
-function readConfiguration() {
+async function readConfiguration(): Promise<{ 'format-as-you-type': boolean; 'keep-comment-forms-trail-paren-on-own-line?': boolean; 'cljfmt-options-string': string; 'cljfmt-options': object; }> {
     const workspaceConfig = vscode.workspace.getConfiguration('calva.fmt');
     const configPath: string = workspaceConfig.get('configPath');
+    if (configPath === LSP_CONFIG_KEY && !lspFormatConfigFetched) {
+        lspFormatConfig = await lsp.getCljFmtConfig().then(c => {
+            lspFormatConfigFetched = true;
+            return c;
+        }).catch(e => {
+            lspFormatConfigFetched = true;
+        });
+    }
     if (configPath === LSP_CONFIG_KEY && !lspFormatConfig) {
         vscode.window.showErrorMessage(
             'Fetching formatting settings from clojure-lsp failed. Check that you are running a version of clojure-lsp that provides "cljfmt-raw" in serverInfo.',
@@ -53,10 +62,6 @@ function readConfiguration() {
         );
         return configuration(workspaceConfig, defaultCljfmtContent);
     }
-}
-
-export function setLspFormatConfig(config: string) {
-    lspFormatConfig = config;
 }
 
 export function getConfig() {
--- a/src/calva-fmt/src/config.ts
+++ b/src/calva-fmt/src/config.ts
@@ -27,16 +27,24 @@ function configuration(
     };
 }
 
-async function readConfiguration(): Promise<{ 'format-as-you-type': boolean; 'keep-comment-forms-trail-paren-on-own-line?': boolean; 'cljfmt-options-string': string; 'cljfmt-options': object; }> {
+async function readConfiguration(): Promise<{
+    'format-as-you-type': boolean;
+    'keep-comment-forms-trail-paren-on-own-line?': boolean;
+    'cljfmt-options-string': string;
+    'cljfmt-options': object;
+}> {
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: src/calva-fmt/src/config.ts [hunk 1]
```diff
@@ -55,10 +72,6 @@ function readConfiguration() {
     }
 }
 
-export function setLspFormatConfig(config: string) {
-    lspFormatConfig = config;
-}
-
 export function getConfig() {
     const config = readConfiguration();
     return config;
```
isacjunior: Suggestion: As the `readConfiguration` return a promise, we need to await the config here, right?
```suggestion
export async function getConfig() {
    const config = await readConfiguration();
    return config;
```

PEZ: Yes, maybe that is better than sprinkling all those awaits around like I have done. Thanks!

src/calva-fmt/src/config.ts
```diff
@@ -26,9 +27,25 @@ function configuration(
...
+    if (configPath === LSP_CONFIG_KEY && !lspFormatConfigFetched) {
+        lspFormatConfig = await lsp
+            .getCljFmtConfig()
+            .then((c) => {
+                lspFormatConfigFetched = true;
+                return c;
+            })
+            .catch((e) => {
+                lspFormatConfigFetched = true;
+            });
```
isacjunior: **🎯  Suggestion:** Maybe a `try catch` here is easier to read.‏

```suggestion
      try {
        lspFormatConfig = await lsp.getCljFmtConfig(); 
        lspFormatConfigFetched = true;
      } catch (_) {
        lspFormatConfigFetched = true;
      }
```

PEZ: 🙏

src/lsp/main.ts
```diff
@@ -430,15 +430,20 @@ async function startClient(
     lspStatus.text = '$(rocket) clojure-lsp';
     lspStatus.show();
     client.start();
-    await onReadyPromise;
+    await onReadyPromise.catch((e) => {
+        console.error('clojure-lsp:', e);
+        setStateValue(LSP_CLIENT_KEY_ERROR, e);
+    });
```
corasaurus-hex, line 436: it's funny, before I worked on this project I didn't know you could mix `await` + `.catch`

src/lsp/main.ts
```diff
@@ -658,6 +663,33 @@ export async function getClojuredocs(
     }
 }
 
+// TODO: This feels a bit brute, what are other ways to wait for the client to initialize?
```
corasaurus-hex, line 666: This is fine as-is but we could pull in a package for this -- like https://github.com/devlato/async-wait-until -- to get something with behavior that's easier to wrap your head around.

corasaurus-hex:
@PEZ this all looks great, and eslint agrees! 😉

It's fine to add back in async/await when it's needed, and eslint/typescript will complain when it is/isn't needed, so you're good.

PEZ:
Thanks. Very good to know.

While working with this I introduced some bugs by not awaiting the promise at some places. Is there a way eslint can be made to help with that?

corasaurus-hex:
It already should be, I thought?

corasaurus-hex:
Whoops, it wasn't turned on, but the PR that just linked to this turns it on.

src/calva-fmt/src/config.ts
```diff
@@ -55,11 +69,7 @@ function readConfiguration() {
     }
 }
 
-export function setLspFormatConfig(config: string) {
-    lspFormatConfig = config;
-}
-
-export function getConfig() {
-    const config = readConfiguration();
+export async function getConfig() {
```
bpringe, line 72: This function is actually redundant. You can just export `readConfiguration` and await it where you call it, which is the same as awaiting a call to this function.

src/cljs-lib/src/calva/fmt/formatter.cljs
```diff
@@ -286,17 +286,19 @@
   [^js m]
   (-> m
       (parse-cljfmt-options-string)
+      (update-in [:config :cljfmt-options] merge-cljfmt)
       (format-text)))
 
 (defn format-text-at-range-bridge
   [^js m]
+  (def m m)
```
bpringe, line 294: Left one of these in 😄 .

src/lsp/main.ts
bpringe: I didn't know this either 😄 .

src/lsp/main.ts
bpringe: @PEZ If you return `onReadyPromise` from `startClient` you can just await `startClient` to know when the client has started and is ready, rather than using this `getClient` function. I could help with that if you want.

src/cljs-lib/src/calva/fmt/formatter.cljs
PEZ: Whoopsie! We should probably clean the ClojureScript code up to where it has zero warnings and check it in CI.

src/lsp/main.ts
PEZ: Oh, that would be less brute. I'll have a go at it and see where it brings me. Thanks!

bpringe: No problem! You may want to put the code that was previously after the await within `startClient` inside a .then.

So it would be like

```typescript
return onReadyPomise.then(_ => { ...do-stuff });
```

That should return the promise as well as run the code we want to run after the client starts.

bpringe: Also there may be a better way to do the above... but the above is a simple refactor.

PR #1568, opened 2022-03-01
Wip/ts null checks

corasaurus-hex:
## What has Changed?

Handled strict null checks for when we get the active text editor and for when we get the mirror doc.

## My Calva PR Checklist

I have:

- [x] Tests


PEZ:
Thanks for taking care of Calva quality!

I need some help to understand the implications of the changes. I see that `mustGetActiveEditor` throws. So what will happen at some call site when there is no active editor?

Is the plan to get rid of `util.getActiveEditor` and `mirror-doc.getDocument`? If so, maybe name them `legacyGet...`, or `dangerouslyGet...` or something and use the shorter name for the new and proper functions?

PEZ:
> I need some help to understand the implications of the changes. I see that mustGetActiveEditor throws. So what will happen at some call site when there is no active editor?

I now saw that you had mentioned this on slack and that it will just retain the current behaviour. That's cool. Baby steps.

corasaurus-hex:
> I need some help to understand the implications of the changes. I see that mustGetActiveEditor throws. So what will happen at some call site when there is no active editor?

`activeTextEditor` can be undefined and accessing a property on undefined throws an error. So if the calling code isn't checking if `activeTextEditor` is undefined then I'm calling `mustGetActiveTextEditor` instead. This way we get an error explaining that the issue is that we don't have an `activeTextEditor` instead of getting `TypeError: Cannot read properties of undefined`. The other upside to doing it this way is that it results in the function only returning a `TextEditor` and not `TextEditor | undefined` which makes subsequent uses of the `activeTextEditor` a lot less painful.

> Is the plan to get rid of util.getActiveEditor and mirror-doc.getDocument? If so, maybe name them legacyGet..., or dangerouslyGet... or something and use the shorter name for the new and proper functions?

I don't see a reason to get rid of them, they're still useful in a few places where the code is checking if it's undefined. If we eventually remove all calls to the functions that can return undefined then sure.

> I now saw that you had mentioned this on slack and that it will just retain the current behaviour. That's cool. Baby steps.

It's _mostly_ the same behavior. Let me give a few examples.

1. In [this section of the code](https://github.com/BetterThanTomorrow/calva/pull/1568/files#diff-e3d18b462c48dff67750adbda073c5eaedfece30f2f169a684bca4716cea3655L93-R95) it calls to get the mirrored document and then immediately calls `getCursorToken` on it, which would already cause error if the mirrored document was missing.  Throwing an error in `mustGetDocument` is basically the same behavior as before except it has a better error message now.
2. In [this section of the code](https://github.com/BetterThanTomorrow/calva/pull/1568/files#diff-91067271c6542b04f84527cb4d1fe37abc1d8b1451b2e1be5f13f353828b4b33L17-R41) I moved the fetching of the editor close to where it's first used so that if an error is thrown it's thrown near the spot it used to be thrown.
3. And in [sections of code like this](https://github.com/BetterThanTomorrow/calva/blob/c2ecb0d29c90231523857cfab1bef0c460bf76e6/src/evaluate.ts#L94-L141) things get a little more difficult to suss out what the correct behavior should be. I made the assumption that we wouldn't want it to evaluate code if there is no `activeTextEditor` (and that it wouldn't happen in practice, but we'll see about that). Similar choices were made [here](https://github.com/BetterThanTomorrow/calva/blob/c2ecb0d29c90231523857cfab1bef0c460bf76e6/src/extension-test/integration/suite/extension-test.ts#L88-L99), [here](https://github.com/BetterThanTomorrow/calva/blob/c2ecb0d29c90231523857cfab1bef0c460bf76e6/src/highlight/src/extension.ts#L268-L315), [here](https://github.com/BetterThanTomorrow/calva/blob/7d5851089ff2364a07d0b44b84962fbbdc8a2f7a/src/paredit/extension.ts#L368-L374), and maybe other places.

So it's _mostly_ the same behavior as before but it's not impossible that it changes the behavior in some important way that I'm not aware of.

corasaurus-hex:
Also, maintenance-wise, is it important to create an issue for these sorts of code quality improvements?

PEZ:
Thanks! And thanks for taking your time to explain the implications to me. It all sounds perfect and wonderful to me. As I've mentioned before we have quite a lot of strange errors where there is very hard for users and us to try figure out what has actually gone wrong.

> Also, maintenance-wise, is it important to create an issue for these sorts of code quality improvements?

The important part is the changelog entry. When a users experiences a regression it is very handy with good descriptions on what has changed when. And that changelog should ideally be linked to somewhere where you can read more about it. The best place is usually the problem statement, so the issue. In this case you can link to this PR instead, if you like.

bpringe:
> In [this section of the code](https://github.com/BetterThanTomorrow/calva/pull/1568/files#diff-e3d18b462c48dff67750adbda073c5eaedfece30f2f169a684bca4716cea3655L93-R95) it calls to get the mirrored document and then immediately calls getCursorToken on it, which would already cause error if the mirrored document was missing. Throwing an error in mustGetDocument is basically the same behavior as before except it has a better error message now.

This is great. I think I've seen errors about calling `getTokenCursor` of `undefined` in logs before, so this should give us more info to fix issues like this.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 Changes to Calva.
 
 ## [Unreleased]
+- [Began work on enabling strictNullChecks in the TypeScript config.](https://github.com/BetterThanTomorrow/calva/pull/1568)
```
bpringe:
```suggestion
- [Begin work on enabling strictNullChecks in the TypeScript config.](https://github.com/BetterThanTomorrow/calva/pull/1568)
```

Just want to stick to present tense like "Fix ...", "Add ..."

package.json
```diff
@@ -2603,7 +2603,8 @@
         "clean": "rimraf ./out && rimraf ./tsconfig.tsbuildinfo && rimraf ./cljs-out",
         "compile-cljs": "npx shadow-cljs compile :calva-lib :test",
         "prewatch": "npm i && npm run clean && npm run update-grammar && npm run compile-cljs",
-        "watch": "npx tsc -watch -p ./tsconfig.json",
+        "watch": "npx tsc --watch --project ./tsconfig.json",
+        "watch-with-strict-nulls": "npx tsc --watch --project ./tsconfig.json --strictNullChecks",
```
bpringe, line 2604: Might we want to just use `--strictNullChecks` on the normal watch? Or does that still throw many warnings/errors until we finish this work? When this work is finished, then we would add it to the normal watch?

bpringe (changes requested):

package.json
corasaurus-hex: It has an incredible number of errors still. This is barely scratching the surface. I just wanted to get this kicked off and then work towards the end goal in sensible chunks, merging in frequently.

corasaurus-hex: I checked and there are [734 errors for this branch when I enable strictNullChecks](https://github.com/BetterThanTomorrow/calva/files/8192013/errors.txt) 😬

corasaurus-hex: I checked and there are 679 [errors](https://github.com/BetterThanTomorrow/calva/files/8192014/errors.txt) for this branch when I enable strictNullChecks 😬

(revised this comment since I was on the wrong branch)

corasaurus-hex:
I'm pretty sure this is ready to merge at this point?

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 Changes to Calva.
 
 ## [Unreleased]
+- [Begin work on enabling strictNullChecks in the TypeScript config.](https://github.com/BetterThanTomorrow/calva/pull/1568)
```
```suggestion
- Maintenance: [Begin work on enabling strictNullChecks in the TypeScript config.](https://github.com/BetterThanTomorrow/calva/pull/1568)
```

As a way to communicate that the users should not expect any new behaviour.

PEZ (approved):
Thanks! LGTM.

I left a nitpick about the changelog.

I still think `muestGet` is not going to be as discoverable in the module API as something that begins with `get`. In Clojure I would solve it with some symbol slapped to the end of the symbol name. but here I am out of better ideas.

package.json
corasaurus-hex: I'm now down to 333 in another branch I'm working on 😎

CHANGELOG.md
corasaurus-hex: Oh, great idea, that'll be helpful ☺️

corasaurus-hex:
It could be `getDocumentOrError`? That would be a little more discoverable, I think? Or perhaps we make strict the default and make the function that can return undefined be something like `maybeGetDocument` or `tryToGetDocument` or something similar?

PEZ:
I like both suggestions. The latter is my favorite, I think.

PR #1581, opened 2022-03-06
Further work on TypeScript strictNullChecks

corasaurus-hex:
## What has Changed?

- Further work on enabling strictNullChecks for typescript.
- Renaming mustGetX -> getX and changing getX -> tryToGetX as @PEZ suggested https://github.com/BetterThanTomorrow/calva/pull/1568#issuecomment-1059921150

## My Calva PR Checklist

I have:

- [x] Tests

Ping @pez, @bpringe


corasaurus-hex:
I tried to keep the behavior the same. There are 242 remaining errors (down from 750-ish!). There is probably some more low-hanging fruit remaining if I sift through some more but very soon we'll need to start revamping pieces to support strictNullChecks.

bpringe:
I've started reviewing, but need to stop for now. I'll try to finish tomorrow.

PEZ:
Thanks for doing this, @corasaurus-hex ! I get dizzy reading through all the changes and can just imagine what an effort it must be providing.

I get dizzy reading through all the changes. It all looks very sane to me. I only see a few places where I start to worry about the current behaviour being kept. Not because I see any error, but because I can't reason about the changes clear enough to understand the implications from reading. This says much more about the current code than the changes, so there is nothing we can do about it in this PR.

I will run with this VSIX today for work. If that works as usual, then I will vote for merge. And we'll get @bpringe's review results in tomorrow, so hopefully we'll merge tomorrow! 🍾 (I like to celebrate in advance. The worst case scenario there is an extra celebration, right?).

PEZ:
I merged latest dev into this, so that my testing gets more real-world. 😄 I do hope I didn't merge away any of your changes, but it merged pretty cleanly at least.

src/providers/signature.ts
```diff
@@ -86,7 +93,7 @@ function getCurrentArgsRanges(document: TextDocument, idx: number): Range[] {
         (previousRangeIndex > 1 &&
             ['->', 'some->'].includes(previousFunction)) ||
         (previousRangeIndex > 1 &&
-            previousRangeIndex % 2 &&
+            previousRangeIndex % 2 !== 0 &&
```
bpringe, line 96: Was this check for the modulus result just missing before?

src/results-output/results-doc.ts
```diff
@@ -102,18 +103,17 @@ export function setSession(session: NReplSession, newNs: string): void {
     }
 }
 
-export function isResultsDoc(doc: vscode.TextDocument): boolean {
-    return doc && path.basename(doc.fileName) === RESULTS_DOC_NAME;
+export function isResultsDoc(doc?: vscode.TextDocument): boolean {
+    return !!doc && path.basename(doc.fileName) === RESULTS_DOC_NAME;
```
bpringe, line 107: What's the difference in using double negation here vs just `doc && ...`?

src/statusbar.ts
```diff
@@ -37,12 +37,25 @@ function colorValue(
...
+
+    const value =
+        workspaceFolderValue || workspaceValue || globalValue || defaultValue;
+
+    // Current behavior is to assert that this is a string even though it may
+    // not be. Maintaining current behavior for the moment but we should
+    // eventually do an assertion here or allow the function to return
+    // undefined.
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion, @typescript-eslint/no-unnecessary-type-assertion
+    return value!;
```
bpringe, line 58: Is this `!` after the variable intentional? If so, what does it do? If I try to evaluate something like that in the console directly (like `myVar!`), I get an exception that `!` is an unexpected token. I'm guessing this is just something that works here that I'm unaware of.

src/providers/signature.ts
corasaurus-hex: yep! it worked before because `0` is `false`-y. however, we're assigning the result of this condition to a `boolean` variable with the previous code if the modulus was `0` then the result of this condition would be `0` instead of a `boolean`

src/results-output/results-doc.ts
corasaurus-hex: with `doc && ...` if `doc` was `undefined` then this function would return `undefined`. the function type says it returns a `boolean`, though, and while we could make it `boolean | undefined` I prefer to keep the types more constrained and simpler if I can

corasaurus-hex:
@PEZ I checked out the merge, all seems well. thanks for doing that!

bpringe (approved):

src/results-output/results-doc.ts
bpringe: Ah, thanks!

bpringe:
TIL what the non-null assertion operator is 😄 . Anyway, let's merge!

bpringe:
> I will run with this VSIX today for work. If that works as usual, then I will vote for merge.

Whoops. I hope that merge was okay. 😅

PR #1582, opened 2022-03-06
def-hover

### LINKED ISSUE #1538: Show def values in tooltip
```clojure
(def a {:foo "bar"})
(def b (assoc a :newProp 2))
```

When writing code like that, it would be nice to hover over b and see the value it currently has. This will only work if connected to a repl and if that var is actually eval'd already, but better that's true a lot of the time.

I'd build this on top of #1471 just to also dogfood that feature.

What do you think @PEZ @bpringe?

Cyrik:
## What has Changed?

- fixed hover markdown newlines (the problem was mentioned on slack)
- added "def var" hover as a default to the config

Closes #1538 

## My Calva PR Checklist

I have:

- [x] Tests

Ping @pez, @bpringe

### DIFF
--- a/src/extension-test/unit/util/cursor-get-text-test.ts
+++ b/src/extension-test/unit/util/cursor-get-text-test.ts
@@ -3,64 +3,60 @@ import { docFromTextNotation } from '../common/text-notation';
 import * as getText from '../../../util/cursor-get-text';
 
 describe('get text', () => {
-    // describe('getTopLevelFunction', () => {
-    //     it('Finds top level function at top', () => {
-    //         const a = docFromTextNotation(
-    //             '(foo bar)•(deftest a-test•  (baz |gaz))'
-    //         );
-    //         const b = docFromTextNotation(
-    //             '(foo bar)•(deftest |a-test|•  (baz gaz))'
-    //         );
-    //         const range: [number, number] = [b.selectionLeft, b.selectionRight];
-    //         expect(getText.currentTopLevelFunction(a)).toEqual([
-    //             range,
-    //             b.model.getText(...range),
-    //         ]);
-    //     });
+    describe('getTopLevelFunction', () => {
+        it('Finds top level function at top', () => {
+            const a = docFromTextNotation(
+                '(foo bar)•(deftest a-test•  (baz |gaz))'
+            );
+            const b = docFromTextNotation(
+                '(foo bar)•(deftest |a-test|•  (baz gaz))'
+            );
+            const range: [number, number] = [b.selectionLeft, b.selectionRight];
+            expect(
+                getText.currentTopLevelFunction(a, a.selection.active)
+            ).toEqual([range, b.model.getText(...range)]);
+        });
 
-    //     it('Finds top level function when nested', () => {
-    //         const a = docFromTextNotation(
-    //             '(foo bar)•(with-test•  (deftest a-test•    (baz |gaz)))'
-    //         );
-    //         const b = docFromTextNotation(
-    //             '(foo bar)•(with-test•  (deftest |a-test|•    (baz gaz)))'
-    //         );
-    //         const range: [number, number] = [b.selectionLeft, b.selectionRight];
-    //         expect(getText.currentTopLevelFunction(a)).toEqual([
-    //             range,
-    //             b.model.getText(...range),
-    //         ]);
-    //     });
+        it('Finds top level function when nested', () => {
+            const a = docFromTextNotation(
+                '(foo bar)•(with-test•  (deftest a-test•    (baz |gaz)))'
+            );
+            const b = docFromTextNotation(
+                '(foo bar)•(with-test•  (deftest |a-test|•    (baz gaz)))'
+            );
+            const range: [number, number] = [b.selectionLeft, b.selectionRight];
+            expect(
+                getText.currentTopLevelFunction(a, a.selection.active)
+            ).toEqual([range, b.model.getText(...range)]);
+        });
 
-    //     it('Finds top level function when namespaced def-macro', () => {
-    //         // https://github.com/BetterThanTomorrow/calva/issues/1086
-    //         const a = docFromTextNotation(
-    //             '(foo bar)•(with-test•  (t/deftest a-test•    (baz |gaz)))'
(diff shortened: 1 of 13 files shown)


Cyrik:
First stab at it. It would be nicer to have access to the pretty printer and so on, but I think this does 80%. I'll have a look at getting the highlighting to work as well.

Cyrik:
Ah brittle integration tests, who wrote that one? 😓

src/providers/hover.ts
```diff
@@ -69,7 +69,13 @@ export async function provideHover(
...
                             const hover = new vscode.MarkdownString();
                             hover.isTrusted = true;
-                            hover.appendMarkdown(_.trim(text, '"'));
+                            hover.appendMarkdown(
+                                _.replace(
+                                    _.replace(_.trim(text, '"'), /\\n/g, '\n'),
+                                    /\\"/g,
+                                    '"'
+                                )
+                            );
```
Cyrik: does anyone know of a better way to do this in js? I wanted to do it with a regex group, but "\$1" is not the same as "\n" even if $1 === "n", which is annoying.

Cyrik:
So after talking with @PEZ I fixed up the snippet and added it to the docs instead. If we find a way to handle large data we can pull it in as a default at a later date.
For now, this PR fixes markdown newline handling which was already requested in slack. So I think we should pull it in.

src/providers/hover.ts
PEZ: I don't think it's possible since the characters are `\` and `n`, $1 will never contain a newline.

Maybe something like this is clearer though:
```typescript
    const hoverText = text
        .replace(/^"|"$/g, '')
        .replace(/\\n/g, '\n')
        .replace(/\\"/g, '"');
    hover.appendMarkdown(hoverText);
```

(Untested)

Cyrik: Yeah. I'll make a to-do for myself to see if I can somehow get the repl to provide me with non-escaped string output so we don't have to run this. But your version does look more readable. We need JS threading macros 🤯

Cyrik:
Ok so I've done multiple extra things now:
- get-text.ts now takes a position in more of its functions (previously it could only work on the current cursor, now you can pass in a position 
- removed the tests for it for now😞  those tests assume that the active cursor should be found by the document itself, which doesn't work if we want to handle hovers. I'll think about how we can still test the functionality here.

src/extension-test/unit/util/cursor-get-text-test.ts
```diff
@@ -3,64 +3,64 @@ import { docFromTextNotation } from '../common/text-notation';
...
+    //         );
+    //         const b = docFromTextNotation(
+    //             '(foo bar)•(deftest |a-test|•  (baz gaz))'
+    //         );
+    //         const range: [number, number] = [b.selectionLeft, b.selectionRight];
+    //         expect(getText.currentTopLevelFunction(a)).toEqual([
+    //             range,
+    //             b.model.getText(...range),
+    //         ]);
+    //     });
```
PEZ: Instead of commenting, you can use `it.skip` (possibly also `describe.skip`, idk). There's a report on the number of skipped tests in at the end of the test output.

src/util/cursor-get-text.ts
```diff
@@ -6,9 +6,12 @@ import { EditableDocument } from '../cursor-doc/model';
 
 export type RangeAndText = [[number, number], string];
 
-export function currentTopLevelFunction(doc: EditableDocument): RangeAndText {
-    const defunCursor = doc.getTokenCursor(doc.selection.active);
-    const defunStart = defunCursor.rangeForDefun(doc.selection.active)[0];
+export function currentTopLevelFunction(
+    doc: EditableDocument,
+    active: number
```
PEZ: Can you make the current document selection the default, maybe?

```suggestion
    active: number = doc.selection.active
```

We do that in a lot of places. I haven't looked super closely, but maybe it will alow the tests to continue working.

PEZ:
Let us think about how we can let the tests stay. I'd rather see tests added for the hover cases. 😄

src/extension-test/unit/util/cursor-get-text-test.ts
Cyrik: Ah interesting, thanks! I don't want to merge with this commented out, just not explode on commit, but skip seems better anyway.

src/util/cursor-get-text.ts
Cyrik: I fixed all the places that use it since they have access to the doc anyway and then just pass in the `doc.selection.active`. The only place this is a problem is the tests, so I'd rather not add a default case that's not being used.
But actually, you fixed my thinking here. I'll get back to you in a minute...

PEZ: You can see it as a way to inject the selection from the tests.

PEZ: Ah, yeah, [that's](https://github.com/BetterThanTomorrow/calva/pull/1582/commits/4bf27621c50166024bb8be759d66772345d9bfa8) a way to do it of course. :smile:

Cyrik:
Turns out I was just having a brain fart with the tests, I can just easily do the same thing I did on all the other callsites...
I was just confused that our internal models have different paths for the same things that `vscode.TextDocument` has.

src/extension-test/unit/util/cursor-get-text-test.ts
PEZ: Yes `skip` doesn't signal ”old dead code that should be cleaned out” 😄 Apart from also being tallied and reported. We currently have 4 pending tests the report says:

image

And I do intend to get back to them. Haha.

src/util/cursor-get-text.ts
PEZ: I still think it makes sense to have a default behaviour of the current cursor position. That has worked very well for us so far.

Cyrik: Ok. I'll add it

PEZ (changes requested):
Thanks for doing this work! Now we only lack a changelog entry. ❤️

PEZ (approved):
🙏

Issue #1551, opened 2022-02-20
Metadata is not passed on form evaluation


daslu:
Just noticed a little gap between CIDER and Calva.
If we have the following code:
```clj
^:abcd [1]
```
(that is, a vector with a number in it, with some metadata before it), and we put the cursor after this piece of code,
then `cider-eval-last-sexp` would pass the code `^:abcd [1]` to nREPL,
while `Calva: Evaluate Current Form` would pass only `[1]`.

PEZ:
Thanks. I agree with CIDER about this. We treat reader tags as part of the form they tag, and this is very similar.

What do we say about ignores? They are reader tags, but treated a bit differently (not part of the form they ignore out) by Calva, to make it convenient to evaluate an ignored form. What does CIDER do here?

PEZ:
Here's a VSIX fixing most of this issue: https://18486-125431277-gh.circle-artifacts.com/0/tmp/artifacts/calva-2.0.247-1551-current-form-metadata-58199575.vsix

Please put it to some tests. See #1566 about what to expect.

PR #1589, opened 2022-03-09
Remove some cursor meta-data awareness

### LINKED ISSUE #1588: Dumb down the token cursor some dealing with meta data and readers
With [Calva v2.0.247](https://github.com/BetterThanTomorrow/calva/releases/tag/v2.0.247) we introduced some changes to the token cursor, making it meta-data-”aware”.

It has caused two critical bugs with structural editing hanging on unbalanced code:

* #1573 
* #1585 

They are now fixed. Fixing #1585 I choose to dumb down the token cursor operation `backwardUpList` which was placing the cursor in front of any metadata. The problem is that `backwardUpList` is being used from many places and in some unbalance cases this caused an infinite loop. Now this placing of the final cursor is done by the Paredit `backwardUpList` command instead, which was where it was needed. Now the token cursor function can safely be used again.

There are more places where this metadata-awareness was added to the token cursor, and for many of them we should do the same. Dumb them back down again, and do the adjustments at the call sites needing it, or make special versions of the functions.

PEZ:
## What has Changed?

Following up on #1587, I set out to remove more of the meta data awareness I added to the token cursor when addressing #1551. Much to my surprise none of that could actually be removed. The addition to handle metadata in `cursor.backWardSecp()` must be there, or we're in for a major rewrite of things.

The only place where we were considering metadata outside of this was in `cursor.downList()`, and was added long ago (by @bpringe) to support metadata when the debugger uses the cursor. Since I have decided that we need to move the support of metadata navigation out of the cursor primitives (except for forward/backward sexp) I then removed this from `cursor.downList()` as well. Then prepared to fix all test cases that now would break.

Surprise number two: No test cases broke, except the one testin that `cursor.downList()` could be made to skip metadata. Not even the debugger util tests that tested this. And everything seems to keep working when I try it manually. I don't quite understand why, but if I have found some potentially problematic code that can be removed without any issues, I'm happy.

Can you throw some testing on this from the debugger use, @bpringe? The unit tests you wrote seem to cover the needed bases, but anyway.

Fixes #1588

## My Calva PR Checklist

I have:

- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ Changes to Calva.
 - Maintenance: [Dumb down the token cursor some dealing with meta data and readers](https://github.com/BetterThanTomorrow/calva/pull/1585)
 
 ## [2.0.253] - 2022-03-09
-- Fix: [over snippet markdown and adds example](https://github.com/BetterThanTomorrow/calva/pull/1582)
+- Fix: [Structural editing hangs in specific cases of unbalanced forms](https://github.com/BetterThanTomorrow/calva/pull/1585)
+- Fix: [Hover snippet markdown and adds example](https://github.com/BetterThanTomorrow/calva/pull/1582)
 - Maintenance: [Begin work on enabling strictNullChecks in the TypeScript config.](https://github.com/BetterThanTomorrow/calva/pull/1568)
 
 ## [2.0.252] - 2022-03-05
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,9 @@
 Changes to Calva.
 
 ## [Unreleased]
+- Maintenance: [Dumb down the token cursor some dealing with meta data and readers](https://github.com/BetterThanTomorrow/calva/pull/1585)
 
 ## [2.0.253] - 2022-03-09
-- Fix: [Structural editing hangs in specific cases of unbalanced forms](https://github.com/BetterThanTomorrow/calva/pull/1585)
 - Fix: [over snippet markdown and adds example](https://github.com/BetterThanTomorrow/calva/pull/1582)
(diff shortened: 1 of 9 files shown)


bpringe:
@PEZ If you try to debug this function using `#dbg`, and step through it to completion using Step Over, you'll see the cursor gets places in the wrong place (because it's going down into the metadata, instead of the form after it), and eventually an error occurs.

```clojure
#dbg
 (defn test-metadata-symbol
   [x]
   (let [y x]
     ^{:hello "world"}
     (+ x ^{:inner "meta"} (+ 1 y))))
```

image

We should add a test for that scenario, and then either not remove the metadata check in `downList`, or find some other solution that works.

bpringe:
Here's an example using `#break`, in which the cursor is placed in the wrong place.

```clojure
(defn test-metadata-symbol
  [x]
  (let [y x]
    ^{:hello "world"}
    (+ x ^{:inner "meta"} (+ 1 #break y))))
```

image

PEZ:
Thanks for the test cases! I'll have a look.

PEZ:
I've extracted the paredit downList logic into a cursor function that handles metadata. Using that function in the debug breakpoint navigation makes your examples work. I added unit tests for it. Took me a little while to figure out the `coor`, but it was worth it. Very smart device!

The debugger still steps a bit weirdly through those functions. But it is weird in the same way it was before this PR, so I'd say we can fix that later, if it should be fixed.

bpringe:
Awesome! Yeah, I believe I noticed the same weird behavior.

> Took me a little while to figure out the coor, but it was worth it. Very smart device!

That comes from cider-nrepl. It's an interesting way to describe a position in a form starting at the top level.

bpringe (approved):

PR #1594, opened 2022-03-13
Explore inlay hints for REPL results

orestis:
## What has Changed?

- I have added an InlayHintsProvider to try and show inline REPL results in a nicer way
- Unfortunately, it doesn't seem that this API is flexible enough for this
- Opening a PR just for archival purposes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,7 +42,7 @@
                 "@types/glob": "^7.1.1",
                 "@types/lodash": "^4.14.167",
                 "@types/mocha": "^9.0.0",
-                "@types/vscode": "^1.45.0",
+                "@types/vscode": "^1.65.0",
                 "@typescript-eslint/eslint-plugin": "^5.11.0",
                 "@typescript-eslint/parser": "^5.11.0",
                 "chai": "^4.2.0",
@@ -79,7 +79,7 @@
                 "webpack-cli": "^4.5.0"
             },
             "engines": {
-                "vscode": "^1.63.0"
+                "vscode": "^1.65.2"
             }
         },
         "node_modules/@babel/code-frame": {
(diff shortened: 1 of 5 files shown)


orestis:
Perhaps the inlay hints API could be useful for something else, but it seems to be very narrowly oriented towards type hints. I couldn't get the tooltip to show up, neither the command to execute, as both seem to be related to a position in the code that actually has a symbol.

orestis:
I used [this](https://fossies.org/linux/vscode/extensions/typescript-language-features/src/languageFeatures/inlayHints.ts) as a guide since the API is very sparsely documented.

PEZ:
Thanks for providing this! Even if it turns out to be a dead end, it can save others from entering. 😄 

The other use for this that I am thinking about, is to place a button after the current form and current top level form that evaluates the respective form. Maybe that's not possible either for the same reasons?

orestis:
I couldn’t get the commands executed. Perhaps I used the wrong syntax though, not sure. To try things out just eval things and if the hints don’t show up immediately then scroll a bit or go to a different file and back again. (The lack of control for making hints show up is another factor against using this for interactive stuff)

orestis:
An addendum: The hovers didn't show up because I have turned off hovers 🤦 

It'd be interesting if someone with a more standard VScode setup could try this branch.

PEZ:
Tried it now. I can reproduce the ”need to scroll to wake the hints up” problem.

However, I get the tooltips and I can execute the commands. (cmd-click)

We'll need to clear previous hints, but regardless I'd say this is success. 😄  

image

Cyrik:
This does look fun, I'll have a look as well.
>The other use for this that I am thinking about, is to place a button after the current form and current top level form that evaluates the respective form. Maybe that's not possible either for the same reasons?

I've also been thinking about how to expose more eval commands as GUI elements so that they are easier to discover for beginners. 
The other three options I thought about are CodeLenses, similar to pyhton:
run codelsens
Adding test arrows to top-level forms:
Screen Shot 2022-03-13 at 12 41 48 PM
Add some of the eval commands to the "ctrl+." CodeActions provider (probably the ones that have more complex hotekyes, like "eval to tap>")

I like the CodeAction and CodeLense versions the best, but the inlay-hints screenshot does look interesting.

riotrah:
Oh wow I'd love to implement all the above and make it configurable or something !!!

In general yes, more "vscode-native"-ish and GUI-integrated the commands are, the better imo

PR #1595, opened 2022-03-13
Add commands and UI for stopping, starting and downloading clojure-lsp

### LINKED ISSUE #1592: Add commands and UI for controlling the clojure-lsp process
It takes a reload of the VS Code window to restart clojure-lsp after having changed version of formatting config or whatever. This can be quite expensive, killing the REPL for instance.

The solution probably is to let the user start and stop clojure-lsp at will. Then we can also add the option to run without clojure-lsp for projects where it consume to much resources.

There is also sometimes troubles with the clojure-lsp download going wrong and then causing problems. So a command for forcefully re-download would be nice as well.

PEZ:
## What has Changed?

* **When context**: `clojure-lsp:active`
* **Command**: Stop clojure-lsp
* **Command**: Start clojure-lsp
* **Command**: Download configured clojure-lsp version
* **Setting**:  Start clojure-lsp when Calva starts (default **enabled**)
* Status bar item always\* active, indicating download, starting, started, stopped
* Trigger download if clojure-lsp binary missing

(\*) If lsp is configured to not auto-start, the status bar item is not displayed until the start or download server commands are issued.

The status bar item has two menus. 

**Clojure lsp active**:

* Show server-info
* Open log file
* Stop server

**Clojure lsp not active**:

* Download configured server version
* Start server

I think we are lacking some management of the status and statusbar when downloading fails or starting fails. It could be something we can add later of course.

I think that the triggering of download if the clojure-lsp binary is missing might not be necessary, now that the user can retrigger download, but it's also silly to try start a client that isn't there.
 

Fixes #1592 

## My Calva PR Checklist

I have:

- [ ] Tests

Ping @pez, @bpringe


docs/site/clojure-lsp.md
```diff
@@ -8,16 +8,18 @@ description: Learn about how Calva leverages clojure-lsp for a lot of the featur
...
+Unless you set `calva.enableClojureLspOnStart` to `false`, Calva will automatically start clojure-lsp. And you won't need to install antyting, because Calva handles that. It can take a while for clojure-lsp to start, though. Especially the first time for a new project, when clojure-lsp (via `clj-kondo`) indexes the project files.
 
 Calva will show a status bar message during the download and while the server is starting, which will go away once the server is ready. However, _much of Calva's functionality is available regardless of the LSP server_, so please start using Calva while this server is starting.
 
 !["Clojure-lsp status bar downloading and intializing messages"](images/clojure-lsp/lsp-status-bar-message.gif "Clojure-lsp status bar downloading and intializing messages")
 
+Therea are commands for stopping and starting the clojure-lsp server. And you can also disable the autostarting using the `calva.enableClojureLspOnStart` setting (we recommend to keeping the default here and always enjoy clojure-lsp goodness).
```
Cyrik: You should probably mention the names of those commands 😇

Cyrik (changes requested):

docs/site/clojure-lsp.md
PEZ: Hmm. You might be right. I often try to avoid that to decouple things a bit. Seems like searching the command palette for **lsp start**  or **start lsp** is so easy at hand... But I'll consider. I still have to update the docs about some more from these changes.

PEZ: Please have another read of the docs now.

Cyrik (approved):

docs/site/clojure-lsp.md
```diff
@@ -7,16 +7,52 @@ description: Learn about how Calva leverages clojure-lsp for a lot of the featur
...
 
-## Starting the LSP server
+## The LSP server starts automatically
 
-You don't need to do anything to start clojure-lsp. No install, no commands, no nothing. Calva downloads the correct binary for your operating system if necessary (this should only happen when the clojure-lsp version is updated in a new release of Calva) and then starts it. It does take a while for clojure-lsp to start, though, especially the first time for a new project, when clojure-lsp (via `clj-kondo`) indexes the project files.
+Unless you set `calva.enableClojureLspOnStart` to `false`, Calva will automatically start clojure-lsp. And you won't need to install antyting, because Calva handles that. It can take a while for clojure-lsp to start. Especially the first time for a new project, when clojure-lsp (via `clj-kondo`) indexes the project files.
```
bpringe:
```suggestion
Unless you set `calva.enableClojureLspOnStart` to `false`, Calva will automatically start clojure-lsp. And you won't need to install antyting, because Calva handles that. It can take a while for clojure-lsp to start, especially the first time for a new project, when clojure-lsp (via `clj-kondo`) indexes the project files.
```

docs/site/clojure-lsp.md
```diff
@@ -7,16 +7,52 @@ description: Learn about how Calva leverages clojure-lsp for a lot of the featur
...
+
+### Active clojure-lsp
+
+When clojure-lsp is downloaded, started and initialized the status bar indicator will look like so:
+
+!["Active clojure-lsp status bar item"](images/clojure-lsp/calva-clojure-lsp-statusbar-active.png "Active clojure-lsp status bar item")
+
+Clicking the item will bring up this menu:
+
+!["Active clojure-lsp  menu"](images/clojure-lsp/calva-clojure-lsp-menu-active.png "Active clojure-lsp menu")
```
bpringe:
```suggestion
!["Active clojure-lsp menu"](images/clojure-lsp/calva-clojure-lsp-menu-active.png "Active clojure-lsp menu")
```

src/lsp/main.ts
```diff
@@ -358,20 +359,18 @@ const generalCommands = [
...
+        context.subscriptions.push(
+            ...generalCommands.map((command) =>
+                vscode.commands.registerCommand(command.name, command.handler)
+            )
+        );
+        context.subscriptions.push(
+            ...clojureLspCommands.map((command) => registerLspCommand(command))
+        );
+    }
+    lspCommandsRegistered = true;
```
bpringe, line 373: You might also be able to store the disposables when registering and dispose of the commands when the server is stopped, which might be better than leaving them registered, but that might not be super important and can be done later.

src/lsp/main.ts
```diff
@@ -411,40 +410,61 @@ function registerEventHandlers(
...
+    setStateValue(LSP_CLIENT_KEY, undefined);
+    if (client) {
+        return client.stop();
+    }
+}
+
+async function startClientCommand() {
+    await maybeDownloadLspServer();
+    lspStatus.show();
+    void startClient();
```
bpringe: Is there a reason to not await this here?

src/lsp/main.ts
```diff
@@ -552,42 +602,120 @@ function registerDiagnosticsCommands(context: vscode.ExtensionContext): void {
...
+            extensionContext.extensionPath,
+            util.isWindows
+        );
     }
+    return readVersionFile(extensionContext.extensionPath);
+}
+
+function deactivate(): Promise<void> {
+    void stopClient();
     return Promise.resolve();
```
bpringe: I think you can just `return stopClient()` here and remove the `Promise.resolve()`.

src/lsp/main.ts
```diff
@@ -681,9 +809,77 @@ function getClient(timeout: number): Promise<LanguageClient> | undefined {
...
+    const STOP_OPTION = 'Stop';
+    const STOP_COMMAND = 'calva.clojureLsp.stop';
+    const INFO_OPTION = 'Show server info';
+    const INFO_COMMAND = 'calva.diagnostics.clojureLspServerInfo';
+    const LOG_OPTION = 'Open log';
+    const LOG_COMMAND = 'calva.diagnostics.openClojureLspLogFile';
+    const commands = {};
+    commands[STOP_OPTION] = STOP_COMMAND;
+    commands[INFO_OPTION] = INFO_COMMAND;
+    commands[LOG_OPTION] = LOG_COMMAND;
```
bpringe, line 867: This might just be me being picky 😄 , but you could make this just an object literal:

```typescript
{
   STOP_OPTION: STOP_COMMAND
}
```

etc.

(That key might need brackets around it - not sure.)

bpringe (changes requested):

src/lsp/main.ts
PEZ: Depends on where is ”this” 😄 If you mean the download, then yes. But that's so obvious, so I'm going to assume `startClient()`. Dunno. Was thinking this is always called in fire-and-forget manner, but maybe it can be used in a promise chain? If so we should await. I'll fix that.

src/lsp/main.ts
PEZ: The difference is that I want an object that is
```typescript
{
  'Stop': 'calva.clojureLsp.stop'
}
```

Whereas what you have there would give me:

```typescript
{
  'STOP_OPTION': 'calva.clojureLsp.stop'
}
```

Maybe you know some trick with brackets that would work? If so, I'm all for it, because the way I'm doing it now is error prone.

Cyrik:
🥳

src/lsp/main.ts
bpringe: Not sure why I suggested that method the first time 😄 . But brackets would work:

image

PR #1596, opened 2022-03-13
merge completions

### LINKED ISSUE #1498: Add configurability for which subsystem (nrepl, clojure-lsp, other) that provides features, such as code navigation, etc
Currently, a lot of features that are provided by LSP get overwritten by nrepl when you are connected to it. There have been a few requests for this to be optional in Slack.

I just ran headfirst into this when working in LSP itself, since I couldn't test some of the features when connected to it by repl 😅 

Not sure if it's worth the effort/complexity, but we could provide an options map that lets users decide which provider is used. Something like:
```
{
    :provide-hover :nrepl
    :provide-completion :lsp
    :provide-snippets :merge
}
```

Even if we don't want to make it optional, we should probably build a merge for some of the providers.

What do you think?

Cyrik:
## What has Changed?

- completion middleware now returns null
- completion provider has a hardcoded option with priority between lsp/repl and a merge option
- completion provider calls both repl and lsp for results, depending on the options set

Closes #1498 

## My Calva PR Checklist

I have:

- [X] Added to or updated docs in this branch, if appropriate
- [X] Tests
     - [X] Smoke tested the extension as such.
     - [X] Tested the VSIX built from the PR (so, after you've submitted the PR). You'll find the artifacts by clicking _Show all checks_ in the CI section of the PR page, and then _Details_ on the `ci/circleci: build` test.
     - [X] If I am fixing just part of the issue, I have just referenced it w/o any of the "fixes” keywords.
- [X] Formatted all JavaScript and TypeScript code that was changed. (use the [prettier extension](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) or run `npm run prettier-format`)
- [X] Confirmed that there are no linter warnings or errors (use the [eslint extension](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint), run `npm run eslint` before creating your PR, or run `npm run eslint-watch` to eslint as you go).

Ping @pez, @bpringe

### DIFF
--- a/src/lsp/main.ts
+++ b/src/lsp/main.ts
@@ -13,8 +13,6 @@ import {
 } from 'vscode-languageclient/node';
 import * as util from '../utilities';
 import * as config from '../config';
-import { provideClojureDefinition } from '../providers/definition';
-import { provideCompletionItems } from '../providers/completion';
 import { setStateValue, getStateValue } from '../../out/cljs-lib/cljs-lib';
 import { downloadClojureLsp, getLatestVersion } from './download';
 import { readVersionFile, getClojureLspPath } from './utilities';
@@ -152,15 +150,10 @@ function createClient(clojureLspPath: string, fallbackFolder: FallbackFolder): L
             return next(document, position, token);
           }
         },
-        async provideDefinition(document, position, token, next) {
-          const definition = await provideClojureDefinition(document, position, token);
-          if (definition) {
-            return null;
-          } else {
-            return next(document, position, token);
-          }
+        provideDefinition(_document, _position, _token, _next) {
+          return null;
         },
-        provideCompletionItem(document, position, context, token, next) {
+        provideCompletionItem(_document, _position, _context, _token, _next) {
           return null;
         },
         async provideSignatureHelp(document, position, context, token, next) {
--- a/src/lsp/main.ts
+++ b/src/lsp/main.ts
@@ -160,13 +160,8 @@ function createClient(clojureLspPath: string, fallbackFolder: FallbackFolder): L
             return next(document, position, token);
           }
         },
-        async provideCompletionItem(document, position, context, token, next) {
-          const items = await provideCompletionItems(document, position, token, context);
-          if (items) {
-            return null;
-          } else {
-            return next(document, position, context, token);
-          }
+        provideCompletionItem(document, position, context, token, next) {
+          return null;
         },
         async provideSignatureHelp(document, position, context, token, next) {
           const help = await provideSignatureHelp(document, position, token);
--- a/src/lsp/main.ts
+++ b/src/lsp/main.ts
@@ -766,7 +766,7 @@ export async function getClojuredocs(symName: string, symNs: string): Promise<an
 }
 
 // TODO: This feels a bit brute, what are other ways to wait for the client to initialize?
-function getClient(timeout: number): Promise<LanguageClient> {
+export function getClient(timeout: number): Promise<LanguageClient> {
   const start = Date.now();
   return new Promise(waitForClientStarted);
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Changes to Calva.
 
 ## [Unreleased]
 - [Add command for formatting away multiple space between forms on the same line](https://github.com/BetterThanTomorrow/calva/issues/1677)
+- [Merge LSP + NREPL completions and goToDefintion](https://github.com/BetterThanTomorrow/calva/issues/1498)
 
 ## [2.0.266] - 2022-04-11
 - Fix: [Jump between source/test does not work properly with multiple workspace folders](https://github.com/BetterThanTomorrow/calva/issues/1219)
--- a/src/providers/completion.ts
+++ b/src/providers/completion.ts
@@ -10,9 +10,7 @@ import {
   CompletionItemProvider,
   CompletionItem,
   CompletionItemLabel,
-  ProviderResult,
 } from 'vscode';
-import * as state from '../state';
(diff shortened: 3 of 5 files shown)


Cyrik:
A few interesting things I learned along the way:
1. cancellation tokens are evil when trying to manually test UI stuff...
2. we are currently calling every provider twice, once directly and once from the middleware...
3. merging completions is actually not too hard, merging resolveCompletions is a lot harder...

src/lsp/main.ts
```diff
@@ -160,13 +160,8 @@ function createClient(clojureLspPath: string, fallbackFolder: FallbackFolder): L
...
         },
-        async provideCompletionItem(document, position, context, token, next) {
-          const items = await provideCompletionItems(document, position, token, context);
-          if (items) {
-            return null;
-          } else {
-            return next(document, position, context, token);
-          }
+        provideCompletionItem(document, position, context, token, next) {
+          return null;
```
riotrah, line 157: From what I gather:

- returning null in this middleware means the lsp client, and thus calva, will not proceed to procure clojure-lsp's default completion item candidates

If we introduce your other changes but not this one, what would happen? Would the merged completions then be overwritten by clojure-lsp's outright?

riotrah (approved):

src/lsp/main.ts
Cyrik: There's basically two ways in vscode to get completions: have a CompletionProvider or have middleware. We had both, so every completion was actually run twice.
I could also have deleted the provider and put the code into the middleware, but the provider feels like a cleaner solution.

And to answer your question: I think what ever returns first gets used, so not ideal.

riotrah: Nice. Yeah makes sense. Was interesting learning about [the two ways you speak of](https://code.visualstudio.com/api/language-extensions/programmatic-language-features). I'm surprised to have found little about the semantics/consequences of "collisions" here. Still, very excited about the potential here.

Cyrik: Yeah, I think clojure/calva is kind of unique in getting its information from two different sources. Usually you'd use the Provider to do the calculations inside the extension if you don't have an LSP or just do everything inside the LSP. Using the middleware for merging is kind of weird, because the `next` call might just cancel your promise :)

PEZ: To the question. I think it might end up with two answers and the user would be offered a choice. A rather confusing choice where the options often would be identical. Not sure about this, but I think I have seen it happen few time when hacking on these things.

> the provider feels like a cleaner solution

Agree.

PEZ:
@Cyrik are we waiting for both to get back with an answer before we choose one? Asking because clojure-lsp is often a lot slower than nREPL. So maybe it makes sense to let the service that is quickest to answer win, and then when we have both answers we let the config decide. And maybe this is how it works, just checking.

Illustrating here with Signature/Parameter help:

https://user-images.githubusercontent.com/30010/163178808-115d7e2a-7149-4d28-8db8-cc24b27db3ba.mp4

Not sure the lsp lag is fully visible, but I certainly notice it when I use it. (Also two weird things with the lsp signatures shown in this vid: 1. It highlights the first param even when the cursor is at the function element. 2. It shows the signature for `f2` when it should when it should show the signature for `f1` in `(f1 |(f2 a1 a2))`.)

PEZ:
Checked the code a bit closer now, and it looks to me like we are waiting for both to come back with answers, before reconciling. (b/c `merge: true`). Maybe we should have `merge: false` for now, and prioritize nrepl before lsp? It might be that the lag is not so noticeable with completions as it is with signatures, but I also don't know how it behaves on a slower computer than mine (which is a performance monster).

src/lsp/main.ts
Cyrik: @riotrah btw you'll actually find a better explanation of the different ways to handle this [here](https://code.visualstudio.com/api/language-extensions/embedded-languages)

Cyrik:
You are correct, we are waiting for both. I could restructure it to at least run them in parallel, which would help.
As for your speed question: clojure-lsp has a debounce on changes, which causes it to always take 300ms on the completion call (because it's still waiting for the change to complete). Not sure how long nrepl takes to do this.

I've just tried to find a way to update the list with "late" results, but vscode doesn't seem to have a good way to do this. We might be able to hack something together by triggering a completion when the other results arrive and just returning those, but it'll probably be pretty brittle and the UX might jump around 😢 

I'd propose to release this with merge: false and nrepl as priority one on completions, but priority two on definitions. Especially the definition feature seems to be very popular so I don't want to keep the finished version stuck in here 😄  In a follow-up PR I'd paralyze the calls to make it faster and we can also think about how to expose the options so people can choose their preferred order.

BTW the signature call takes around 180ms on my machine

PEZ (approved):
Thanks for clarifying/confirming. I started to think about how to reconcile in step two, and while I could think up that particular scheme, like you I concluded that VS Code wouldn't really participate.

Having played a bit with this in dev, I think it is quite fine as is. clojure-lsp actually has a bit richer completions than we currently create with nrepl. We can fix that with nrepl, but until we do, I think it makes sense wait for lsp and do the merge.

I think the code looks great as well. I could obviously follow along what happens. 😄

PEZ:
From my perspective, you only need to update the changelog and then you can merge this.

Cyrik:
@PEZ just to be clear, you don't want me to change the default priority or the default merge behavior?

PEZ:
Exactly. I think we can give this a shot.

Cyrik:
Ah, I somehow forgot to push the definition version of this...
The idea behind it is the same, it's just easier because there are no merges involved.

Issue #1601, opened 2022-03-17
clojure-lsp does not start in the Getting Started REPL


PEZ:
Clojure-lsp fails to start when firing up a standalone REPL in a folder-less VS Code window.

~I am pretty sure this is a regression cause by~

* #1592

PEZ:
I no longer think this is a regression. It is probably rather that the clojure-lsp status indicator exposes the error. And then the status-indicator does not get updated because we don't handle the error that happens:

When running the serverInfo command on the client there happens an internal error in the lsp server:

```
'java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
\tat org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.lambda$null$0(GenericEndpoint.java:67)
\tat org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.request(GenericEndpoint.java:120)
\tat org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.handleRequest(RemoteEndpoint.java:261)
\tat org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.consume(RemoteEndpoint.java:190)
\tat org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.handleMessage(StreamMessa…173)
\tat clojure_lsp.feature.format$resolve_user_cljfmt_config.invokeStatic(format.clj:21)
\tat clojure_lsp.feature.format$resolve_user_cljfmt_config.invoke(format.clj:19)
\tat clojure_lsp.handlers$server_info$fn__35173.invoke(handlers.clj:186)
\tat clojure_lsp.handlers$server_info.invokeStatic(handlers.clj:185)
\tat clojure_lsp.handlers$server_info.invoke(handlers.clj:177)
\tat clojure.lang.Var.invoke(Var.java:380)
\tat clojure_lsp.server.ClojureLSPServer.serverInfoRaw(server.clj:404)
\t... 15 more'
```

Testing v2.0.253 of Calva, clojure-lsp is indeed working some. But running the serverInfo command (the user command) results in an ”Internal error” message box.

@ericdallo: Do you have some idea about what I could do to try diagnose this further? It only happens when there is no folder opened in VS Code, so might be something with clojure-lsp running into troubles finding the project root.

ericdallo:
[...]

PR #1602, opened 2022-03-17
Handle that clojure-lsp doesn't always work

PEZ:
## What has Changed?

In #1601 we learn that clojure-lsp doesn't actually work when VS Code is not having a folder open. It's unclear if Calva can do anything about it. But what we can do is to make Calva a bit less stupid and unaware about possible problems with clojure-lsp when they are discovered during startup.

The changes here rearranges some code to run only if the clojure-lsp server seems to be playing along. And we try to avoid alerting too loud about the problems since it makes the user think that Calva is broken, when Calva actually is mostly OK and can help with getting started stuff just fine. This is how things mostly behaved before v2.0.254 anyway, even if that was by mistake.

Let's hope we figure out how to get clojure-lsp working also in the folder-less scenarios.

## My Calva PR Checklist

I have:

- [ ] Tests

Ping @pez, @bpringe


src/lsp/main.ts
```diff
@@ -431,7 +430,14 @@ async function startClientCommand() {
...
 }
 
-async function startClient(): Promise<void> {
+async function startClient(): Promise<boolean> {
+    const runningClient = getStateValue(LSP_CLIENT_KEY);
+    if (runningClient) {
+        await runningClient.stop().catch((e) => {
+            console.log('Error stopping running client:', e);
+        });
+    }
```
corasaurus-hex, line 442: Is this a thing we're running into, where there is already a running client and we need to start a new one?

PEZ: Astute observation! What happens right now is that if we stop the client when we notice it is not working, the user gets an error alert. (See separate TODO comment). So we let it run. The user I have in mind here is the beginner, using the getting started repl. Showing that alert could make things seem much more broken then they are.

PEZ: I added a comment here as well, informing why we have this `runningClient` check.

src/lsp/main.ts
```diff
@@ -755,8 +780,10 @@ async function getDocumentSymbols(
     return result;
 }
 
-async function getServerInfo(lspClient: LanguageClient): Promise<any> {
-    return lspClient.sendRequest('clojure/serverInfo/raw');
+function getServerInfo(lspClient: LanguageClient): Promise<any> {
+    return lspClient.sendRequest('clojure/serverInfo/raw').catch((e) => {
+        console.error('clojure/serverInfo/raw failed:', e);
+    });
```
bpringe, line 786: This might return a Promise that resolves to `undefined`, so we might get an error from code like the below which expects the `getServerInfo` promise to be rejected if the `serverInfo` request fails.

So with the code before, the code below the `await` (in the code below) wouldn't have been run if the `serverInfo` request fails, but with the `catch` above, it will be run (since the promise was resolved instead of rejected), but `serverInfo` would be `undefined`.

```clojure
async function openLogFile(): Promise<void> {
  const client = getStateValue(LSP_CLIENT_KEY);
  if (client) {
    const serverInfo = await getServerInfo(client);
    const logPath = serverInfo['log-path'];
    void vscode.window.showTextDocument(vscode.Uri.file(logPath));
  } else {
    void vscode.window.showInformationMessage(SERVER_NOT_RUNNING_OR_INITIALIZED_MESSAGE);
  }
}
```

I'm not positive about this though. Maybe @corasaurus-hex can verify.

PR #1603, opened 2022-03-17
Handle error when nREPL server does not support claspath op

### LINKED ISSUE #1600: An extra quick-pick prompt pops up after jack-in
Sometimes at jack-in, there's an extra quick-pick menu/prompt that shows up even though Calva has all information it needs to spawn the REPL, and also spawns it during this prompt displaying.

NB: As a user, you can dismiss this extra prompt using `escape` (if you see that the jack-in has started).

PEZ:
## What has Changed?

In the new code for collecting runtime configuration we assume that all nREPL servers handle the classpath op. Causing an error to be thrown. This error was later caught by what seems to me to be a very old catch-all in the code asking for url to the nREPL server, that then showed the prompt again.

I've fixed both issues here. The latter one needs some testing, since it is very old code that has not been causing problem before (out of sheer luck, I think, but anyway). 

Fixes #1600 

## My Calva PR Checklist

I have:

- [ ] Tests

Ping @pez, @bpringe


PEZ:
This error happened with shadow-cljs, @Cyrik . I'm not sure it is generally true that `classpath` is not supported.

Cyrik (approved):

Cyrik:
Thanks for fixing this @PEZ. I'll create an issue to look into getting the classpath when using shadow. Would be sad if all the extensibility went away in that case.

src/connector.ts
```diff
@@ -727,7 +731,6 @@ export async function connect(
         }
     } catch (e) {
         console.log(e);
```
bpringe, line 733: @PEZ It might be a good idea to show a window error message here to the user indicating that an error occurred while connecting. We might also want to use `console.error` instead of `console.log` here. Just adding this comment here for later.

PEZ: I don't know in which situations this actually fails. I thought it better to not alarm the user and ask for the port instead. Agree about `console.error`!

PR #1606, opened 2022-03-19
#610 Multi Cursor for paredit 

### LINKED ISSUE #610: Paredit commands don't propagate to multiple cursors
When I have multiple cursors, paredit commands just affects the first one.
This gif shows an example when I try to slurp 3 forms, but it just affects the first one.
ezgif-7-f84d4d5d9ca1
The expected result was:
```clojure
[[:a 1]
 [:b 2]
 [:c 3]]
```

### LINKED ISSUE #1662: Help beta test multi-cursor Calva Paredit, please ❤️ 🙏
@riotrah is bringing something wonderful to Calva, **multi cursor support to Paredit!**

The way this beta test will work is that Calva Team members will maintain a checklist of known issues right here and all who participate in the testing can report findings in the comments. See [How to test custom VSIX packaged](https://github.com/BetterThanTomorrow/calva/wiki/Testing-VSIX-Packages#installing-a-vsix-package) on the Calva wiki if this is unfamiliar to you. (TL;DR: It is super easy.)

* **Issue**: #610
* **PR**: #1606
* **Latest VSIX**: [calva-2.0.266-pull-1606-62a922f0.vsix](https://output.circle-artifacts.com/output/job/70b4e6fd-d010-4050-b47a-50c59b85282d/artifacts/0/tmp/artifacts/calva-2.0.266-pull-1606-62a922f0.vsix)

This is recorded *April 3, 2022*, with the build: [calva-2.0.263-pull-1606-298af21c.vsix](https://output.circle-artifacts.com/output/job/b6601f11-9da8-4a45-8609-8e299028b125/artifacts/0/tmp/artifacts/calva-2.0.263-pull-1606-298af21c.vsix)

https://user-images.githubusercontent.com/30010/161421143-7058384e-447d-4da9-b7f2-7dcb00f16e43.mp4

As you can see it is largely working! As you also can see, sometimes it doesn't do the right thing. So this is a call for help testing this to help us find bugs and to help us discover what is *the right thing* for certain cases.

See [Paredit, a Visual Guide](https://calva.io/paredit/) for how Calva Paredit works for a single cursor (and to appreciate what an effort this is for @riotrah to pull off, and why the rest of us should help with testing.

## Test notation

In the Calva tests for structural navigation and editing we use a special document notation, which is easy to read for both the machine and for us humans:

```typescript
/**
 * Text Notation for expressing states of a document, including
 * current text and selection.
 * See this video for an intro: https://www.youtube.com/watch?v=Sy3arG-Degw
 * * Since JavasScript makes it clumsy with multiline strings,
 *   newlines are denoted with the paragraph character: `§`
 * * Selections are denoted like so
 *   * Cursors, (which are actually selections with identical start and end) are denoted with a single `|`, with <= 10 multiple cursors defined by `|1`, `|2`, ... `|9`, etc, or in regex: /\|\d/. 0-indexed, so `|` is 0, `|1` is 1, etc.
 *   * This is however actually an alternative for the following `>\d?` notation, it has the same left->right semantics, but looks more like a cursor/caret lol, and more importantly, drops the 0 for the first cursor.
 *   * Selections with direction left->right are denoted with `>0`, `>1`, `>2`, ... `>9` etc at the range boundaries
 *   * Selections with direction right->left are denoted with `<0`, `<1`, `<2`, ... `<9` etc at the range boundaries
 */
```

See this CalvaTV video for tutorial: https://www.youtube.com/watch?v=Sy3arG-Degw

A typical report has

1. a description of a before-document 
2. a command
3. a description of a an expected after-document
4. a description of a the actual after-document

We try to keep the document descriptions as minimal as possible. See the first report in the comments for an example.

### There's a command for it

The test VSIX has a command for printing the text notation from the state of the current document.

image

Which makes creating a report involve these steps:

1. Create your minimal document (an Untitled doc in Clojure mode will do) 
1. Place your cursors/selections
1. Issue the *Print TextNotaion*. command
1. Issue the Paredit command
1. Issue the *Print TextNotaion*. command again

Now you have the **Before** and **Actual** texts in the Output channel *Calva says*. You only need to add the **Expected** texts yoursefl.

## Reported issues

* [x] **Delete Forward to End of List** only deletes for the first cursor, and loses the other cursors
* [x] **Delete Backward to End of List** only deletes for the first cursor, and loses the other cursors
* [x] **Raise sexp** wrong placement of cursor after the edit
* [x] **Delete Sexp Forwards** only deletes at primary cursor, loses the others
* [x] **Delete Sexp Backwards** only deletes at primary cursor, loses the others
* [x] **Expand selection** when all text in a list is selected, doesn't select the list. It selects the opening paren
* [ ] **Paredit Delete Backward** does not move cursor inside list when used from adjacent after the list
* [ ] **Paredit Delete Forward** does not move cursor inside list when used from adjacent before the list
* [ ] **Forward Sexp** moves cursor out of line comment instead of moving by word
* [ ] Multi-cursor **Paredit Delete Backward** destroys structure when a cursor is at the beginning of a list
* [ ] Multi-cursor **Paredit Delete Forward** destroys structure when a cursor is at the end of a list
* [ ] Multi-cursor edit commands loose all but the primary cursor (because formatting, which is not multi-cursor aware)
* [x] Multi-cursor **Splice sexp** misplaces cursors
* [x] **Wrap around** commands only wraps primary cursor and loses the other cursors
* [ ] **Delete backward** does not move into list from behind closing bracket

rayat-amperity:
Fixes #610

## What has Changed?

- Multi cursor support for most Paredit commands
- Significant changes to ModelEditSelection API, EditableDocument API
- Added ModelEditResult, ModelEditFunction, ModelEditFunctionArgs types
- Almost all paredit functions, esp those exposed directly to extension as a command handler, now take an array of offsets, ranges, ModelEditSelections etc, where they previously took individual instances of each, and similarly return an array of whatever they previously returned.
  - Where single cursor calculations are required, just pass and expect a single-item array 

Fixes #610 

## My Calva PR Checklist

I have:

- [x] Tests

Ping @pez, @bpringe


PEZ:
Looks like you have had a nice start! Please let me know when you think I should test something. Consider basing this PR onto the one where you removed left/right selection confusion.

rayat-amperity:
@PEZ I already started rebasing this PR on #1608. I'll switch back to prepping **that** PR for merging.

I do have a(nother) question:

> This might be better implemented in the cleanup PR

- The paredit command handlers (eg. `selectRangeForward` (`src/cursor-doc/paredit.ts#49`)) each take args usually like `(doc: EditableDocument, ranges: [number, number]) {}`.
  - Why not take a ModelEditSelection directly? Eg `(doc: EditableDocument, selection: ModelEditSelection) {}`
  - Then, if we extend ModelEditSelection with `start, end, isReversed` properties as I proposed in the Cleanup API Issue, instead of logic like `const rangeRight = Math.max(range[0], range[1]);` as some currently have, we could simply call `const rangeRight = selection.end`

riotrah:
Just putting a note here that the tests have to be rewritten at least somewhat for this.

Your shorthand/dsl for describing documents/text assumes two | symbols are a single directionless selection, instead of 2 cursors. 

This will need to change but I want to do so in the least disruptive way possible. Any thoughts?

PEZ:
I've been thinking a bit about this, actually. I'd like to keep the current semantics as long as we only have one cursor to consider. How about that for multi cursor cases, we number the cursors? `|1`, etcetera?

rayat-amperity:
> I've been thinking a bit about this, actually. I'd like to keep the current semantics as long as we only have one cursor to consider. How about that for multi cursor cases, we number the cursors? `|1`, etcetera?

Sure thing, that's a reasonable idea - it might be tricky if, for any reason, any of the existing tests had a numeral after the `|` like in your example, but if they exist, we'll prob find them.

Similarly then, we're assuming that if there's something like `"(:a {:b '(:c |1 2 [3 |1 2])})"` the 2 `|1` would defined an enclosed direction-less selection, and similarly if they had the `>|1` syntax it would be the same, but directionless, right?

rayat-amperity:
I do need help with another thing @PEZ :

`src/cursor-doc/cursor-context.ts#determineContexts()` is used for setting vsc contexts for some of the more 'intelligent' (or at least contextually aware) commands, that pertain to or would benefit from knowledge of the forms or document contents _around_ the "current" cursor.

That function checks some stuff around the current primary cursor, and then sets vsc contexts like:
 
- 'calva:cursorAtStartOfLine'
- 'calva:cursorAtEndOfLine'
- 'calva:cursorInString'
- 'calva:cursorInComment'
- 'calva:cursorAfterComment'
- 'calva:cursorBeforeComment'

Which are then checked by those aforementioned cursor contextual commands. What I should prob do at some point is list all of them, and see how likely it is users would expect or desire multi cursor for those in particular.

In any case, this clearly doesn't work as-is for multi-cursor.

### A few options:

1. [ ] Figure out some other approach for contextually aware commands that would benefit from multi-cursor
2. [ ] Concede the "loss" or compromise here and simply make it clear to users that those particular commands only operate on the context of the current primary cursor.
3. [ ] Figure this out later, as long as most of the multi cursor works now 😆

PEZ:
Tricky with the contexts! I think that for all commands affected this cursors contexts should be removed and the commands should handle the check on a per cursor basis.

If it should be done in this PR or not... Depends. Mostly on how much work you think is involved. I guess there is some MVP we can cut out. That said, one place I miss multi cursors a lot is when moving by sexpression. 😄

riotrah:
> Tricky with the contexts! I think that for all commands affected this cursors contexts should be removed and the commands should handle the check on a per cursor basis.
> 
> If it should be done in this PR or not... Depends. Mostly on how much work you think is involved. I guess there is some MVP we can cut out. That said, one place I miss multi cursors a lot is when moving by sexpression. 😄

Well, the contexts are really helpful for users making contextual keybindings no? For the "when expressions"?

PEZ:
> Well, the contexts are really helpful for users making contextual keybindings no? For the "when expressions"?

Yes, but they also get to be an ”API” that now proves tricky to maintain. If should keep the context we need to figure out what the semantics of them is in a multi-cursor world. Probably they would refer to ”all cursors”? But that will still lead to weird effects in situations where cursors are in different contexts, like one is in a comment and some others are not, what would `!calva.cursorInComment` mean then?

Hmmm... I think we probably should remove the contexts.

riotrah:
# Regarding `when contexts`:

I made an exhaustive search of all the when contexts available by default in the latest version of VSC, and none of the general purpose ones involve the specific circumstances of one or any cursors.

Ie. there are no `when contexts` that say something 'cursorAtIndentation` or `cursorInComment` or `selectionSpansMultipleLines` etc. The only ones are `editorHasSelection` and `editorHasMultipleSelections`.

This would suggest that it's not a pattern to have the when context evaluated per cursor, and suggests the when context is a boolean check vscode does _for_ the user/extension etc. 

My hope was that the end-user could still use calva's `cursorInComment` etc when expressions in keybindings, and then instead of calva setting that context true or false based on primary cursor and then vscode deciding at keypress-time whether or not to execute the command, that instead, vscode would "inform" calva that the user executed whatever command, but had also specified the when context for it, and then calva would evaluate the expression _per cursor_, and execute per cursor when passing the condition.

Don't think that's possible from my research, tho vsc is massive and full of features.

# Options:

1. As you suggested, simply remove this feature for users
2. Leave em, but they only apply to primary cursor (perhaps rename them to "primaryCursorInComment" ?)
3. Move them from when contexts to standard settings configuration. Not as convenient, but at least possible.

PEZ:
If I remember correctly, vscode does not have contexts for cursor location, so the lack of prior art isn't really telling us much. 

> My hope was that the end-user could still use calva's cursorInComment etc when expressions in keybindings, and then instead of calva setting that context true or false based on primary cursor and then vscode deciding at keypress-time whether or not to execute the command, that instead, vscode would "inform" calva that the user executed whatever command, but had also specified the when context for it

Pretty sure this is not possible.

What's important is that we can keep Calva's default behavior for moving the cursor. I don't think the when contexts should be kept if we can achieve the correct movement without them. And that should be possible by making forward/backward sexp check the cursor position with the code we already have and relay to the builtin move by word commands in those cases where the when contexts would have solved that.

Adding more settings makes Calva more complicated for the users and for us. I'd rather avoid that, if we can.

package.json
```diff
@@ -2722,7 +2722,7 @@
     "watch-docs": "mkdocs serve",
     "clean": "rimraf ./out && rimraf ./tsconfig.tsbuildinfo && rimraf ./cljs-out",
     "update-grammar": "node ./src/calva-fmt/update-grammar.js ./src/calva-fmt/atom-language-clojure/grammars/clojure.cson clojure.tmLanguage.json",
-    "precompile": "npm i && npm run clean && npm run update-grammar && npm run prettier-format",
+    "precompile": "npm i && npm run clean && npm run update-grammar",
```
riotrah, line 2725: Why remove this @PEZ ?

PEZ: It's not needed to compile. It's needed before we run the `prettier-format-watch` script. And there's a `preprettier-format-watch` for that.

dosbol:
any news guys?

PEZ:
@dosbol, unfortunately not. I tried to resolve the conflicts some months ago, but failed miserably. It doesn't mean this work can't be salvaged and continued, just that it needs a focused effort over a longer period.

PR #1609, opened 2022-03-20
Account for multiple project roots

### LINKED ISSUE #1254: Support Polylith and monorepo jack-in/connect better
[Polylith](https://polylith.gitbook.io/polylith/) repositories are monorepos where you are supposed to start one REPL to be able to develop and ”reach” all sub-projects, etcetera. `deps.edn` files are used to configure things on many levels, but there is seldom any use for starting separate REPLs for these.

Today Calva only considers the regular monorepo case with separate projects and separate REPLs. So when jacking in to a project, Calva looks for the ”closest” (from the current directory and up to the repo/window/workspace root) project file (deps.edn, project.clj, shadow-cljs.edn) and presents a jack-in menu based in this file. This doesn't make sense for Polylith projects.

Generally, I think what makes sense is for Calva to ”collect” project files from the current file up to the repo/window/workspace root and present a menu where the user can select one of those. This makes sense regardless of if it is a monorepo or not and regardless if it is a regular monorepo or a Polylith project.

We could consider recognizing Polylith projects as well and give even more guidance, but that might be for a separate issue to request.

PEZ:
Glob for all project roots in the workspace

Let the user pick if there are more than one
Changes the way we find the closest project root
Not using node fs module at all
Fixes #1254

## My Calva PR Checklist

I have:

- [x] Tests

Ping @pez, @bpringe


src/state.ts
```diff
@@ -134,100 +135,66 @@ function getProjectWsFolder(): vscode.WorkspaceFolder | undefined {
...
+    const projectRootPath = await pickProjectRootPath(candidatePaths, closestRootPath);
+    setStateValue(PROJECT_DIR_KEY, projectRootPath);
+    setStateValue(PROJECT_DIR_URI_KEY, vscode.Uri.file(projectRootPath));
   }
 }
 
+async function findProjectRootPaths() {
+  const projectFileNames: string[] = ['project.clj', 'shadow-cljs.edn', 'deps.edn'];
+  const projectFilesGlob = `**/{${projectFileNames.join(',')}}`;
+  const excludeDirsGlob = `**/{${config.getConfig().projectRootsSearchExclude.join(',')}}`;
```
corasaurus-hex: I see `target` is in the list of excluded directories. Would that exclude a directory named `target-files` or does it need to be just `target` by itself in order to match this glob?

src/state.ts
```diff
@@ -134,100 +135,66 @@ function getProjectWsFolder(): vscode.WorkspaceFolder | undefined {
...
+  const t0 = new Date().getTime();
+  const candidateUris = await vscode.workspace.findFiles(projectFilesGlob, excludeDirsGlob, 10000);
+  console.debug('glob took', new Date().getTime() - t0, 'ms');
+  const projectFilePaths = candidateUris.map((uri) => path.dirname(uri.fsPath));
+  const candidatePaths = [...new Set(projectFilePaths)];
+  console.log(candidatePaths);
```
corasaurus-hex: It might be cool to make this `console.log({candidatePaths})` so the log entry contents are labeled?

src/state.ts
```diff
@@ -134,100 +135,66 @@ function getProjectWsFolder(): vscode.WorkspaceFolder | undefined {
...
+
+function findClosestProjectRootPath(candidatePaths: string[]) {
+  const doc = util.tryToGetDocument({});
+  console.log(doc);
+  const docDir = doc && doc.uri ? path.dirname(doc.uri.fsPath) : undefined;
+  const closestRootPath = docDir
+    ? candidatePaths
+        .filter((p) => docDir.startsWith(p))
+        .sort()
+        .reverse()[0]
```
corasaurus-hex: Nice! That's a good way to find the longest match 😄

corasaurus-hex (approved):
I had a couple questions but other than that this is beautiful 😚👌🏻

src/config.ts
```diff
@@ -161,6 +161,7 @@ function getConfig() {
       useDeprecatedAliasFlag: configOptions.get<boolean>('jackIn.useDeprecatedAliasFlag'),
     },
     enableClojureLspOnStart: configOptions.get<boolean>('enableClojureLspOnStart'),
+    projectRootsSearchExclude: configOptions.get<string[]>('projectRootsSearchExclude'),
```
corasaurus-hex: Oh! I missed one thing! We should probably add a default here so that it satisfies the gods of `strictNullChecks`:

```
configOptions.get<string[]>('projectRootsSearchExclude') ?? []
```

src/state.ts
PEZ: It will only exclude `target`, exactly as that. Wherever it appears, though...

src/state.ts
PEZ: TIL.

Wasn't intending to keep this logging in there, though. Except the timing of the glob, which could be good to be able to ask users about for a while.

src/config.ts
PEZ: Thanks. I used the default value signature of the `get`:

```ts
configOptions.get<string[]>('projectRootsSearchExclude', []) 
```

Does that also satisfy the rule? Less syntax to parse...

corasaurus-hex: ooooh, I didn't know about that one! I think that should fix it? and If not I can fix it later when i'm fixing null check things 😊

PEZ:
@Cyrik : can you have a look at e356cfd and see if you think it is viable way to deal with the quick-pick menus in e2e tests? I didn't figure out a very nice way to select the correct sequence...

I'd be awesome if @corasaurus-hex and @bpringe had a look as well.

:heart:

PEZ:
I wonder why that `saveAs` doesn't do the trick now... I'll have a closer look. Thanks!

src/connector.ts
```diff
@@ -668,27 +659,29 @@ async function standaloneConnect(
 export default {
   connectNonProjectREPLCommand: async (context: vscode.ExtensionContext) => {
     status.updateNeedReplUi(true);
+    let projectDirUri = state.getProjectRootUri();
+    if (!projectDirUri) {
+      projectDirUri = await state.getOrCreateNonProjectRoot(context, true);
+    }
```
corasaurus-hex: it looks like `getOrCreateNonProjectRoot` calls `getProjectRootUri` internally, first thing, and returns it later. Could we ditch the `getProjectRootUri` here and just call `getOrCreateNonProjectRoot` on its own here? Or should we remove the `getProjectRootUri` from `getOrCreateNonProjectRoot`? Or just leave everything as-is?

src/connector.ts
```diff
@@ -668,27 +659,29 @@ async function standaloneConnect(
...
+    await state.initProjectDir(projectDirUri);
```
corasaurus-hex: We add a `.catch` to this function in `connectCommand`, do we need one here?

src/utilities.ts
```diff
@@ -547,6 +558,7 @@ export {
   logSuccess,
   getCljsReplStartCode,
   getShadowCljsReplStartCode,
+  quickPickActive as quicPickActive,
```
corasaurus-hex: Can we get rid of the `as quicPickActive` here? We'll need to fix the references to it in the tests, too.

corasaurus-hex (approved):
I think this looks great! I love to see code improvements like this.

src/connector.ts
PEZ: I'll have to check and think. You are probably right. This has been somewhat cleaned up, but I might have missed this opportunity. Tomorrow, though, because right now too 💤 !

src/connector.ts
PEZ: Iirc i added the one in `connectCommand` because the user can do funny stuff with the menu. Here we don't allow that. But it could also go wrong because of non-validated glob patterns, so yes, we should have one here anyway. =)

src/utilities.ts
PEZ: No idea why this coercion is here... I'll ditch it!

PEZ:
> I love to see code improvements like this.

❤️ 

When I tried to extract the find-root-paths stuff to a module named `project-root.ts`, it turned out there was already such a module! And it had yet another implementation of the same traversing! Crazy. I just replaced it with the new stuff and updated the only call site. It seems to work. But also makes us glob more often, of course. Maybe we should keep the other way to find closest root around anyway...

docs/site/connect.md
```diff
@@ -61,6 +63,19 @@ All this said, I still recommend you challenge the conclusion that you can't use
 !!! Note
     There is a Calva command for copying the Jack-in command line to the clipboard.
 
+## Monorepos / multiple Clojure projects in one workspace
+
+If the workspace is a monorepo, Polylith repo or just a repository with more than one Clojure project, Calva will start the connect sequence with prompting for which project to start/connect to.
+
+![The project roots menu](images/calva-monorepo-project-roots-menu.png)
+
+When searching for project roots in your workspace, Calva will glob for all files matching `project.clj`, `deps.edn`, or `shadow-cljs.edn`. This is done using VS Code's workspace search engine, and is very efficient. However, in a large monorepo, it is still a sufficient task. In order to not waste resources Calva will exclude any directories in the setting `calva.projectRootsSearchExclude`.
```
bpringe: Did you mean to say "still a substantial task" here? It seems, judging by the context, that you mean to point out that it even though the search engine is very efficient, a large monorepo might still take a noticeable amount of time.

src/project-root.ts
```diff
@@ -1,54 +1,50 @@
...
-  return findProjectRootUri(projectFileNames, doc, workspaceFolder);
+export async function pickProjectRootPath(candidatePaths: string[], closestRootPath: string) {
+  const pickedRootPath =
+    candidatePaths.length < 2
+      ? undefined
+      : await util.quickPickSingle({
+          title: 'Project root',
+          values: candidatePaths,
+          default: closestRootPath,
+          placeHolder: 'Multiple Clojure projects found, pick one',
```
bpringe:
```suggestion
          placeHolder: 'Multiple Clojure projects found; pick one.',
```

bpringe (changes requested):

docs/site/connect.md
PEZ: Haha, yes. Thanks!

PEZ:
🎉

PR #1623, opened 2022-03-25
Fix warning when error loading file

### LINKED ISSUE #1567: Warning when loading a file produces an error: n.filter is not a function
To reproduce, while connected to a repl, put a symbol at the top level of the file that's not defined (like `a`, assuming that is not defined) and load the file. See the warning:

image

bpringe:
## What has Changed?

- Accesses the stacktrace array on an nrepl eval result correctly (before it was just accessing the stacktrace object, but not the array within it, which caused an error when another function called .filter on what it expected to be an array)
- Removes appending of the prompt from `printLastStacktrace` in favor of leaving the responsibility up to the caller and removing a responsibility from the function that isn't related to its main task. (This follows the single responsibility principle.)

Fixes #1567 

## My Calva PR Checklist

I have:

- [x] Tests

Ping @pez, @bpringe

### DIFF
--- a/src/evaluate.ts
+++ b/src/evaluate.ts
@@ -436,7 +436,7 @@ async function loadFile(
     } catch (e) {
       outputWindow.append(`; Evaluation of file ${fileName} failed: ${e}`);
       if (res.stacktrace) {
-        outputWindow.saveStacktrace(res.stacktrace);
+        outputWindow.saveStacktrace(res.stacktrace.stacktrace);
         outputWindow.printLastStacktrace();
       }
     }
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -229,8 +229,9 @@ async function activate(context: vscode.ExtensionContext) {
     vscode.commands.registerCommand('calva.selectCurrentForm', select.selectCurrentForm)
   );
   context.subscriptions.push(
-    vscode.commands.registerCommand('calva.loadFile', () => {
-      void eval.loadFile({}, config.getConfig().prettyPrintingOptions);
+    vscode.commands.registerCommand('calva.loadFile', async () => {
+      await eval.loadFile({}, config.getConfig().prettyPrintingOptions);
+      outputWindow.appendPrompt();
     })
   );
   context.subscriptions.push(
@@ -385,7 +386,10 @@ async function activate(context: vscode.ExtensionContext) {
     )
   );
   context.subscriptions.push(
-    vscode.commands.registerCommand('calva.printLastStacktrace', outputWindow.printLastStacktrace)
+    vscode.commands.registerCommand('calva.printLastStacktrace', () => {
+      outputWindow.printLastStacktrace();
+      outputWindow.appendPrompt();
+    })
   );
   context.subscriptions.push(
     vscode.commands.registerCommand(
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -229,11 +229,8 @@ async function activate(context: vscode.ExtensionContext) {
     vscode.commands.registerCommand('calva.selectCurrentForm', select.selectCurrentForm)
   );
   context.subscriptions.push(
-    vscode.commands.registerCommand('calva.loadFile', async () => {
-      await eval.loadFile({}, config.getConfig().prettyPrintingOptions);
-      return new Promise((finished) => {
-        outputWindow.appendPrompt(finished);
-      });
+    vscode.commands.registerCommand('calva.loadFile', () => {
+      void eval.loadFile({}, config.getConfig().prettyPrintingOptions);
     })
   );
   context.subscriptions.push(
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -231,7 +231,9 @@ async function activate(context: vscode.ExtensionContext) {
   context.subscriptions.push(
     vscode.commands.registerCommand('calva.loadFile', async () => {
       await eval.loadFile({}, config.getConfig().prettyPrintingOptions);
-      outputWindow.appendPrompt();
+      return new Promise((resolve) => {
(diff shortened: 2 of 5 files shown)


src/evaluate.ts
```diff
@@ -436,7 +436,7 @@ async function loadFile(
     } catch (e) {
       outputWindow.append(`; Evaluation of file ${fileName} failed: ${e}`);
       if (res.stacktrace) {
-        outputWindow.saveStacktrace(res.stacktrace);
+        outputWindow.saveStacktrace(res.stacktrace.stacktrace);
```
bpringe, line 439: The first stacktrace is an object, which contains a stacktrace property, which refers to an array.

src/extension.ts
```diff
@@ -229,11 +229,8 @@ async function activate(context: vscode.ExtensionContext) {
     vscode.commands.registerCommand('calva.selectCurrentForm', select.selectCurrentForm)
   );
   context.subscriptions.push(
-    vscode.commands.registerCommand('calva.loadFile', async () => {
-      await eval.loadFile({}, config.getConfig().prettyPrintingOptions);
-      return new Promise((finished) => {
-        outputWindow.appendPrompt(finished);
-      });
```
bpringe, line 236: The prompt is already appended when the file is loaded, so there's no need to do it here, which actually causes it to be printed twice.

bpringe:
@PEZ The stacktrace is printed in the output window when an error occurs when loading the file. Is that expected/intended? (Normally, stacktraces are not printed on error, but are saved for printing later if desired.)

bpringe:
@Cyrik Any idea why that test is failing? I tested jacking in and loading a file manually and it seems to work fine. It looks like maybe the file loading part of the test isn't run by the time it makes its assertion.

src/extension.ts
```diff
@@ -231,9 +231,7 @@ async function activate(context: vscode.ExtensionContext) {
   context.subscriptions.push(
     vscode.commands.registerCommand('calva.loadFile', async () => {
       await eval.loadFile({}, config.getConfig().prettyPrintingOptions);
-      return new Promise((finished) => {
-        outputWindow.appendPrompt(finished);
-      });
+      outputWindow.appendPrompt();
```
PEZ: Could this be the reason the integration tests are now flaky?

PEZ:
I also wonder what goes wrong with the integration test. The test results:

```
AssertionError [ERR_ASSERTION]: Expected values to be loosely deep-equal:

[
  '',
  'clj꞉test꞉> ',
  'nil',
  'bar',
  '; Evaluating file: test.clj'
]

should loosely deep-equal

[
  '',
  'clj꞉user꞉> ',
  '; Jack-in done.',
  'clj꞉user꞉>  ; Use `alt+enter` to evaluate',
  ';   - Clojure lines in stack traces are peekable and clickable.'
]
```


(We could consider not running a `reverse()` on the array before testing to make it a bit easier to map in your head to what s printed.)

Anyway. Running the integration tests locally I get:

```
AssertionError [ERR_ASSERTION]: Expected values to be loosely deep-equal:

[
  '',
  'clj꞉test꞉> ',
  'nil',
  'bar',
  '; Evaluating file: test.clj'
]

should loosely deep-equal

[
  '',
  '; Evaluating file: test.clj',
  'clj꞉user꞉> ',
  '; Jack-in done.',
  'clj꞉user꞉>  ; Use `alt+enter` to evaluate'
]
```

Which seems to indicate that it is a bit arbitrary when we read the results. But it used to not be flaky here, so probably something in the PR changes. I've commented on a change I suspect.

src/extension.ts
bpringe: I don't think this change itself really produced a behavioral change, but I could be wrong. If it did though, maybe we should update the test, as I the behavior seems to be what's desired.

bpringe: The Promise there is unnecessary, but I can add it back and try the tests again when I have time.

Cyrik: I don't think we should keep a promise just to make the test happy 😄 
But I think the promise might be useful outside the test as well. If you don't return that promise here, there is no way to wait for the file load + output window update to be done.

Cyrik:
@PEZ Yeah the missing promise means that we'll check the repl output before it has been written to.
The way I do the checks isn't ideal I think, but it feels like it's the closest thing to really checking what the user experiences.

src/extension.ts
PEZ: I agree about the value of being able to wait for the window updates to be done.

Disagree about not keeping the promise for testing purposes only. 😄

PEZ: How about:

```typescript
      // We use this promise in the E2E tests
      return new Promise((finished) => {
         outputWindow.appendPrompt(finished);
       });
```

Cyrik: I meant that I usually dislike having code that only exists for testing purposes. This promise is currently only used in tests, but it does make general sense for it to exist anyway. If I had a feature that wants to load a file and wait for that to be done, that promise is actually important. 
So I'm in favor of keeping it without the extra comment 😄

PEZ: And I meant that I usually like having code that exists only for testing. 😄 Testability is one of those altars I am ready to bring sacrifice to.

bpringe: > I don't think we should keep a promise just to make the test happy

I wasn't suggesting that - I just wanted to test and see if it was affecting the test for some reason.

The function inherently returns a promise since it's async, and if you await it, the promise is resolved after the `await loadFile` execution completes. There's no need to return a separate promise like that.

bpringe: Wow, somehow I missed that the promise was actually being resolved after the prompt was appended in the output window 😑 . I added it back as I agree it could be useful besides just for testing.

PR #1646, opened 2022-03-29
Update expected result

phreed:
The evaluate cursor example had an incorrect expected result.


PEZ (approved):
Thanks for improving Calva! 🙏 ❤️

PR #1666, opened 2022-04-06
Provide clojure-lsp with fallback rootUri and defer starting until we have a Clojure file

### LINKED ISSUE #1601: clojure-lsp does not start in the Getting Started REPL

### LINKED ISSUE #1664: clojure-lsp initialized in a non-working state when there is no VS Code folder
VS Code assumes it has a folder open when starting the LSP client. Otherwise it initializes the server with a `null` for `rootUri`. And this leaves clojure-lsp in a non-working state.

This can happen with the standalone REPLs which as a feature can be started standalone from a VS Code folder (see #1601). It can also happen if the user opens Clojure file, or creates a Clojure document. Or, in fact, any of the actions where Calva is configured to activate.

This gets a bit extra bad for the user because the main options for diagnostics we provide **LSP serverInfo**, and **Open LSP log** both crash inside clojure-lsp when the rootUri is not set. And as Calva uses **serverInfo** as part of the activation of clojure-lsp things are a bit in limbo here.

My analysis is that we are starting clojure-lsp prematurely, we need to know the rootUri, or that there is a VS Code folder open, before we start clojure-lsp.

PEZ:
## What has Changed?

VS Code sends `rootUri: null` to the LSP server if there is no folder open in the window. This doesn't work for clojure-lsp. So now, Instead of just blindly starting the server when Calva activates:

* We check if there is a folder open in the window
    * If so we proceed as before
    * If not:
        * We check if we have a Clojure file
           * If we do we start the clojure-lsp server
           * If we don't we don't start it
    * When staring the server
       * We do our best to figure out a fallback rootUri from the file, falling back to a generic fallback if it is untitled.
       * Start the server providing this fallback

The above approach alone would result in no clojure-lsp when starting standalone REPLs. So what we do there is that we: 

1.  wait until we have prepared the ”project” for the standalone REP
2. then we start clojure-lsp.

Fixes #1664 
Fixes #1601

## My Calva PR Checklist

I have:

- [ ] Tests

Ping @pez, @bpringe, @corasaurus-hex, @Cyrik, @ericdallo

### DIFF
--- a/src/lsp/main.ts
+++ b/src/lsp/main.ts
@@ -19,7 +19,6 @@ import { setStateValue, getStateValue } from '../../out/cljs-lib/cljs-lib';
 import { downloadClojureLsp, getLatestVersion } from './download';
 import { readVersionFile, getClojureLspPath } from './utilities';
 import { TestTreeHandler, TestTreeParams } from './types';
-import * as os from 'os';
 import * as path from 'path';
 import * as state from '../state';
 import { provideHover } from '../providers/hover';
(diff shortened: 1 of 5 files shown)


ericdallo:
From my point of view that logic seems to make sense!

PEZ:
@ericdallo thanks for quick help!

The only ”failing” case I've found during my testing is that we don't get any clojure-lsp help on documents which are not on disk, VS Code calls these ”Untitled". But that is regardless if we have a folder/rootUri or not, and actually unrelated to this issue/PR. But anyway. more for the record of things.

ericdallo:
Hum, not sure how hard would be to make that happen, but clojure-lsp should not expect file is on disk indeed

PEZ:
With nREPL, the client sends the file contents. Maybe we can do the same here? Subject for another ticket, though. 😄

Cyrik:
We already send the file contents on open and on change. It's most likely the filename that's making problems

PEZ:
>  It's most likely the filename that's making problems

The file extension?

src/lsp/main.ts
```diff
@@ -350,10 +365,85 @@ async function stopClient() {
...
+        }
+      }
+    }
+  }
+
+  if (!clojureFilePath) {
+    if (!folderType) {
+      folderType = FolderType.FROM_NO_CLOJURE_FILE;
+    }
+    clojureFilePath = path.join(util.calvaTmpDir(), 'clojure-lsp-fallback-root', 'Untitled');
```
Cyrik, line 421: Have you tried calling it untitled.clj? That might make it work in lsp

Cyrik:
Yeah, I think lsp ignores anything that doesn't have the right extension

src/lsp/main.ts
PEZ: I'm not naming the file here. I'm creating a path on which I can do `dirname`. Maybe needs a clarifying comment...

PEZ:
Untitled files have no extension, by defnition. Seems we can't avoid discussing this off-topic thing here. My fault, for starting it. 😄

PEZ:
I created an issue to discuss the Untitled problem now:

* #1668

PR #1698, opened 2022-04-30
Create clojure-lsp drag commands

### LINKED ISSUE #1697: Enable users to bind keyboard shortcuts to clojure-lsp drag commands
The clojure-lsp version of the drag commands offer some improvements over the built-in drag commands. See #1695 for a detailed explanation of how the clojure-lsp commands work.

This is a request to allow users to bind keyboard shortcuts to the clojure-lsp drag commands.

I'll contribute a PR, but this is my first contribution to Calva, so I may need some guidance. Thanks in advance!

mainej:
## What has Changed?

This is the start of exposing the clojure-lsp `drag-backward` and `drag-forward` commands as targets for keybindings. This was discussed in #1695.

These commands are similar to the built-in Calva commands `dragSexprForward` and `dragSexprBackward`. But users may want to use the clojure-lsp versions to take advantage of additional features that clojure-lsp offers:

- Better comment affinity. Comments stay affiliated with code blocks as they are dragged.
- Better contextual support. Clauses within many common forms (e.g. `case` and `cond`) move intuitively.

So far, I've been unable to test that this works as intended, probably because I'm new to Calva development.

The `drag-forward` and `drag-backward` commands appear in the keybindings menu, but:

1. They don't have any of the details defined in package.json. For example, I'd expect `drag-forward` to be named "Calva refactoring: Drag Sexpr Forward", but it isn't.
2. If I bind keys to these commands, the keys don't do anything. This may also be because the package.json info isn't picked up.

Even after we get the above solved, I'm a bit worried about whether the commands will work as intended. You can already invoke the commands from the clojure-lsp action menu, where they appear as 'Drag forward' or 'Drag backward'. When I do this, the code is edited as intended, but the cursor isn't moved to the correct location. Getting down into the LSP weeds, clojure-lsp sends a notification to the editor asking it to reposition the cursor after these edits. The notification is async... it happens not as part of the refactoring (in LSP terms, `workspace/executeCommand`) but later, after the refactoring has been completed (`window/showDocument`). It surprises me that this doesn't work, as I assume Calva is using the built-in LSP client tooling. Is there something special about Calva that would make it not respond to the `window/showDocument` notification?

When all of this is working, it will resolve #1697.

## My Calva PR Checklist

I have:

- [ ] Tests
     - ~If I am fixing just part of the issue, I have just referenced it w/o any of the "fixes” keywords.~

### DIFF
--- a/package.json
+++ b/package.json
@@ -1630,103 +1630,103 @@
         "category": "Calva"
       },
       {
-        "command": "calva.refactor.cleanNs",
+        "command": "clojureLsp.refactor.cleanNs",
         "title": "Clean NS Form",
         "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
-        "command": "calva.refactor.addMissingLibspec",
+        "command": "clojureLsp.refactor.addMissingLibspec",
         "title": "Add Missing Require",
         "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
-        "command": "calva.refactor.dragBackward",
+        "command": "clojureLsp.dragBackward",
         "title": "Drag Sexp Backward",
         "category": "clojure-lsp",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
-        "command": "calva.refactor.dragForward",
+        "command": "clojureLsp.dragForward",
         "title": "Drag Sexp Forward",
         "category": "clojure-lsp",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
-        "command": "calva.refactor.cyclePrivacy",
+        "command": "clojureLsp.refactor.cyclePrivacy",
         "title": "Cycle/Toggle Privacy",
         "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
-        "command": "calva.refactor.expandLet",
+        "command": "clojureLsp.refactor.expandLet",
         "title": "Expand Let",
         "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
-        "command": "calva.refactor.inlineSymbol",
+        "command": "clojureLsp.refactor.inlineSymbol",
         "title": "Inline Symbol",
         "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
-        "command": "calva.refactor.threadFirst",
+        "command": "clojureLsp.refactor.threadFirst",
         "title": "Thread First",
         "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
-        "command": "calva.refactor.threadFirstAll",
+        "command": "clojureLsp.refactor.threadFirstAll",
         "title": "Thread First All",
         "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
-        "command": "calva.refactor.threadLast",
+        "command": "clojureLsp.refactor.threadLast",
         "title": "Thread Last",
         "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
-        "command": "calva.refactor.threadLastAll",
+        "command": "clojureLsp.refactor.threadLastAll",
         "title": "Thread Last All",
         "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
-        "command": "calva.refactor.unwindAll",
+        "command": "clojureLsp.refactor.unwindAll",
         "title": "Unwind All",
         "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
-        "command": "calva.refactor.unwindThread",
+        "command": "clojureLsp.refactor.unwindThread",
         "title": "Unwind Thread",
         "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
-        "command": "calva.refactor.introduceLet",
+        "command": "clojureLsp.refactor.introduceLet",
         "title": "Introduce let",
         "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
-        "command": "calva.refactor.moveToLet",
+        "command": "clojureLsp.refactor.moveToLet",
         "title": "Move to Previous let Box",
         "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
-        "command": "calva.refactor.extractFunction",
+        "command": "clojureLsp.refactor.extractFunction",
         "title": "Extract to New Function",
         "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
-        "command": "calva.diagnostics.clojureLspServerInfo",
+        "command": "clojureLsp.diagnostics.clojureLspServerInfo",
         "title": "Clojure-lsp Server Info",
         "category": "clojure-lsp Diagnostics",
         "enablement": "editorLangId == clojure && clojureLsp:active"
--- a/package.json
+++ b/package.json
@@ -1642,6 +1642,18 @@
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
+        "command": "calva.refactor.dragBackward",
+        "title": "Drag Sexp Backward",
+        "category": "Calva Refactor",
+        "enablement": "editorLangId == clojure && clojureLsp:active"
+      },
+      {
+        "command": "calva.refactor.dragForward",
+        "title": "Drag Sexp Forward",
+        "category": "Calva Refactor",
+        "enablement": "editorLangId == clojure && clojureLsp:active"
+      },
+      {
         "command": "calva.refactor.cyclePrivacy",
         "title": "Cycle/Toggle Privacy",
         "category": "Calva Refactor",
--- a/package.json
+++ b/package.json
@@ -1632,103 +1632,103 @@
       {
         "command": "calva.refactor.cleanNs",
         "title": "Clean NS Form",
-        "category": "Calva Refactor",
+        "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
         "command": "calva.refactor.addMissingLibspec",
         "title": "Add Missing Require",
-        "category": "Calva Refactor",
+        "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
         "command": "calva.refactor.dragBackward",
         "title": "Drag Sexp Backward",
-        "category": "Calva Refactor",
+        "category": "clojure-lsp",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
         "command": "calva.refactor.dragForward",
         "title": "Drag Sexp Forward",
-        "category": "Calva Refactor",
+        "category": "clojure-lsp",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
         "command": "calva.refactor.cyclePrivacy",
         "title": "Cycle/Toggle Privacy",
-        "category": "Calva Refactor",
+        "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
         "command": "calva.refactor.expandLet",
         "title": "Expand Let",
-        "category": "Calva Refactor",
+        "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
         "command": "calva.refactor.inlineSymbol",
         "title": "Inline Symbol",
-        "category": "Calva Refactor",
+        "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
         "command": "calva.refactor.threadFirst",
         "title": "Thread First",
-        "category": "Calva Refactor",
+        "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
         "command": "calva.refactor.threadFirstAll",
         "title": "Thread First All",
-        "category": "Calva Refactor",
+        "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
         "command": "calva.refactor.threadLast",
         "title": "Thread Last",
-        "category": "Calva Refactor",
+        "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
         "command": "calva.refactor.threadLastAll",
         "title": "Thread Last All",
-        "category": "Calva Refactor",
+        "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
         "command": "calva.refactor.unwindAll",
         "title": "Unwind All",
-        "category": "Calva Refactor",
+        "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
         "command": "calva.refactor.unwindThread",
         "title": "Unwind Thread",
-        "category": "Calva Refactor",
+        "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
         "command": "calva.refactor.introduceLet",
         "title": "Introduce let",
-        "category": "Calva Refactor",
+        "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
         "command": "calva.refactor.moveToLet",
         "title": "Move to Previous let Box",
-        "category": "Calva Refactor",
+        "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
         "command": "calva.refactor.extractFunction",
         "title": "Extract to New Function",
-        "category": "Calva Refactor",
+        "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
         "command": "calva.diagnostics.clojureLspServerInfo",
         "title": "Clojure-lsp Server Info",
-        "category": "Calva Diagnostics",
+        "category": "clojure-lsp Diagnostics",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Changes to Calva.
 ## [Unreleased]
 - Fix: [Toggle between implementation and test command should set editor focus](https://github.com/BetterThanTomorrow/calva/issues/1707)
 - Fix: [Hijacked shortcuts only works on strict](https://github.com/BetterThanTomorrow/calva/issues/1711)
+- [Enable users to bind keyboard shortcuts to clojure-lsp drag commands](https://github.com/BetterThanTomorrow/calva/issues/1697)
 
 ## [2.0.270] - 2022-05-02
 - Fix: [Downloaded clojure-lsp not working for static linux distros](https://github.com/BetterThanTomorrow/calva/issues/1692)
--- a/docs/site/clojure-lsp.md
+++ b/docs/site/clojure-lsp.md
@@ -114,3 +114,4 @@ See also:
 
 * [Connecting the REPL](connect.md)
 * [Refactoring](refactoring.md)
+* [Dragging forms](paredit.md#drag-bindings-forwardbackward)
--- a/docs/site/paredit.md
+++ b/docs/site/paredit.md
@@ -135,6 +135,9 @@ And like so (wait for it):
 
 ![](images/paredit/drag-pairs-in-maps.gif)
 
+!!! Note
+    clojure-lsp also contributes commands for dragging forms backward and forward. You might prefer those, because they have better comment affinity—comments stay affiliated with clauses as they're dragged—and better contextual support—clauses within many common functions (e.g. assoc, case and cond) move intuitively. Note however that these commands won't work in the REPL and other places where clojure-lsp is unavailable. See [this PR](https://github.com/BetterThanTomorrow/calva/pull/1698) for more details. To use the clojure-lsp commands, search for "clojure-lsp drag" in the command palette or the keyboard shortcut preferences menu, or choose them from the _Quick Fix_ suggestion lightbulb.
+
 ## About the Keyboard Shortcuts
 
 Care has been put in to making the default keybindings somewhat logical, easy to use, and work with most keyboard layouts. Slurp and barf forward are extra accessible to go with the recommendation to learn using these two super handy editing commands.
--- a/docs/site/paredit.md
+++ b/docs/site/paredit.md
@@ -136,7 +136,7 @@ And like so (wait for it):
 ![](images/paredit/drag-pairs-in-maps.gif)
 
 !!! Note
-    clojure-lsp also contributes commands for dragging forms backward and forward. You might prefer those, because they have better comment affinity—comments stay affiliated with clauses as they're dragged—and better contextual support—clauses within many common functions (e.g. assoc, case and cond) move intuitively. Note however that these commands won't work in the REPL and other places where clojure-lsp is unavailable. See [this PR](https://github.com/BetterThanTomorrow/calva/pull/1698) for more details. To use the clojure-lsp commands, search for "clojure-lsp drag" in the command palette or the keyboard shortcut preferences menu, or choose them from the _Quick Fix_ suggestion lightbulb.
+    clojure-lsp also contributes commands for dragging forms backward and forward. You might prefer those, because they have better comment affinity—comments stay affiliated with clauses as they're dragged—and better contextual support—clauses within many common functions (e.g. assoc, case and cond) move intuitively. Note however that these commands won't work in the REPL and other places where clojure-lsp is unavailable. Unlike Calva's drag commands, clojure-lsp's drag commands will not move words within strings. To use the clojure-lsp commands, search for "clojure-lsp drag" in the command palette or the keyboard shortcut preferences menu, or choose them from the _Quick Fix_ suggestion lightbulb. See [this PR](https://github.com/BetterThanTomorrow/calva/pull/1698) and its links for more discussion of this feature.
 
 ## About the Keyboard Shortcuts
 
--- a/docs/site/refactoring.md
+++ b/docs/site/refactoring.md
@@ -5,32 +5,32 @@ description: Powered by clojure-lsp, Calva comes with some quite nifty refactori
 
 # Refactoring
 
-There are two ”flavours” to refactoring support. Some (just a few) refactorings are made available as _Quick Fix_ suggestions (the light bulb), the rest are regular commands in the *Calva Refactoring* category.
+There are two ”flavours” to refactoring support. Some (just a few) refactorings are made available as _Quick Fix_ suggestions (the light bulb), the rest are regular commands in the *clojure-lsp Refactoring* category.
 
 ![](images/refactoring/quick-fix.png)
 
 You can enable or disable the _Quick Fix_ suggestion lightbulb using the VS Code setting `editor.lightbulb.enabled`.
 
-The refactoring commands do not have default keyboard shortcuts. You find them all by typing ”Calva Refactor” in the Command Palette.
+The refactoring commands do not have default keyboard shortcuts. You find them all by typing ”clojure-lsp Refactor” in the Command Palette.
 
 ## Commands
 
 Command Title | Command Key | Description
 ------------- | ----------- | -----------
-Clean NS Form | `calva.refactor.cleanNs` | ![](images/refactoring/cleanNs.gif)
-Add Missing Require | `calva.refactor.addMissingLibspec` | ![](images/refactoring/addMissingLibspec.gif)
-Extract to New Function | `calva.refactor.extractFunction` | ![](images/refactoring/extractFunction.gif)
-Cycle/Toggle Privacy | `calva.refactor.cyclePrivacy` | ![](images/refactoring/cyclePrivacy.gif)
-Inline Symbol | `calva.refactor.inlineSymbol` | ![](images/refactoring/inlineSymbol.gif)
-Introduce let | `calva.refactor.introduceLet` | Creates a new let box with the binding. Follow up with ”Expand let” to move it upwards.<br>![](images/refactoring/introduceLet.gif)
-Expand Let | `calva.refactor.expandLet` | ![](images/refactoring/expandLet.gif)
-Move to Previous let Box | `calva.refactor.moveToLet` | ![](images/refactoring/moveToLet.gif)
-Thread First | `calva.refactor.threadFirst` | ![](images/refactoring/threadFirst.gif)
-Thread First All | `calva.refactor.threadFirstAll` | ![](images/refactoring/threadFirstAll.gif)
-Thread Last | `calva.refactor.threadLast` | ![](images/refactoring/threadLast.gif)
-Thread Last All | `calva.refactor.threadLastAll` | ![](images/refactoring/threadLastAll.gif)
-Unwind All | `calva.refactor.unwindAll` | ![](images/refactoring/unwindAll.gif)
-Unwind Thread | `calva.refactor.unwindThread` | ![](images/refactoring/unwindThread.gif)
+Clean NS Form | `clojureLsp.refactor.cleanNs` | ![](images/refactoring/cleanNs.gif)
+Add Missing Require | `clojureLsp.refactor.addMissingLibspec` | ![](images/refactoring/addMissingLibspec.gif)
+Extract to New Function | `clojureLsp.refactor.extractFunction` | ![](images/refactoring/extractFunction.gif)
+Cycle/Toggle Privacy | `clojureLsp.refactor.cyclePrivacy` | ![](images/refactoring/cyclePrivacy.gif)
+Inline Symbol | `clojureLsp.refactor.inlineSymbol` | ![](images/refactoring/inlineSymbol.gif)
+Introduce let | `clojureLsp.refactor.introduceLet` | Creates a new let box with the binding. Follow up with ”Expand let” to move it upwards.<br>![](images/refactoring/introduceLet.gif)
+Expand Let | `clojureLsp.refactor.expandLet` | ![](images/refactoring/expandLet.gif)
+Move to Previous let Box | `clojureLsp.refactor.moveToLet` | ![](images/refactoring/moveToLet.gif)
+Thread First | `clojureLsp.refactor.threadFirst` | ![](images/refactoring/threadFirst.gif)
+Thread First All | `clojureLsp.refactor.threadFirstAll` | ![](images/refactoring/threadFirstAll.gif)
+Thread Last | `clojureLsp.refactor.threadLast` | ![](images/refactoring/threadLast.gif)
+Thread Last All | `clojureLsp.refactor.threadLastAll` | ![](images/refactoring/threadLastAll.gif)
+Unwind All | `clojureLsp.refactor.unwindAll` | ![](images/refactoring/unwindAll.gif)
+Unwind Thread | `clojureLsp.refactor.unwindThread` | ![](images/refactoring/unwindThread.gif)
 
 !!! Note "Formatting"
     The way that some of the refactorings are applied to the document, makes it difficult for Calva to format the results. So, sometimes you'll need to navigate the cursor to the enclosing form and hit `tab` to tidy up the formatting after a refactoring. See also [Formatting](formatting.md).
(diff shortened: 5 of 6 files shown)


mainej:
As a bit more information about what I've tried for testing... As described in the [guide](https://github.com/BetterThanTomorrow/calva/wiki/How-to-Hack-on-Calva), I've ran the build task, started the extension host, and connected a REPL. The REPL doesn't really work (it wants a browser to connect to—do you have to open a browser tab?) but I'm not sure that's critical to the flow. I'm not sure what my next step should be—please advise.

PEZ:
Thanks! 🙏 I appreciate you helping us with Calva. 

Looks like you're having progress. Some quite odd problems there, but I'm sure we'll figure it out.

PEZ:
I'll run this on my machine and see what I can make of it.

You don't actually need to connect the REPL for this task, but it is strange, and could be some symptom, with the error you get. To answer your question, there is not browser involved. Did you really choose the project type `Calva` when prompted?

PEZ:
> 1. They don't have any of the details defined in package.json. For example, I'd expect drag-forward to be named "Calva refactoring: Drag Sexpr Forward", but it isn't.

Strange, might be some caching issue. I see the named commands as we expect...

> 2. If I bind keys to these commands, the keys don't do anything. This may also be because the package.json info isn't picked up.

I can bind the named commands to keys The commands behave the same using the key shortcuts as from the fix-menu. Same observation as you do, the drag happens, but the cursor is misplaced.

Also, it seems the whole document is rererendered a bit funny, because I get more of clojure-lsp's semantic tokens flicker with these commands than with Calva's own versions.

Here's a recording:

https://user-images.githubusercontent.com/30010/166097180-96fefc3d-a383-44fc-ae5c-30822886b47f.mp4

We really need to figure out about the cursor. 😄 

But now I am stomped. Not particularly savvy with the lsp protocol. What is `showDocument`? 


Side note: Also seems a bit like it could break to do the cursor placement async after the edit. What if the user has issued more edit commands then? But maybe I am just not understanding how it's supposed to work.

package.json
```diff
@@ -1641,6 +1641,18 @@
...
+        "command": "calva.refactor.dragBackward",
+        "title": "Drag Sexp Backward",
+        "category": "Calva Refactor",
+        "enablement": "editorLangId == clojure && clojureLsp:active"
+      },
+      {
+        "command": "calva.refactor.dragForward",
+        "title": "Drag Sexp Forward",
+        "category": "Calva Refactor",
+        "enablement": "editorLangId == clojure && clojureLsp:active"
```
PEZ: You've probably thought of this too. These are not refactors, in any meaningful sense of the word. We should name these something else. Maybe Instead of ”Calva Refactor”, we can use "clojure-lsp". That should make the option clearer about which system is handling it.

While at it we could recategorize the other ”Calva Refactor” commands to say ”clojure-lsp Refactor”.

PEZ (changes requested):
Cool that it so far is this straightforward!

We'll need to update the `registerLspCommand` function some.

Cyrik:
Welcome to this side of the lsp divide @mainej! 👋 

> * They don't have any of the details defined in package.json. For example, I'd expect `drag-forward` to be named "Calva refactoring: Drag Sexpr Forward", but it isn't.
> * If I bind keys to these commands, the keys don't do anything. This may also be because the package.json info isn't picked up.

This probably happened because you have the compile watcher running, but updated the project.json. As far as I can tell, the watcher doesn't pick up project.json changes, which means you need to restart it when you change something there. Let me know if that helps or if it's still broken.

> ... I assume Calva is using the built-in LSP client tooling.

You are correct, we're using the node lsp client library to do most of the heavy lifting and only have a few places where we take away its control (completion, definition, hover, etc.). So most of the time, lsp features "just work" without Calva adding anything. Sadly this also means that it's often annoying to debug when something doesn't work, because the lsp client is written in such a generic way.
When I run into those situations I usually add debug stops into the lsp-client to see what it does with the incoming lsp notifications. I don't have time now, but I can post a few useful breakpoints tonight, if you're still struggling. I'm also happy to help with anything directly on slack, if that's easier.

PEZ:
> This probably happened because you have the compile watcher running, but updated the project.json. As far as I can tell, the watcher doesn't pick up project.json changes, which means you need to restart it when you change something there. Let me know if that helps or if it's still broken.

Not my experience at all. The manifest i completely static. VS Code picks it up when restarting the dev extension host. Restarting the watcher/recompiling the code makes no difference.

That said, I don't understand what's going wrong for @mainej. Just that this particular source of error shouldn't be it. If restarting the watcher helps for such issues for you. @Cyrik, I think something else is broken in your setup.

package.json
ericdallo: I like it

ericdallo:
> Also, it seems the whole document is rererendered a bit funny, because I get more of clojure-lsp's semantic tokens flicker with these commands than with Calva's own versions.

Semantic tokens are calculated after changes to the document, it should not behave differently between namnually changing the document and using a command, both will trigger it the same way.

> We really need to figure out about the cursor. smile

Unfortunately, I think this is happening for all editors as well, probably something we would need to make it work somehow on clojure-lsp side, right @mainej ?

> But now I am stomped. Not particularly savvy with the lsp protocol. What is showDocument?

Right, [window/showDocument](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showDocument) it's a way a server can tell to client to show a particular file/buffer with cursor on a specific location, this is useful for commands/refactors that need to move cursor or even move cursor to other buffer like `create-test` refactor.
This is a request, so clojure-lsp waits for client to move to then continue processing next actions, not sure about the async but not sure would be that common this kind of issue as the execution of a command like this is pretty fast, I can't even make it doesn't work if I try to move the cursor after executed the command, so not sure we should worry about that for now.
About not working on Calva, one thing to check is that client needs to tell it supports that when initializating the server via a `window.showDocument` client capability, maybe check if vscode sends that as true

mainej:
Wow, thanks for all the feedback and pointers everyone! I've tried re-starting the extension host. If that's what's supposed to pick up the contents of package.json, my guess is somehow that's not actually working. How can I check?

> might be some caching issue

Sounds likely! Where are the caches? 

To ensure we're all on the same page about what I'm seeing, here's a screenshot of my keybindings config options. As you can see `drag-forward` is present), which I think must be coming from the change to `b/src/lsp/main.ts`, where those are defined as `clojureLspCommands`. But, I'm not seeing the "Calva Refactoring: Drag Sexp Forward" that you can see in @PEZ's screen recording, which comes from package.json.

drag-keybindings

Here's a question... in @PEZ's screen recording, the "source" of the new commands is User, whereas the source of the paredit commands is Extension. Is that a hint to what's going on? What mechanism defines User commands?

> semantic tokens flicker

Here's my guess... suppose you drag within a map. clojure-lsp instructs the client to replace the entire map, even though only two clauses are changing. I'd guess that when that happens, VS Code invalidates the semantic tokens for the whole map, requests a semantic token range from clojure-lsp, and waits for the response to fix the coloring of the entire map. On the other hand, I'd guess that Calva's internal implementation of drag is narrower... it replaces only the two clauses that are moving, meaning the editor can hold onto the semantic tokens for the non-moving clauses while it waits for the replacement of the semantic tokens of the moving clauses, leading to less flicker. I'd classify this as a defect in the clojure-lsp implementation—it really should edit only the clauses that are moving. I'm not actually sure how hard it would be to fix, but it's something I can look into. If it can be fixed, it should benefit all editors.

> What is showDocument?

@ericdallo 's explanation is great, and points out some flaws in my original description of what's happening. I had said that showDocument is a server notification (a fire-and-forget thing) whereas it's actually a server request (which waits for a response from the client). @ericdallo do you understand the lsp4j threading well? Does the server stop all of its activity while it waits for the client to confirm receipt of the showDocument request? Or does it keep processing other client requests?

In any case, the flow looks like this

```
--- synchronously ---
lsp client-request: executeCommand -> server calculates drag edit
                                   <- lsp server-response: edits
client applies edits

--- moments later, started asynchronously but performed synchronously ---
                                           <- lsp server-request: showDocument
client moves cursor (in theory)
lsp client-response: ok                    ->
```

Anyway, something isn't happening correctly here. In Emacs it's possible to inspect the protocol level logs. That is, you can turn on logs of the actual messages going across the wire. These are separate from the language server's logs about what it's doing. Does VS Code have a way to inspect the wire messages?

>  I usually add debug stops into the lsp-client to see what it does with the incoming lsp notifications. I don't have time now, but I can post a few useful breakpoints tonight, if you're still struggling. I'm also happy to help with anything directly on slack, if that's easier.

Sounds tricky, but this is probably the right path. We're deep into Calva dev stuff that I don't understand. I'd need pretty detailed instructions to understand how to "add debug stops into the lsp-client".

> Side note: Also seems a bit like it could break to do the cursor placement async after the edit

I've wondered about this too... it hasn't been a problem in Emacs and nvim. But it might be why things aren't working in Calva. This is total speculation, so take with a large grain of salt... If between applying the edits and receiving the request to position the cursor, perhaps VS Code is deciding to do its own re-positioning of the cursor. And perhaps it decides that the clojure-lsp's request to position the cursor is "old", and that therefore it decides to ignore it.

>> We really need to figure out about the cursor.

> Unfortunately, I think this is happening for all editors as well, probably something we would need to make it work somehow on clojure-lsp side

@ericdallo, no, this isn't a problem in other editors. If you watch @PEZ's recording, in the top section he's using clojure-lsp drag and in the bottom section he's using calva drag. In other editors, if you use make the same edits PEZ is making with clojure-lsp, the cursor is re-positioned to the start of the moved clause, whereas in Calva, it's left in the middle of the clause. (In PEZ's recording, it doesn't really matter that the cursor is left in the middle; the correct clauses are still moved. But if he had been moving multi-line clauses, things would quickly get whacky, as the cursor would end up outside of the clause that had just been moved.) In any case, this implies the problem isn't on the clojure-lsp side, or at least not in a way that's a problem for other editors.

> one thing to check is that client needs to tell it supports that when initializating the server via a window.showDocument client capability

Good point! I'll confirm whether VS Code sends this capability. If it doesn't, it'd explain why showDocument isn't working.

ericdallo:
>  Does the server stop all of its activity while it waits for the client to confirm receipt of the showDocument request? Or does it keep processing other client requests?

AFAIK, the client should only send new requests when already processed existing requested by server, not sure tho, we could check on the logs between client and server I think.

> Does VS Code have a way to inspect the wire messages?

Yes, check [here](https://calva.io/clojure-lsp/) "Viewing the Logs Between the Client and Server"

>  In any case, this implies the problem isn't on the clojure-lsp side, or at least not in a way that's a problem for other editors.

That makes sense, thanks

PEZ:
Very strange with the command registration on your machine, @mainej . This is what I see:

image

> If you watch @PEZ's recording, in the top section he's using clojure-lsp drag and in the bottom section he's using calva drag.

Not really. 😄 To clarify what's shown in my recording:

1. Calva Paredit  drag the pairs in the map
2. Calva Paredit drag bindings in the let box
3. clojure-lsp  drag the pairs in the map
4. clojure-lsp drag bindings in the let box

In both **3** & **4** the flicker is happening. In **4** it also is shown that the cursor misplacement causes problems, because all of a sudden I am not dragging the let binding, but instead stuff inside the `(if ...)` form. This does not happen with **2**.

mainej:
Making [progress](https://github.com/clojure-lsp/clojure-lsp/pull/955) on the clojure-lsp side. In the meantime, some follow-up...

> We'll need to update the registerLspCommand function some

What needs to be changed @PEZ? Or does it seem ok as is?

> These are not refactors, in any meaningful sense of the word. We should name these something else. Maybe Instead of ”Calva Refactor”, we can use "clojure-lsp". That should make the option clearer about which system is handling it.

> While at it we could recategorize the other ”Calva Refactor” commands to say ”clojure-lsp Refactor”.

I'm not quite sure I understand... Do you want everything to be named "clojure-lsp Refactor: ..."? Or most things, with the exception of drag which should be named "clojure-lsp: ..." since it's not really a refactor?

PEZ:
> Or most things [be named ”clojure-lsp Refactor”], with the exception of drag which should be named "clojure-lsp: ..." since it's not really a refactor?

This interpretation of my ramblings is the correct one. 😄

PEZ:
> We'll need to update the `registerLspCommand` function some
>> What needs to be changed @PEZ? Or does it seem ok as is?

I was referring to the renaming. The function looks a bit hardcoded on the current naming scheme.

PEZ:
**Rationale behind my naming suggestion**: It's a bit tricky with these names... Since Calva is sourcing clojure-lsp power, it could be argued that they should be prefixed with `Calva` to make it clear which extension brings the command in. But it is more important, I think, that it is clear that clojure-lsp contributes them. Then people can easier figure out where to reach for support and how to phrase their questions. If they reach out in Calva channels, they can say ”This clojure-lsp command, ...” And if they reach out in clojure-lsp channels: ”I use this command in Calva ...”. 

The risk is then that they don't find the commands because they search for `Calva` in the command palette. But it gets verbose covering all bases. We can instruct users in documentation and such to look for commands in both ”namespaces”.

mainej:
@PEZ I've fixed the merge conflicts (I had to force push... sorry if that breaks any branches you had) and edited the titles of the clojure-lsp commands. Will you double check that they're what you were expecting? Also, do you want to rename the commands themselves, e.g. `calva.refactor.moveToLet` -> `???.moveToLet`? I worry that would break everyone's existing keybindings.

package.json
```diff
@@ -1630,93 +1630,105 @@
...
-        "command": "calva.refactor.addMissingLibspec",
+        "command": "clojureLsp.refactor.addMissingLibspec",
         "title": "Add Missing Require",
-        "category": "Calva Refactor",
+        "category": "clojure-lsp Refactor",
         "enablement": "editorLangId == clojure && clojureLsp:active"
       },
       {
-        "command": "calva.refactor.cyclePrivacy",
+        "command": "clojureLsp.dragBackward",
```
PEZ, line 1645: I'd like it to be the same name as the calva counterpart, only with a different prefix/namespace.

```suggestion
        "command": "clojureLsp.dragSexprBackward",
```

PEZ (changes requested):
Thanks again for your help with this, @mainej !

I've left a comment about the naming.

PEZ:
This all behaves much better now. From the behaviour it should be good to merge, I think. Though there are some issues. Pinging in some help in assessing: @bpringe @corasaurus-hex @Cyrik @riotrah This is also your call, @mainej , but I think it is better to have the feature in place and then fix whatever feels most important to fix. We can mention the WIP nature in the release notes. As I understand things some of these will be able to fix from the clojure-lsp side, and some we will just have to accept for now.

Here's a recording showing some of what I am talking about when I talk about issues. These are just observations. Some of the issues probably have a common root cause. I have made no attempts to speculate in causes at all.  

https://user-images.githubusercontent.com/30010/166653916-87ef4dcd-ae72-4982-b624-ceabaceb7464.mp4

As always it is hard to follow along in a video like this. Here's somekind of caption that might help:

1. Sometimes does not drag...
1. Does not retain cursor position. Moves the cursor to the start of the dragged text.
   1. Sometimes misplaces the cursor, I can't immediately see a pattern to it.
1. Drags the sexpr in front when adjacent behind of a form. And generally does not adhere to Calva's [Current Form](https://www.youtube.com/watch?v=8ygw7LLLU1w) concept at all.
1. Does not drag words inside strings
1. The window scrolls around in a surprising way.
1. Significant clojure-lsp semantic tokens lag/flicker. Especially in large files.
1. Performance drop in large files. Nothing surprising. But it is more prominent than with Calva's versions of the commands. I'm using an Apple M1 Max, haven't tested om less powerful hardware.

**2** & **3** in combination makes for some extra weirdness.

To me the differences in what is the dragged form is a big concern. I'd like to wish for that Calva's concept of this is adapted by clojure-lsp. It will cause less cognitive load for Calva users. And since Calva's concept here is somewhat compatible with CIDER's (Calva prioritizes the previous form when there is a competition), I think it will be less surprises also for CIDER users.

mainej:
Thanks for the thorough review @PEZ! Let's go through these one-by-one.

> 1. Sometimes does not drag...

To restate, sometimes a particular element gets "stuck" and won't drag, but then later on it will. That's definitely a bug, and a perplexing one. I'll note that I only see the odd behavior when there's a quote node (`'foo`) present. I haven't done enough research to know if that's actually related, but if it is, it means there's much less likelihood that users will see the problem often—quote nodes are much rarer than symbols, keywords, or other forms. Anyway, needs more research.

> 2. Does not retain cursor position. Moves the cursor to the start of the dragged text.
>    i. Sometimes misplaces the cursor, I can't immediately see a pattern to it.

There are two parts to this.

First, clojure-lsp moves the cursor to the start of the dragged text. In some situations that means the cursor will be at the start of a comment node. The code is working as designed, although that design doesn't align with your expectations. You expect the cursor to stay in the "same spot" within the dragged clause (the quotes are my way of saying that "same" is a dwim thing, and that to stay the same, the row/column would actually change.) Does that sound right?

It's a valid request, but one I'll have to think about. It'll involve a bunch of fiddly math. The current implementation was chosen for two reasons: it's comparatively easy to calculate where the beginning of the dragged clause is, and it's a reliable spot to put the cursor so that repeated invocations work as expected. (That second one may seem obvious, but took a surprising amount of work initially.)

Second, it appears that sometimes the cursor is placed in the middle of a moved clause. That's a bug I can't reproduce. We were seeing the same problem earlier, but it should have been fixed with the recent patch to clojure-lsp. Will you double-check your clojure-lsp version and include the output of the clojure-lsp logs (which print where the server has instructed the cursor to move to)? ~Can VS Code log the requests and responses it's seeing as an LSP client?~ (edit: To answer my own question, yes, you can log LSP level communication via `clojure.trace.server`, which Eric already pointed me to earlier. It'd be informative to have these logs too.)

> 3. Drags the sexpr in front when adjacent behind of a form. And generally does not adhere to Calva's Current Form concept at all.

The video is helpful, but isn't technical enough to understand exactly what you expect. Consider the three cursor positions in the following form:
```clojure
{:a 1|
|
| :b 2}
```

From the video, I think Calva considers the first position part of the `:a 1` clause. What about the second and third cursor positions?

I made an (arbitrary) decision early on that all 3 of those spots belong to the `:b 2` clause. If you try to drag from any of them, you'll drag the "b" clause. I could easily alter that decision and allocate them all to the first `:a 1` clause (unless @ericdallo objects to that behavior change). Any other variation—allocating some of them to the first and some to the second clause—would require a significant re-architecture of the clojure-lsp code, which I'm not sure I'm willing to take on at the moment.

To complicate the discussion a bit, how do comments affect the Current Form concept? Here's a schematic of how clojure-lsp breaks up a form:

```
  [;; comment before a

   ;; another comment before a
   :a ;; comment after a

   ;; comment before b
   :b ;; comment after b
   ]

  ;; is broken up into

  [<><;; comment before a

   ;; another comment before a
   :a ;; comment after a><

   ><;; comment before b
   :b ;; comment after b><
   >]
```

In clojure-lsp terms, there are two clauses in this vector ("a" and "b") interposed between three chunks of padding.

If the user invokes drag from the chunk of padding between `;; comment after a` and `;; comment before b`, the "b" clause is dragged. As described above, that's arbitrary, and it could just as easily be the "a" clause that is dragged. But allocating some padding to the prior clauses and some to the later is much harder.

A final note... clojure-lsp does not currently have any special treatment for `;; =>` comments. If they're on the same line as the last element in the clause they're moved with the clause, but if they're on the following line they're allocated to the next clause. Since it's quite common to put these kind of comments on the following line, this behavior is probably surprising, and so it'd be reasonable to expect them to be allocated to the preceding clause. But that'd be an enhancement, which I may take on, but not as part of this PR.

> 4. Does not drag words inside strings

It's nifty that Calva interprets strings, but as far as clojure-lsp is concerned, in strings "we're not in Kansas anymore, Toto". clojure-lsp does not, and is not likely to ever tokenize or parse strings. The reason for this is that neither of its major data structures—the analysis from clj-kondo and the zipper from rewrite-clj—make any attempt to understand strings, and I doubt we'd ever get them to do so.

If this is to be fixed in Calva, I imagine it'd need a much deeper integration, where sometimes you use the clojure-lsp tools and sometimes you don't.

For now, I think this should be noted in Calva's docs as one of the differences between Calva's drag and clojure-lsp's drag.

> 5. The window scrolls around in a surprising way.

True, but... definitely not a clojure-lsp problem, and not something I feel at all qualified to address.

> 6. Significant clojure-lsp semantic tokens lag/flicker. Especially in large files.
> 7. Performance drop in large files. Nothing surprising. But it is more prominent than with Calva's versions of the commands. I'm using an Apple M1 Max, haven't tested om less powerful hardware.

I'm going to discuss these together. These are a bit hard to see from the video, so I'm more going off what I see on my machine. 

First we need to confirm you have the patched version of clojure-lsp (see response above to 2.) Otherwise, you're still seeing flicker from when clojure-lsp rewrites the entire parent, not just the two clauses.

Second, though I see different types of flicker when using Calva's drag than I see with clojure-lsp's drag, the amount of time that the tokens are mis-colored is about the same for both (in small files—I'll address large files in a second). Conceptually, regardless of which drag implementation you use, the time between the editor making the change and receiving updated semantic tokens should be similar. The only thing I can think of is that perhaps when Calva does its own drag, it also relocates semantic tokens, then waits for clojure-lsp to confirm it did that correctly. If so, that's not something we'll be able to fix since the LSP protocol doesn't define how to edit semantic tokens as you're editing text.

Perhaps it'd be easier for me to see what you're seeing if instead of a video you can provide a bit of code, a cursor position, and a drag direction, so that I can compare the two drags on my machine.

There's one last change I can make that may (or may not) reduce flicker. Currently clojure-lsp asks the editor to make one big edit, replacing two clauses and their intervening padding all at once. It could just as easily ask the editor to make two edits, one for each clause. By excluding their padding there's a chance the editor will be happier, more able to preserve whatever semantic token data it had while it's waiting for updated semantic tokens. I'll try that at some point.

Beyond that, at least from clojure-lsp's side, I'm at a loss for what could be done to improve things. 

Moving on to large files. Dragging a clause in a large file takes an excessively long time to fix semantic tokens. At the moment for me in clojure.core it's taking 25s between the drag and the time that the tokens are fixed. The time seems to be getting longer the more I play with it. What's particularly strange is that the logs show that clojure-lsp is actually returning the semantic token data pretty quickly. After the edit, there's a 300ms debounce time, followed by 1500-2000ms of clj-kondo analysis time, and then 30ms later the tokens are returned. Let's say 2 seconds on average. However, watching the CPU tells a different story. clojure-lsp uses an entire core for tens of seconds after the edit. The editor, even though it's been given the fixed semantic token data, doesn't make any changes until clojure-lsp is done doing whatever it's doing. Unfortunately, clojure-lsp doesn't emit any logs at the end of its CPU hogging, so I don't know what it's been doing.

Anecdotally, the experience in Emacs is similar. Long pauses, getting longer the more I play with it.

This is definitely a performance problem, but one that'll require some analysis and presumably follow-up work.

**To summarize**
* Quote nodes should not get stuck. clojure-lsp bug
* Dragging should leave cursor in the "same spot" (dwim). clojure-lsp enhancement
* Dragging should not put cursor in arbitrary spot in middle of clause. Potentially clojure-lsp bug, but needs steps for reproduction. Otherwise, potentially VS Code bug.
* Dragging from intervening padding should drag prior clause, not subsequent. clojure-lsp enhancement, if all parties agree
* Alternatively, dragging from intervening padding should sometimes drag prior clause, sometimes subsequent, depending on... a yet to be clarified Calva rule about Current Forms. Won't fix, for now, but may consider as an enhancement after more discussion.
* Dragging from a ;; => comment node on a subsequent line should drag prior clause. clojure-lsp enhancement, but needs clarification, because these lines could just as easily belong to subsequent clause.
* Dragging should move words inside strings. Won't fix. Needs Calva doc update.
* Dragging should not scroll window. Won't fix.
* To minimize flicker, dragging should edit clauses separately, not clauses and intevening padding as one large edit. clojure-lsp enhancement, if it improves token coloring in Calva and other editors. Otherwise, may not fix.
* Semantic tokens should update quickly after edits, even in large files. Doesn't seem to be a bug in clojure-lsp's drag code, but is likely a yet-to-be-diagnosed clojure-lsp performance bug. May partially be an editor bug, if it's ignoring valid semantic token data while clojure-lsp is hogging the CPU.

I will:
1. Edit the Calva docs to note that clojure-lsp does not drag within strings.
2. File a clojure-lsp bug about quote nodes getting stuck.
3. File a clojure-lsp bug about CPU hogging and long pauses waiting for semantic tokens to be applied.
4. Collect the enhancements, either into one big Issue, or several smaller ones.

@PEZ will you:
1. Confirm that you have the current release of clojure-lsp.
2. See if you can still reproduce the cursor being put in an arbitrary spot in the middle of the clause. If so, you can file that as a clojure-lsp bug, preferrably with logs.
3. If you'd like, provide a small example of code where flicker is acceptable after Calva's drag, but not after clojure-lsp's.
4. Clarify how multi-line whitespace is allocated to the Current Form.

package.json
mainej: Before I make this change... I want to double-check before we break a convention. The other command names are generated programmatically from the LSP command that is sent over the wire. That is `cycle-privacy`, which is sent to the LSP server, becomes  `cyclePrivacy` via
```javascript
command.replace(/-[a-z]/g, (m) =>
  m.substring(1).toUpperCase()          
```

Are you OK abandoning that convention for drag?

mainej:
As to whether to merge this PR now (after these last few edits) or later, I'll defer to you all. I agree @PEZ, I think the outstanding work can be done incrementally, and I have no objections to noting at as WIP until a few more things are changed. But if you want to wait for it to be more polished, that's fine too. I'll note that I'm going to have pretty limited time to dedicate to open-source over the next few months, so I can't promise I'll get to any of the bugs or enhancements any time soon.

PEZ:
Late here. Can't address much of this right now. I'll pick a few which are hopefully quick:

Firstly. I meant what I said with both that I think it's OK to merge with the current issues I see and that those are mostly observations. I'm not requiring any changes really, maybe with two qualifiers:

1. I wish for a better match with Calva's current form. Right now it is a mindfuck to try figure out what happened in the form when it moves something else than what a Calva user expects.
2. The scrolling around/centering of the window when dragging things ”feels” super weird.

**the Current Form** technical info... I think the place to look is in the code where I have tried to document the behaviour:

https://github.com/BetterThanTomorrow/calva/blob/c78ea3d51c5d52b6d34c7b66952223192533bb2b/src/cursor-doc/token-cursor.ts#L634

There are also tests for this that might clarify.

**Quoted symbols**. In my code var-quotes are pretty common. 😄 And it isn't that they sometimes move and sometimes get stuck. They don't move. Period. I think this is good because probably easier to fix than some Heisenbug behaviour. What you might see as ”sometimes” in the video is that the cursor is placed in a few different spots in the symbol.  It's when the cursor is in the actual ”name” that it doesn't move.

**Lag/Performance**: Hard to see in the video, but I can ”feel” the lag dragging in large files.

**Fliicker**: In a large file it is about the same regardless if I use Calva's drag or clojure-lsp. The difference is that with Calva's drag the enclosing form is losing its semantic styling in its entirety and then it all comes back. With clojure-lsp, it is more that different characters get misscolored within the same symbols. But this flicker has not so much to do with the drag feature, it is more that it is exposing the problem a bit more prominently.

**Comments**: Calva does no attempts to figure about what a comment might belong to. It has never occurred to me to do that. A comment is treated very similar to whitespace. If you wonder about some specific situation for what Calva will regard as the current form, use `ctrl+w` to select the current form placing the cursor at some different locations in some text.

**clojure-lsp message logs**: Yes, you can enable those via a setting. `trace` something... I don't recall right now. But set it to `verbose` and check the clojure-lsp output channel for the logs.

**clojure-lsp version**: I was using `2022.05.03-12.35.40` when I tested.

package.json
PEZ: Hmmm, well, the convention is not all that important to me. I was thinking it would be easier to get the keybindings right if they commands had the same name and I just had to replace the prefix. (I got it wrong several times myself, but maybe I was tired and extra stupid.)

I do see how it makes the code a bit more predictable and easier... So let's skip harmonizing these names with their Calva equivalents. People should read the docs instead of guessing, anyway. 😄

PEZ (approved):
Thanks again!

package.json
mainej: I apologize, but I'm still not clear on what you want. `clojureLsp.dragSexprBackward` or `clojureLsp.dragBackward`?

mainej:
@PEZ you're right about when the quoted symbol moves. Here's a somewhat more minimal reproduction

```clojure
[var |'quote :keyword] ;; <- moves
[var '|quote :keyword] ;; <- doesn't move
```

Thanks. That'll help me debug.

mainej:
Oh, I should note that Calva does a much better job of handling metadata. clojure-lsp makes no attempt to do this intelligently, meaning it'll probably break any code that uses metadata. That would definitely be a nice addition to the clojure-lsp side.

package.json
PEZ: I guess my words leak that I am very ambivalent about it. But let's go for `clojureLsp.dragBackward`

mainej:
Thanks for the `ctrl+w` pointer. Here's a recording of invoking it from various positions:

https://user-images.githubusercontent.com/6774/166832550-2ddd5859-602f-46fa-b4dc-5af141e446bf.mov

I see the internal consistency, but it's very different from clojure-lsp's view, which is also internally consistent. I think what's going to be a sticking point is code like this:

```clojure
{:a 1 ;; comment after a
 |
 ;; comment before b
 :b 2}
```

To me, it's "obvious" that the cursor is on `:b`'s clause. And I think to you @PEZ, and by extension most Calva users, it's "obvious" that the cursor is on `:a`'s clause. Changing clojure-lsp to agree with Calva would be as much of a mind-fuck to me as not changing it would be to Calva users. I'm not sure what to do here. Configuration maybe? It seems like such an un-intuitive thing to configure.

And that's the easy case. Here's a harder case:

```clojure
{:a 1 ;; comment after a
 
 ;; comment before b
 |
 :b 2}
```

To Calva, the cursor is on `:a`'s clause (or more accurately, on `1`, because Current Form doesn't have a concept of clauses.) But to me, the cursor is definitively on `:b`'s clause. I'd oppose any proposal to change that interpretation, which means Current Form and clojure-lsp drag will always be somewhat at odds.

I'm apprehensive that this difference will mean clojure-lsp's drag isn't very useful to Calva users. One of the advertised benefits of clojure-lsp's drag is "comment affinity", but if it moves different comments than Calva users would expect, it's not really an improvement over not moving comments at all. These mental patterns are really hard to alter, especially when different behavior is in effect in different contexts. I remember long ago abandoning paredit for smartparens, because my brain's way of thinking about moving sexps is completely at odds with paredit's. I know lots of people who feel exactly the opposite.

Anyway, I don't have any easy answers here. I'm not inclined to change anything, but it sounds like it's too confusing to be useful to you @PEZ. I suppose if you offer it to Calva users as is, they'll either use it or not. 🤷 

As a mostly unrelated aside: to Calva, comments lines between nodes are sort of a No-Man's Land. `ctrl+w` opens a menu, instead of selecting a form, and Calva's drag is completely disabled. I'm not sure if this is purposeful or not, but it confuses me.

package.json
mainej: Gotcha... in that case, it's OK as is. I'm going to mark this conversation as resolved.

PEZ:
I note that the **Expand selection** keybinding was not optimal for what you wanted to try as we relax that binding in comments. You could use the **Calva: Select Current Form** command, which has a keybinding (`ctrl+alt+c ctrl+s`) that stays active wherever the cursor is in the form. That said, it will do what you have started to suspect it will do. 😄 

I think that for now we can let the edgiest cases be as they are. Doubt many users would be dragging around form starting with the cursor in the middle nowhere. The important one is the one I mentioned specifically:

> Drags the form in front **when adjacent behind** of a form.

This is where CIDER user will place the cursor to evaluate a given form. So when I wanted it to be less of precision activity to evaluate a form in Calva, I opted for prioritizing the previous form, when in doubt, for consistency and to respect CIDER user's expectations.

Anyway, while not my dream solution, to me it's not all that crazy if `:b 2` be dragged in those edgy cases.

I don't at all fancy a solution of making this a configuration. If we were talking about a solution I would instead want to have the drag commands take two arguments:
* _current selection_
* _current form_
And the behaviour be:
* use _current form_  for finding out what to drag
* use _current selection_ to re-position the cursor ”same spot”

Then the editor would make this call, as it makes all other current form calls, and the users will not have to make do with a compromise of their expectations.

mainej:
drag from quoted symbols: https://github.com/clojure-lsp/clojure-lsp/issues/969
slow update of semantic tokens in large files: https://github.com/clojure-lsp/clojure-lsp/issues/970

mainej:
> when adjacent behind of a form

I don't know if you define **adjacent behind** as the single position immediately to the right of the form, or all positions until the next form. But in either case, I agree, I have also been conditioned by CIDER to think of that region as part of the prior form.

But CIDER is not the authority on how to read code. One of clojure-lsp's drag's main purposes is to ensure that the file continues to read correctly even as you drag, meaning both comments and code should continue to "make sense". The way people write comments and code is usually with the comments above the code that they reference. That is, even if a compiler evaluates the _prior_ form, a human reads a comment and applies it to the _subsequent_ form.

From that perspective, to me at least, there's a lot of logic in assuming that comments on their own lines between forms "belong" to the later form.

Still, I agree, a character immediately "behind" a form is a gray zone. But maybe there's a way to resolve the ambiguity:

> Doubt many users would be dragging around form starting with the cursor in the middle nowhere

This is a really important point, and suggests a solution which might be my favorite so far. This is kind of from the worse-is-better school of thought, specifically [the part](https://en.wikipedia.org/wiki/Worse_is_better#New_Jersey_style) that says, "it is better to drop those parts of the design that deal with less common circumstances than to introduce either complexity or inconsistency."

It is less common to drag from the padding between clauses. (I've used the drag tool extensively over the last several months and I almost never try to drag from those locations.) In such padding, it's unclear which clause to drag. Either would make sense. So rather than choosing either, and being inconsistent to half the users, maybe we should do nothing at all. That is, maybe the best thing to do is to disable drag in those locations: as a user, if you want to drag, you have to be clear about your intentions by moving onto the clause you want to drag.

That's a precise, and IMO consistent, rule that could be applied to the padding between clauses. But it leaves open another question... what should drag do when invoked from a comment at the beginning or end of a clause?

I occasionally drag from these comments. But, it's rare, and there's a chance I do it only because I implemented the feature and I know it will work. If I hadn't implemented it, I would probably assume that you had to be "on" one of the forms in the clause you wanted to drag, or maybe between them, but not any of the preceding or ensuing comments.

I'll note that currently the drag tool _needs_ to drag from preceding comments, but that this is an implementation detail. Currently after a drag the cursor is positioned at the beginning of the leading comment. To support repeated drags, the code needs to drag from that comment. But that's all a side-effect of re-positioning the cursor to the head of the clause. If the cursor stayed at its original location within the clause, it wouldn't be necessary to drag from the leading comment. So, I think we should revisit this part of the discussion when I'm making the cursor maintain its position.

What do you think of this @PEZ and @ericdallo? Is it clear what I'm proposing? Does it seem like a reasonable compromise?

mainej:
To preemptively clarify (sorry if this is clear already) consider the following:

```clojure
{:apples 1

 :bananas 2
 }
```

Which gets split into:

```
{<><:apples 1><

 ><:bananas 2><
 >}
 ```
 
 Let's call these: initial (empty) padding, apples clause, interstitial padding, bananas clause, final padding.
 
 At the moment, here's how drag works from the padding:
 1. initial padding: there aren't actually any positions in the initial padding in this code, so it's not possible to invoke drag from here. But if there were, drag would move the apples clause.
 2. intersitial padding: moves the bananas clause
 3. final padding: does nothing

 I'm proposing that drag should do nothing from any of the padding.
 
 @PEZ, that means that the **adjacent behind** position, immediately after `1` would be disabled. I concede that might be occasionally disappointing to users, but also immediately discoverable and easily fixable. If they try to drag, nothing will happen, so they'll probably move back to the `1` character and try again. And the fact that nothing happens will be more logical to them (and honestly to me) than moving the bananas clause.
 
 Interestingly, trailing comments make things better for everyone:

```clojure
{:apples 1 ;; one apple

 :bananas 2}
```

Which gets split into:

```
{<><:apples 1 ;; one apple><

 ><:bananas 2>}
 ```
 
 In this case, the position after `1` _is_ part of the apples clause, and so that's the clause that would be dragged. I think this agrees with both the CIDER perspective and the human reader perspective.

PEZ:
I'm not sure I follow, actually. Especially that it wouldn't be possible to drag from the initial padding there. It looks like the position where the current clojure-lsp drag would move the cursor after a drag. I am probably misunderstanding things.

Anyway, think this PR is the wrong place to keep the discussion, since we have already concluded that we think it's good enough for now. We can use the feedback we get from people trying it to guide us further.

I just want to remind us that drag fwd/back is participating in a collection of drag commands. So we can e.g. drag something down a list, then through the list, and then out of the list in the other end. For someone who wants to use the clojure-lsp variants of drag/fwd backward, it would get weird if when they have entered the list with something, dragging through the list would start dragging something else.

If you are good with how it works now, then I think we should merge, and then do whatever improvements on that we think should be made, when the time allows. It might also happen that we, maybe with your help, can improve Calva's drag fwd/back to a point where the clojure-lsp versions are not significantly more ”smart” and that we then don't want to add the extra complexity that two sets of the ”same” command brings.

mainej:
> If you are good with how it works now, then I think we should merge, and then do whatever improvements on that we think should be made, when the time allows.

I agree, it's time to merge. 🚀 We can do incremental improvements on the clojure-lsp side. In the short term, there aren't any changes to be made on the Calva side. I agree, in the long term, it'd be nice to enhance the Calva implementation to have some of the clojure-lsp features, while aligning with Calva's vision of how positions and forms fit together.

> I'm not sure I follow, actually. Especially that it wouldn't be possible to drag from the initial padding there

I apologize—the "initial padding" is a red-herring for what I was trying to convey. The important part was about the interstitial padding, and how I think we should disable drag in certain ambiguous spots between clauses. Rather than go into the implementation-specific details about what initial padding is, I'll clarify: your intuition is right. From `{|:apples 1 ...` the apples clause would be moved.

PEZ:
I can't thank you enough, @mainej. Looking forward to work with you again!

mainej:
> Looking forward to work with you again!

Same here... I really like the symbiosis, where integrating clojure-lsp closely with one editor turns into bug fixes and enhancements that will help all the editors.

Anyway, thanks! Let me know if you hear about bugs and/or complaints.

Issue #1686, opened 2022-04-18
Indent not working as expected in this specific case


bpringe:
With the below code, if I place my cursor after the closing `]` for the let binding form and hit enter, the cursor ends up where denoted by `|` which is not expected.

Actual:

```clj
(let [x {:some-top-key
         {:some-things
          {2021
           {:my-year 2021,
            :my-things
            [{:some-number 200000, :some-date #inst "2021-11-15T00:00:00.000-00:00"}
             {:some-number 200000, :some-date #inst "2022-02-15T00:00:00.000-00:00"}
             {:some-number 300000, :some-date #inst "2022-05-16T00:00:00.000-00:00"}
             {:some-number 300000, :some-date #inst "2022-07-16T00:00:00.000-00:00"}]},
           2022
           {:my-year 2022,
            :my-things
            [{:some-number 200000, :some-date #inst "2022-11-15T00:00:00.000-00:00"}
             {:some-number 200000, :some-date #inst "2023-02-15T00:00:00.000-00:00"}
             {:some-number 300000, :some-date #inst "2023-05-16T00:00:00.000-00:00"}]}},
          :some-other-things
          [{:some-thing-id "1",
            :some-year 2021,
            :some-person "someone",
            :some-date #inst "2021-11-13T00:00:00.000-00:00",
            :foo :bar}]}}]
            |)
```

Expected:

```clj
(let [x {:some-top-key
         {:some-things
          {2021
           {:my-year 2021,
            :my-things
            [{:some-number 200000, :some-date #inst "2021-11-15T00:00:00.000-00:00"}
             {:some-number 200000, :some-date #inst "2022-02-15T00:00:00.000-00:00"}
             {:some-number 300000, :some-date #inst "2022-05-16T00:00:00.000-00:00"}
             {:some-number 300000, :some-date #inst "2022-07-16T00:00:00.000-00:00"}]},
           2022
           {:my-year 2022,
            :my-things
            [{:some-number 200000, :some-date #inst "2022-11-15T00:00:00.000-00:00"}
             {:some-number 200000, :some-date #inst "2023-02-15T00:00:00.000-00:00"}
             {:some-number 300000, :some-date #inst "2023-05-16T00:00:00.000-00:00"}]}},
          :some-other-things
          [{:some-thing-id "1",
            :some-year 2021,
            :some-person "someone",
            :some-date #inst "2021-11-13T00:00:00.000-00:00",
            :foo :bar}]}}]
  |)
```

[...]

PR #1703, opened 2022-05-01
WIP: YASNCB (Yet Another Strict Null Checks Branch)

corasaurus-hex:
## What has Changed?

This turns on `strictNullChecks` in TypeScript and fixes all the remaining issues in the codebase that were preventing us from turning this setting on.

## My Calva PR Checklist

I have:

- [ ] Tests
- [~] Formatted all JavaScript and TypeScript code that was changed. (use the [prettier extension](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) or run `npm run prettier-format`)
- [~] Confirmed that there are no linter warnings or errors (use the [eslint extension](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint), run `npm run eslint` before creating your PR, or run `npm run eslint-watch` to eslint as you go).

### DIFF
--- a/src/connector.ts
+++ b/src/connector.ts
@@ -39,6 +39,7 @@ async function readJarContent(uri: string) {
 }
 
 async function readRuntimeConfigs() {
+  util.assertIsDefined(nClient, 'Expected there to be an nREPL client!');
   const classpath = await nClient.session.classpath().catch((e) => {
     console.error('readRuntimeConfigs:', e);
   });
@@ -79,7 +80,10 @@ async function connectToHost(hostname: string, port: number, connectSequence: Re
       host: hostname,
       port: +port,
       onError: (e) => {
-        const scheme = state.getProjectRootUri().scheme;
+        const projectRootUri = state.getProjectRootUri();
+        util.assertIsDefined(projectRootUri, 'Expected a project root URI');
+
+        const scheme = projectRootUri.scheme;
         if (scheme === 'vsls') {
           outputWindow.append('; nREPL connection failed; did the host share the nREPL port?');
         }
@@ -112,12 +116,9 @@ async function connectToHost(hostname: string, port: number, connectSequence: Re
 
     if (connectSequence.afterCLJReplJackInCode) {
       outputWindow.append(`\n; Evaluating 'afterCLJReplJackInCode'`);
-      await evaluateInOutputWindow(
-        connectSequence.afterCLJReplJackInCode,
-        'clj',
-        outputWindow.getNs(),
-        {}
-      );
+      const ns = outputWindow.getNs();
+      util.assertIsDefined(ns, 'Expected outputWindow to have a namespace!');
+      await evaluateInOutputWindow(connectSequence.afterCLJReplJackInCode, 'clj', ns, {});
     }
 
     outputWindow.appendPrompt();
@@ -188,7 +189,10 @@ function setUpCljsRepl(session, build) {
 }
 
 async function getFigwheelMainBuilds() {
-  const res = await vscode.workspace.fs.readDirectory(state.getProjectRootUri());
+  const projectRootUri = state.getProjectRootUri();
+  util.assertIsDefined(projectRootUri, 'Expected a project root URI');
+
+  const res = await vscode.workspace.fs.readDirectory(projectRootUri);
   const builds = res
     .filter(([name, type]) => type !== vscode.FileType.Directory && name.match(/\.cljs\.edn/))
     .map(([name, _]) => name.replace(/\.cljs\.edn$/, ''));
@@ -371,20 +375,28 @@ function createCLJSReplType(
         'cljsReplTypeHasBuilds',
         cljsType.buildsRequired
       );
-      let initCode = cljsType.connectCode,
-        build: string = null;
+      let initCode: typeof cljsType.connectCode | undefined = cljsType.connectCode,
+        build: string | null = null;
       if (menuSelections && menuSelections.cljsDefaultBuild && useDefaultBuild) {
         build = menuSelections.cljsDefaultBuild;
         useDefaultBuild = false;
       } else {
         if (typeof initCode === 'object' || initCode.includes('%BUILD%')) {
+          const projectRootUri = state.getProjectRootUri();
+          util.assertIsDefined(projectRootUri, 'Expected a project root URI');
+
+          const buildsForSelection = startedBuilds
+            ? startedBuilds
+            : await figwheelOrShadowBuilds(cljsTypeName);
+          util.assertIsDefined(
+            buildsForSelection,
+            'Expected there to be figwheel or shadowcljs builds!'
+          );
+
           build = await util.quickPickSingle({
-            values: startedBuilds ? startedBuilds : await figwheelOrShadowBuilds(cljsTypeName),
+            values: buildsForSelection,
             placeHolder: 'Select which build to connect to',
-            saveAs: `${state.getProjectRootUri().toString()}/${cljsTypeName.replace(
-              ' ',
-              '-'
-            )}-build`,
+            saveAs: `${projectRootUri.toString()}/${cljsTypeName.replace(' ', '-')}-build`,
             autoSelect: true,
           });
         }
@@ -415,12 +427,10 @@ function createCLJSReplType(
       );
     },
     connected: (result, out, err) => {
-      if (cljsType.isConnectedRegExp) {
-        return (
-          [...out, result].find((x) => {
-            return x.search(cljsType.isConnectedRegExp) >= 0;
-          }) != undefined
-        );
+      const { isConnectedRegExp } = cljsType;
+
+      if (isConnectedRegExp) {
+        return [...out, result].find((x) => x.search(isConnectedRegExp) >= 0) !== undefined;
       } else {
         return true;
       }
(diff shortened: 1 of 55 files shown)


src/connector.ts
```diff
@@ -521,7 +528,8 @@ async function makeCljsSessionClone(session, repl: ReplType, projectTypeName: st
     outputWindow.append(
       ';   The Calva Connection Log might have more connection progress information.'
     );
-    if (repl.start != undefined) {
+    if (repl.start !== undefined) {
+      util.assertIsDefined(repl.started, "Expected repl to have a 'started' check function!");
```
corasaurus-hex: Calls within `repl.start` depend on `repl.started` being defined.

src/cursor-doc/indent.ts
```diff
@@ -61,6 +61,17 @@ export function collectIndents(
...
+  const patterns = _.keys(rules);
+  const regexpPatterns = patterns.reduce((regexpMap, pattern) => {
+    const match = pattern.match(/^#"(.*)"$/);
+
+    if (match) {
+      regexpMap[pattern] = RegExp(match[1]);
+    }
+
+    return regexpMap;
+  }, {} as Record<string, RegExp>);
```
corasaurus-hex, line 73: I needed to fix some nulls here anyways so I took the chance to put into place some of the optimizations I mentioned in that [other issue](https://github.com/BetterThanTomorrow/calva/issues/1686#issuecomment-1102955634).

src/cursor-doc/model.ts
```diff
@@ -155,12 +155,12 @@ export class LineInputModel implements EditableModel {
           }
           return x;
         })
-        .filter((x) => x !== null)
+        .filter((x) => x !== null) as number[]
```
corasaurus-hex, line 158: TypeScript wasn't smart enough to know that filtering out nulls would make a `(number|null)[]` into a `number[]`. We know from the code this is the correct type so we can just assert it.

src/cursor-doc/model.ts
```diff
@@ -170,12 +170,12 @@ export class LineInputModel implements EditableModel {
           }
           return [a, b];
         })
-        .filter((x) => x !== null)
+        .filter((x) => x !== null) as [number, number][]
```
corasaurus-hex, line 173: Same reasoning as above.

src/cursor-doc/model.ts
```diff
@@ -185,7 +185,7 @@ export class LineInputModel implements EditableModel {
...
```
corasaurus-hex, line 188: Same same.

src/cursor-doc/model.ts
```diff
@@ -224,7 +224,7 @@ export class LineInputModel implements EditableModel {
     const seen = new Set<number>();
     this.dirtyLines.sort();
     while (this.dirtyLines.length) {
-      let nextIdx = this.dirtyLines.shift();
+      let nextIdx = this.dirtyLines.shift() as number;
```
corasaurus-hex, line 227: Again, TypeScript isn't smart enough to figure out that the length check on the previous line means this will never be undefined and will always be a number.

src/cursor-doc/token-cursor.ts
```diff
@@ -117,34 +117,29 @@ export class TokenCursor {
...
-  } else {
-    if (!cursor.backwardList()) {
-      return undefined;
-    }
-  }
-  const ranges = [];
+  cursorStartField: StartFieldKey,
+  cursorEndField: EndFieldKey
+): [LispTokenCursor[StartFieldKey], LispTokenCursor[EndFieldKey]][] => {
+  const ranges: [LispTokenCursor[StartFieldKey], LispTokenCursor[EndFieldKey]][] = [];
```
corasaurus-hex, line 128: The explanation for this function is that we needed a generic version of `_rangesForSexpsInList` that didn't need really long types and weird type contortions in order to use it.

src/cursor-doc/undo.ts
```diff
@@ -114,17 +114,21 @@ export class UndoManager<T> {
   undo(c: T) {
     if (this.undos.length) {
       const step = this.undos.pop();
-      step.undo(c);
-      this.redos.push(step);
+      if (step) {
+        step.undo(c);
+        this.redos.push(step);
+      }
```
corasaurus-hex, line 120: More TypeScript silliness. The length check means pop will always work but TypeScript can't figure that out.

src/custom-snippets.ts
```diff
@@ -94,21 +94,23 @@ async function evaluateCodeOrKey(codeOrKey?: string) {
...
+  const code = pick !== undefined ? replSnippet?.snippet : codeOrKey;
+  const ns = pick !== undefined ? replSnippet?.ns : editorNS;
+  const repl = pick !== undefined ? replSnippet?.repl : editorRepl;
 
   const options = {};
 
   if (pick !== undefined) {
-    options['evaluationSendCodeToOutputWindow'] =
-      snippetsDict[pick].evaluationSendCodeToOutputWindow;
+    options['evaluationSendCodeToOutputWindow'] = replSnippet?.evaluationSendCodeToOutputWindow;
```
corasaurus-hex, line 109: strange note: `evaluationSendCodeToOutputWindow` was missing from `customREPLCommandSnippet` and I had to add it.

src/debugger/calva-debug.ts
```diff
@@ -271,6 +271,7 @@ class CalvaDebugSession extends LoggingDebugSession {
     // Pass scheme in path argument to Source contructor so that if it's a jar file it's handled correctly
     const source = new Source(basename(debugResponse.file), debugResponse.file);
     const name = tokenCursor.getFunctionName();
+    util.assertIsDefined(name, 'Expected to find a function name!');
```
corasaurus-hex: StackFrame doesn't allow for an undefined name.

src/doc-mirror/index.ts
```diff
@@ -169,18 +169,21 @@ export class MirroredDocument implements EditableDocument {
   }
 
   public delete(): Thenable<boolean> {
-    return vscode.commands.executeCommand('deleteRight');
+    return vscode.commands.executeCommand('deleteRight').then((v) => !!v);
```
corasaurus-hex, line 173: The actual return type here was `Thenable<true | undefined>` and so this fixes the type so it's `Thenable<boolean>` again.

src/nrepl/index.ts
```diff
@@ -381,6 +382,7 @@ export class NReplSession {
           if (evaluation.onMessage(msg, pprintOptions)) {
             return true;
           }
+          return false;
```
corasaurus-hex, line 385: I'm unsure of this behavior but we must return a boolean from this function and it clearly shouldn't return true.

src/type-checks.ts
```diff
@@ -0,0 +1,21 @@
...
+function assertIsDefined<T>(
+  value: T | undefined | null,
+  message: string | (() => string)
+): asserts value is T {
+  if (isNullOrUndefined(value)) {
+    throw new Error(typeof message === 'string' ? message : message());
+  }
+}
+
+export { isNonEmptyString, isNullOrUndefined, isDefined, assertIsDefined };
```
corasaurus-hex, line 22: We needed to split this out into its own file so it could be imported in files where vscode isn't available. Importing utilities was breaking unit tests because of this.

PEZ:
I found this in the noise where the integration test broke:

```
UNRESPONSIVE extension host: 'betterthantomorrow.calva' took 100% of 5100.624ms, saved PROFILE here: 'file:///tmp/exthost-19cc55.cpuprofile' [ { id: 'betterthantomorrow.calva', total: 5100381, percentage: 100 } ]
```

https://app.circleci.com/pipelines/github/BetterThanTomorrow/calva/5020/workflows/5edc3168-da3a-4cf8-ab29-02c887c079c0/jobs/21872?invite=true#step-104-114

Before that I also find:

```
[137:0501/213503.162639:ERROR:bus.cc(397)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
```

Which I am not sure if we usually see or not, but if not maybe it could be the cause of the timeout.

Anyway, those `UNRESPONSIVE extension host: 'betterthantomorrow.calva' took 100% of 5100.624ms` messages I sometimes see in the dev console when I have introduced some change causing Calva to go into a spin while activating.

I would see if I could run these tests locally. If you haven't before, it's surprisingly easy:

0. Quit VS Code Insiders if you have it running
1. In an integrated terminal of regular VS Code: `npm run integration-test`

corasaurus-hex:
I ran it locally and profiled the code to find where it's stuck. I just need to figure out _why_ it's stuck. I have the last known working commit sha for integration tests but it's not clear why it's broken by the following commit

PEZ:
Which commits does the bisect give you? Maybe we can see what could be the cause with more eyes on it.

PEZ:
Is it here?

https://github.com/BetterThanTomorrow/calva/pull/1703/commits/fcf14e04179ab763d5b3e601b64b5a679f08b2fd

If so, that is still a bit opaque to me. 😄 I don't know what the implications are for enabling strict null checks.

src/connector.ts
```diff
@@ -371,15 +370,23 @@ function createCLJSReplType(
...
-      let initCode = cljsType.connectCode,
-        build: string = null;
+      let initCode: typeof cljsType.connectCode | undefined = cljsType.connectCode,
+        build: string | null = null;
       if (menuSelections && menuSelections.cljsDefaultBuild && useDefaultBuild) {
         build = menuSelections.cljsDefaultBuild;
         useDefaultBuild = false;
       } else {
         if (typeof initCode === 'object' || initCode.includes('%BUILD%')) {
+          const buildsForSelection = startedBuilds
```
PEZ, line 380: I stopped and wondered if we somehow consider the current selection here. Maybe name it `builds`? Or `buildCandidates`?.

src/cursor-doc/indent.ts
PEZ: If there is some behaviour you think could get affected, add some unit tests for that.

src/cursor-doc/model.ts
PEZ: Cool. Thanks for mention I learn from them. ❤️

src/cursor-doc/paredit.ts
```diff
@@ -671,7 +672,9 @@ export function backwardSlurpSexp(
   cursor.backwardList();
   const tk = cursor.getPrevToken();
   if (tk.type == 'open') {
-    const offset = cursor.clone().previous().offsetStart;
+    const previous = cursor.clone().previous();
+    assertIsDefined(previous, 'Expected a token to be before the cursor!');
+    const offset = previous.offsetStart;
```
PEZ, line 677: I'm not sure I am reading this correctly. But generally in Paredit and token cursor code we need to gracefully  handle when the structure does not match some expectation. We most often leave things unchanged in those cases. Does this change mean that we croak? I realize that that would be happening before the change as well, I'm more wondering about what we _should_ be doing here...

src/cursor-doc/token-cursor.ts
PEZ: It's probably an improvement. But with my limited command over the typing system, right now this turned from ”this is messy” to ”I don't understand what I'm reading at all”. Not sure what to do about that...

PEZ: It might also be that I am having a hard time trying to puzzle together the code in my head from the diffs.

PR #1745, opened 2022-05-27
Align behavior of backspace in paredit with other editors

### LINKED ISSUE #1741: Align behavior of backspace in paredit with other editors (Cursive, Emacs)
When your code+cursor is like this:

```
(defn nil-if-blank [s]
  (if (str/blank? s) 
    |
    nil s))
```

and you press backspace, your code and cursor should look like this afterwards:

```
(defn nil-if-blank [s]
  (if (str/blank? s) |
    nil s))
```

Currently, you get this:

```
(defn nil-if-blank [s]
  (if (str/blank? s) 
   |
    nil s))
```

That isn't helpful in paredit mode, because one would never want to start writing anything at an incorrect indentation level. Changing the behavior of Calva to do this would make it match Cursive and Emacs, and wouldn't harm anyone's ability to write correct code.

As for what the Calva code would need to do, I think it is this:

1. If: on your cursors line, everything to the left of the cursor is whitespace
2. And your cursor is not inside a string literal
3. Then backspace does this: 
    - Deletes all whitespace to the left of the cursor
    - Deletes a newline, going to the end of the previous line
    - Trims all trailing whitespace
    - Inserts one space if that line is not blank

[Slack discussion](https://clojurians.slack.com/archives/CBE668G4R/p1653581016040119)

isaksky:
## What has Changed?

This PR brings Calva paredit backspace more in alignement with other editors like Cursive and Emacs.

Fixes #1741

## My Calva PR Checklist

I have:

- [x] Tests

### DIFF
--- a/src/cursor-doc/paredit.ts
+++ b/src/cursor-doc/paredit.ts
@@ -792,19 +792,9 @@ export async function close(
   }
 }
 
-function onlyWhitespaceOnSameLineAsCursor(cursor: LispTokenCursor) {
-  cursor = cursor.clone();
-  while (!cursor.atStart()) {
-    switch (cursor.getPrevToken().type) {
-      case 'eol':
-        return true;
-      case 'ws':
-        cursor.previous();
-        continue;
-      default:
-        return false;
-    }
-  }
+function onlyWhitespaceLeftOfCursor(cursor: LispTokenCursor) {
+  const prevToken = cursor.getPrevToken();
+  return prevToken.type === 'ws' && prevToken.offset === 0;
 }
 
 function backspaceOnWhitespaceEdit(doc: EditableDocument, cursor: LispTokenCursor) {
@@ -838,8 +828,7 @@ function backspaceOnWhitespaceEdit(doc: EditableDocument, cursor: LispTokenCurso
   let selectionStart = trimStart;
   const modelEdits = [new ModelEdit('deleteRange', [trimStart, start - trimStart])];
   const cursorTokenType = cursor.getToken().type;
-  const destLineIsAllWhitespace =
-    cursorTokenType === 'ws' && onlyWhitespaceOnSameLineAsCursor(cursor);
+  const destLineIsAllWhitespace = cursorTokenType === 'ws' && onlyWhitespaceLeftOfCursor(cursor);
   if (!destLineIsAllWhitespace && cursorTokenType !== 'open') {
     modelEdits.push(new ModelEdit('insertString', [trimStart, ' ']));
     selectionStart += 1;
@@ -880,7 +869,7 @@ export async function backspace(
           selection: new ModelEditSelection(p - prevToken.raw.length),
         }
       );
-    } else if (!cursor.withinString() && onlyWhitespaceOnSameLineAsCursor(cursor)) {
+    } else if (!cursor.withinString() && onlyWhitespaceLeftOfCursor(cursor)) {
       return backspaceOnWhitespaceEdit(doc, cursor);
     } else {
       if (['open', 'close'].includes(prevToken.type) && docIsBalanced(doc)) {
--- a/src/cursor-doc/paredit.ts
+++ b/src/cursor-doc/paredit.ts
@@ -1,4 +1,5 @@
 import { validPair } from './clojure-lexer';
+import { getIndent } from './indent';
 import { ModelEdit, EditableDocument, ModelEditSelection } from './model';
 import { LispTokenCursor } from './token-cursor';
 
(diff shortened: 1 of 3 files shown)


src/cursor-doc/paredit.ts
```diff
@@ -792,6 +792,64 @@ export async function close(
...
+      case 'eol':
+        return true;
+      case 'ws':
+        cursor.previous();
+        continue;
+      default:
+        return false;
+    }
+  }
+}
```
PEZ: This checks if there is only whitespace before the cursor, right? Thinking that the name says something else.

Also, I think we collect all consecutive whitespace on a line as a single token. Maybe that means that there is only whitespace before the cursor if the `getPrevToken().type === 'ws' && getPrevToken().offset === 0`? (Not sure about this.)

src/cursor-doc/paredit.ts
```diff
@@ -792,6 +792,64 @@ export async function close(
...
+  const cursorTokenType = cursor.getToken().type;
+  const destLineIsAllWhitespace =
+    cursorTokenType === 'ws' && onlyWhitespaceOnSameLineAsCursor(cursor);
+  if (!destLineIsAllWhitespace && cursorTokenType !== 'open') {
+    modelEdits.push(new ModelEdit('insertString', [trimStart, ' ']));
+    selectionStart += 1;
+  }
+  return doc.model.edit(modelEdits, {
+    selection: new ModelEditSelection(selectionStart),
+    skipFormat: false,
```
PEZ: Shouldn't we skip format to match your desired behaviour with leaving a space before the cursor in case we don't have any text after?

src/cursor-doc/paredit.ts
```diff
@@ -792,6 +792,64 @@ export async function close(
...
+      case 'ws':
+        cursor.previous();
+        continue;
+      default:
+        return false;
+    }
+  }
+}
+
+function backspaceOnWhitespaceEdit(doc: EditableDocument, cursor: LispTokenCursor) {
```
PEZ, line 806: My intuition here would be to check if all the text before the cursor on the same line is whitespace, and if so send a cursor `backwardWhitespace()` (without skipping comments), then replace the range we have moved with '' if the previous token is an open and ' ' otherwise. Maybe you considered that and there is a reason not to keep it that simple?

src/extension-test/unit/cursor-doc/paredit-test.ts
```diff
@@ -1268,6 +1268,23 @@ describe('paredit', () => {
...
+        );
+        const b = docFromTextNotation(
+          `
+(defn nil-if-blank [s]
+  (if (str/blank? s) |nil s))
+        `.trim()
+        );
+        await paredit.backspace(a);
+        expect(textAndSelection(a)).toEqual(textAndSelection(b));
+      });
```
PEZ: I think we should cover more of the cases with tests. So that it is clear how it should behave in the different situations we've discussed.

We usually replace newlines with that dot-character. Maybe writing the text like this with literal newlines is better, though, so let's keep it.

However, we can use much less text, I think. It doesn't need to be real code.

PEZ (commented):
Thanks for doing this! 🙏 I have yet to test it, but have left some comments on the changes.

src/cursor-doc/paredit.ts
isaksky: I didn't have any problems with the space. Formatting may be ideal to avoid, but then we'd need a way to insert the correct number of spaces on your destination line. For example if you press backspace in this form:

https://user-images.githubusercontent.com/607986/170801894-965fd2bb-3c29-48ed-a8da-f4b7717d9342.mp4

You see how it shifts all the way to the left, then gets indented properly? If I change to skipFormat: true, it does not get indented properly. Do you know of a way to determine the correct number of spaces to insert?

src/cursor-doc/paredit.ts
isaksky: Ok, I didn't know about `backwardWhitespace`, but one problem would be that it skips unlimited line endings. I can use it, but then I'd need to add a parameter to limit the number of 'eol' tokens to skip. Is that ok?

Here is an example from Cursive showing how we shouldn't kill all blank lines before the cursor:

https://user-images.githubusercontent.com/607986/170802133-e05c2403-efcf-4c66-bf71-bf4367fad8f2.mp4

src/cursor-doc/paredit.ts
isaksky: You are right, fixed here: https://github.com/isaksky/calva/commit/4457629ee446cba9e0da9cef5e09d48c9e7b81f1

src/cursor-doc/paredit.ts
isaksky: Here is how it works with the current code, despite doing the reformatting:

https://user-images.githubusercontent.com/607986/170802715-ca0a07cf-935a-4c4e-bf8f-562f39d38d4b.mp4

To me that looks correct, at least for those cases.

src/extension-test/unit/cursor-doc/paredit-test.ts
isaksky: Ok, I'll work on adding more tests and using less text next time I get a chance to work on this.

src/cursor-doc/paredit.ts
PEZ: Ah, yes, we do want the indents in the first case. Right now we only have a mechanism for post-formatting, maybe we should consider doing post-indentation instead/as an alternative... (for later, of course). We do have a way to calculate the indents, in case you want to explore using it for this edit:

https://github.com/BetterThanTomorrow/calva/blob/dev/src/cursor-doc/indent.ts#L147

(I'm fine with formatting, but putting in the correct indents would be less flickery.)

src/cursor-doc/paredit.ts
PEZ: We try to preserve empty lines in Calva while typing as well, so that Cursive behaviour looks right to me.

Maybe what should be replaced is ”the text before the cursor on the same line + the newline before that”? The range would then be `prevPrevToken.startOffset` to `prevToken.endOffset`. (Haven't had my morning coffee yet, so might be missing something important here.) And as long as we're doing post-formatting we might get away with always replacing with a space.

PEZ: Thinking a bit more about this... if from prevPrevToken,startOffset, there is only whiteSpaceLeftOfCursor, and we do `getIndents()` from there, then we can replace the range with that many spaces and get rid of the post-formatting.

isaksky:
@PEZ It works with `getIndent` now, and I added a bunch of test cases. I'm not sure it is as concise as you were hoping for, but it seems to work well in my testing. Can you check it out?

PEZ:
Hi @isaksky! Sorry for the radio silence. Looking at it now.

PEZ (approved):
This is awesome. Thank you! 🙏 ❤️

armed:
Hey, folks. I think I found a bug in a new version with this commit. 

If you press backspace in this snippet
```
(def foo
  #|{:bar})
```
it turns to
```
(def foo
  # |{:bar})
```

Or it's just me?

armed:
Same with `'`, `@`, `#'` etc. It continues to add new whitespaces when you press backspace multiple times.

isaksky:
😬 You're right, I get the same thing. I'll look into it.

isaksky:
Fixed here: https://github.com/BetterThanTomorrow/calva/pull/1768

PR #1756, opened 2022-05-31
Symbols provider

### LINKED ISSUE #1755: Provide document symbols via Calva's provider
It seems that #1752 was not enough to make Joyride provided symbols to appear together with the ones Calva provide. I think this might be because it is the clojure-lsp server providing it. Let's try use the lsp-client from a Calva document symbols provider and see if that makes it start to work.

PEZ:
## What has Changed?

Implementing a document symbols provider by using clojure-lsp for it, and disabling the ”automatic” provider in lsp client middleware.

It doesn't work yet. I don't understand why. I'm merely passing on the promise and the function in the clojure-lsp main.ts file clearly works.

Fixes #1755

## My Calva PR Checklist

I have:

- [ ] Tests

### DIFF
--- a/src/providers/symbols.ts
+++ b/src/providers/symbols.ts
@@ -1,0 +1,15 @@
+import * as vscode from 'vscode';
+import lsp from '../lsp/main';
+import * as util from '../utilities';
+
+export class CalvaDocumentSymbolsProvider implements vscode.DocumentSymbolProvider {
+  provideDocumentSymbols(
+    document: vscode.TextDocument,
+    _token: vscode.CancellationToken
+  ): vscode.ProviderResult<vscode.SymbolInformation[] | vscode.DocumentSymbol[]> {
+    return lsp.getDocumentSymbols(
+      util.cljsLib.getStateValue(lsp.LSP_CLIENT_KEY),
+      decodeURIComponent(document.uri.toString())
+    );
+  }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Changes to Calva.
 
 ## [Unreleased]
 
+- [Provide document symbols via Calva's provider](https://github.com/BetterThanTomorrow/calva/issues/1755)
+
 ## [2.0.279] - 2022-05-30
 
 - [Expose Calva's `registerSymbolProvider` function in the extension API](https://github.com/BetterThanTomorrow/calva/issues/1752)
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -39,6 +39,7 @@ import * as joyride from './joyride';
 import * as api from './api/index';
 
 import * as clojureDocs from './clojuredocs';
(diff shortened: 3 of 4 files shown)


src/providers/symbols.ts
```diff
@@ -0,0 +1,12 @@
+import * as vscode from 'vscode';
+import lsp from '../lsp/main';
+import * as util from '../utilities';
+
+export class CalvaDocumentSymbolsProvider implements vscode.DocumentSymbolProvider {
+  provideDocumentSymbols(
+    document: vscode.TextDocument,
+    _token: vscode.CancellationToken
+  ): vscode.ProviderResult<vscode.SymbolInformation[] | vscode.DocumentSymbol[]> {
+    return lsp.getDocumentSymbols(util.cljsLib.getStateValue(lsp.LSP_CLIENT_KEY), document.uri);
```
bpringe, line 10: This returns a promise, and from what I see in the debugger decorations code, once the promise is resolved, you may need to take the `children` of the first item of the resulting array.

https://github.com/BetterThanTomorrow/calva/blob/25f18d2821a7846a9045d0ab0ab4581e610a73d1/src/debugger/decorations.ts#L50

bpringe: So maybe something like this would work:

Make `provideDocumentSymbols` async, then do this:

```typescript
const symbolsResult = await lsp.getDocumentSymbols(util.cljsLib.getStateValue(lsp.LSP_CLIENT_KEY), document.uri);
return symbolsResult[0].children;
```

bpringe: I don't remember the structure or the details of the response from clojure-lsp at the moment, but if it's the case that we always want the `children` of the first item in the array, maybe we should move that into the `getDocumentSymbols` function.

PEZ: I have actually tried what you suggest there, and confirmed that the symbols to get assigned to `symbolsResult`, and also tried to return only the children. However, I don't think that this with `children` is the culprit. clojure-lsp creates the symbols as children to the namespace symbol, and in the case of the debugger we are only interested in these children. The problem I have with this PR is that _nothing_ gets provided.

bpringe: Does the disabling of the middleware cause this, or did you verify that it's not called when you manually make the request?

PEZ: Not sure I follow. But this provider is being called, and the symbols are being gathered. They are not being provided, though. (The Outline remains un-populated.)

PEZ: I think I'll close this PR. It is a very limited use case and having a separate group of symbols can be good enough.

bpringe: I thought you meant `lsp.getDocumentSymbols` was not returning symbols. And I was wondering if the middleware being changed to return `null` was the cause, but it seems you're saying that you were getting symbols from lsp, just that after returning

PR #1786, opened 2022-06-26
Add `definitionProviderPriority` configuraion

### LINKED ISSUE #1785: Add configuration for which symbol definition provider to use
Currently Calva prioritizes clojure-lsp's symbol definitions over those provided by nREPL. It's hard to navigate the the pros and cons with both provider options, it is better to let the user's make this choice.

PEZ:
## What has changed?

Adding a user setting for the choice between lsp and nrepl as definition provider. Defaulting to nrepl.

Fixes #1785

## My Calva PR Checklist

I have:

- [ ] Tests

### DIFF
--- a/src/providers/definition.ts
+++ b/src/providers/definition.ts
@@ -12,11 +12,6 @@ import { DefinitionRequest } from 'vscode-languageclient';
 
 const converter = createConverter(undefined, undefined);
 
-const definitionProviderOptions = {
-  replThenLsp: { priority: ['repl', 'lsp'] },
-  lspThenRepl: { priority: ['lsp', 'repl'] },
-};
-
 const definitionFunctions = { lsp: lspDefinition, repl: provideClojureDefinition };
 
 async function provideClojureDefinition(document, position: vscode.Position, _token) {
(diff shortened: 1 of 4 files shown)


src/providers/definition.ts
```diff
@@ -5,13 +5,17 @@ import annotations from './annotations';
...
 import { createConverter } from 'vscode-languageclient/lib/common/protocolConverter';
 import { getClient } from '../lsp/main';
 import { DefinitionRequest } from 'vscode-languageclient';
 
 const converter = createConverter(undefined, undefined);
 
-const definitionProviderOptions = { priority: ['lsp', 'repl'] };
+const definitionProviderOptions = {
+  replThenLsp: { priority: ['repl', 'lsp'] },
+  lspThenRepl: { priority: ['lsp', 'repl'] },
```
bpringe: @PEZ Do you think we could make this code + setting a bit more flexible by just letting them specify the priority array in the setting, rather than specifying a string that maps to a priority array? We could just explain that the providers will be tried in the order they specify in the setting.

bpringe: I suppose it's unlikely we'll add other providers, but if we did, it would be easier to just add a string to the enum of the setting, then add the definitionFunctions, and there wouldn't be a need to modify this mapping.

PEZ: Afaik there is no way to let the users specify order other than letting them specify the array. Then we would need to validate the entry and figure about error messages and such.

If there is a way to do this while ensuring the user can only specify our two options, I'd be most happy to change to that, for sure.

PEZ: Figured out how to specify the arrays in the enum. It renders a bit weird in the Settings view, but that's for the VS Code team to figure about, I think:

image


The user can still provide something bad, but the editor does a good job in highlighting this. (Also we had the same problem with the strings, anyway.)

image

Cyrik: I agree with @bringe, having the array as configuration is more flexible. I'm hoping to add custom definition providers if they are useful, so it's good to keep the system as open as possible

PEZ: Yeah, I think we all agree. The only question was if we could provide a helpful settings UI. I'm all for open, but not if the Ux makes it hard to provide correct settings.

PEZ: As for custom definition providers... Maybe Calva isn't the place to add them? Please file a separate issue about that and we can discuss.

bpringe: I like the change! I was thinking more of an array of values from an enum, instead of an enum of arrays, but I don't know if that's possible with the settings. This is good too, though.

PEZ: Thanks for reviewing!

Issue #1850, opened 2022-08-31
LiveShare participants incorrectly opening every Clojure file as if via "Open with Notebook"


pithyless:
## TL;DR

In `v2.0.292` Clojure Notebooks were introduced.

When using Calva with LiveShare the participants cannot open any Clojure files as regular files; they always open as if they chose "Open with Notebook".

Downgrading to `v2.0.291` removes problem (running with VSCode `1.70.2`).

## Steps to reproduce:

* Host opens Clojure project
* Host starts LiveShare and shares link with participant
* Participant opens link, connects to Shared system
* Participant opens any Clojure file, either via:
  - cmd-p > find file > enter
  - cmd-p > find file > click
  - cmd-p > find file > click the icon for "open to the side"
* Participant sees the file opened as a Clojure Notebook, not a regular Clojure file

Furthermore, after some exploration we found that if the participant opens the context menu for a `.edn` file in the sidebar, we see:

![vscode-edn](https://user-images.githubusercontent.com/21281/187597124-80c0ee87-ed31-4a89-8985-1990e5afe933.jpg)

In comparison, a `.clj` file has the following context-menu:

![vscode-clj](https://user-images.githubusercontent.com/21281/187597115-2ce41410-07e9-4e3c-ba06-c6c45ca04b61.jpg)

Notice the distinct lack of "Open With..." for the `.clj` file.  The host sees "Open With..." both for the `.edn` and `.clj` file.

In contrast, if the host opens a Clojure file and uses the "Focus Participants" icon, the file will open correctly as a Clojure file for the participant.

## Reducing false-positives

We have tried the following so far to reduce false-positives:
* Host disabled any other VSCode extension that may have added Notebook features, etc.
* Participant has VSCode with minimal extensions enabled
* Participant tried connecting via the VSCode LiveShare in a browser
* Sharing a project without a running REPL (so even without host connecting to any REPL, we still see participants opening files always as Notebooks)

## Candidates for additional tests (if necessary)

We have not yet tried:
* REPL via jack-in (we have a multi-project monorepo and use connect-to existing REPL). But considering the issue is visible without any connected REPL, I don't think this is a primary candidate.
* Running a sample Clojure project (we've only tried to reproduce on work projects for now)
* Participant is logged in to LiveShare system (participants have up to now been connecting as Guests; the host is logged in)
[...]

PR #1873, opened 2022-09-20
Disable notebook support for normal clojure file extensions

svdo:
This is a work-around because currently Live Share causes clojure files to always be opened as notebooks.

See:

https://github.com/BetterThanTomorrow/calva/issues/1850

This PR is not intended actually be merged. It is a method (proposed by @PEZ) to use custom Calva builds without notebook support, so that we're not stuck on an older version of Calva. Hopefully the dynamic notebook APIs will be added to VSCode soon, so that we don't have to configure it in `package.json` anymore and this PR can be deleted.


svdo:
This PR is automatically built by CircleCI. You can click "show all checks", click on the link to "ci/circleci: build", go to the artefacts tab, and grab the built `vsix` there, and manually install it.

PEZ:
Not sure about this, but you might need to bump the version as well. Because if people have auto-update of extensions on, I think that a release version will win over a pre-release version in a semantic versioning compare.

Also consider adding info about where to pick up this VSIX to https://calva.io/live-share/

Issue #1889, opened 2022-10-10
Error popups when editing some invalid forms


skylize:
All Calva's dependencies that use `rewrite-clj`, such as `clojure-lsp`, commonly propogate parsing exceptions. (I'm guessing Calva's direct use of `rewrite-clj` does the same?) Calva then fails to handle these exceptions or actively throws them at the user, causing a popup modal, which shows the opaque message of `Request textDocument/codeAction failed.`, and does not auto-dismiss itself.

An example case that occurs frequently during normal (even if perfect) editing is caused by an unbalanced map anywhere between 2 parens.

```clojure
;; The | represents the cursor.
;; Typing anything at all from this point (i.e. to type the first key)
;;  will cause and exception popup.
(foo-fn {|})
```

The user experience is severely degraded by these unavoidable popups that only go away with manual intervention. Calva should catch these exceptions and route them to a Diagnostic. They can be reported to the user as equivalent to a linting error.

One question here is where in the document the error should be reported. (Pretty sure we won't have meaningful line and column numbers from a failed parse?) The obvious choice would be either the start or end of the document. My personal opinion is it would be more helpful to use the start or end of whichever line the cursor happens to be on. This ensures the user receives visible feedback from anywhere in the document.

PEZ:
Thanks for the issue!

I can't reproduce the issue with your example there. Maybe there is something more in your file you are testing with?

skylize:
I am getting intermittent results on that particular reduction of the problem as well. Typing into `({|})` will throw 100%.

<details>
<summary>With only <code>({})</code>, we get a lint error but otherwise no problems:</summary>
<pre>
[Trace - 10:22:54 AM] Sending notification 'textDocument/didChange'.
Params: {
    "textDocument": {
        "uri": "file:///home/sky/edu/clojure/lsp-bug/lsp_bug.clj",
        "version": 10
    },
    "contentChanges": [
        {
            "text": "({})"
        }
    ]
}


[Trace - 10:22:54 AM] Sending request 'textDocument/codeAction - (34)'.
Params: {
    "textDocument": {
        "uri": "file:///home/sky/edu/clojure/lsp-bug/lsp_bug.clj"
    },
    "range": {
        "start": {
            "line": 0,
            "character": 2
        },
        "end": {
            "line": 0,
            "character": 2
        }
    },
    "context": {
        "diagnostics": [],
        "triggerKind": 2
    }
}


[...]

PR #1904, opened 2022-10-19
Filter out code action errors

bpringe:
## What has changed?

- This filters out the codeAction errors that popup in some imbalanced form situations, which can be annoying to users. When you have this form: `({})`, if you add `:` in the middle, the error pops up. With this PR, the error is filtered out and no popup is shown.

Addresses #1889.

## My Calva PR Checklist

- [x] Tests
  - ~[ ] Figured if the change might have some side effects and tested those as well.~
  - ~[ ] Smoke tested the extension as such.~
  - ~[ ] Tested the VSIX built from the PR (so, after you've submitted the PR). You'll find the artifacts by clicking _Show all checks_ in the CI section of the PR page, and then _Details_ on the `ci/circleci: build` test.~
  - ~[ ] If I am fixing the issue, I have used [GitHub's fixes/closes syntax](https://help.github.com/en/articles/closing-issues-using-keywords)~

### DIFF
--- a/src/overrides.ts
+++ b/src/overrides.ts
@@ -1,0 +1,14 @@
+import * as vscode from 'vscode';
+
+const vscodeShowErrorMessage = vscode.window.showErrorMessage;
+
+async function calvaShowErrorMessage(message: string) {
+  const exlusionRegexps = [/Request textDocument\/codeAction failed./];
+  if (!exlusionRegexps.some((re) => re.test(message))) {
+    return vscodeShowErrorMessage(message);
+  }
+}
+
+export function activate() {
+  vscode.window.showErrorMessage = calvaShowErrorMessage;
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Changes to Calva.
 
 ## [Unreleased]
 
+- [Filter out code action errors](https://github.com/BetterThanTomorrow/calva/pull/1904), addressing [this issue](https://github.com/BetterThanTomorrow/calva/issues/1889)
+
 ## [2.0.308] - 2022-10-19
 
 - [A more flexible evaluate-to-cursor command](https://github.com/BetterThanTomorrow/calva/issues/1901)
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -40,6 +40,7 @@ import * as joyride from './joyride';
 import * as api from './api/index';
 import * as depsClj from './nrepl/deps-clj';
 import * as clojureDocs from './clojuredocs';
+import * as overrides from './overrides';
 
 async function onDidSave(testController: vscode.TestController, document: vscode.TextDocument) {
   const { evalOnSave, testOnSave } = config.getConfig();
@@ -97,6 +98,9 @@ function initializeState() {
 
 async function activate(context: vscode.ExtensionContext) {
   console.info('Calva activate START');
+
+  overrides.activate();
+
   initializeState();
   await config.updateCalvaConfigFromUserConfigEdn(false);
   await config.updateCalvaConfigFromEdn();


bpringe:
@PEZ I just said "addressing" the issue since the root cause still remains to be fixed upstream. See the changelog entry too. I can change that if you want.

PEZ (commented):
Thanks! 🙏  So good you found this workaround. Looks great!

src/overrides.ts
```diff
@@ -0,0 +1,14 @@
+import * as vscode from 'vscode';
+
+const vscodeShowErrorMessage = vscode.window.showErrorMessage;
+
+async function calvaShowErrorMessage(message: string) {
+  const exlusionRegexps = [/Request textDocument\/codeAction failed./];
+  if (!exlusionRegexps.some((re) => re.test(message))) {
```
skylize, line 7: We should probably only run exclusions when `document.languageId !== 'clojure'`.

src/overrides.ts
```diff
@@ -0,0 +1,14 @@
...
+    return vscodeShowErrorMessage(message);
+  }
+}
+
+export function activate() {
```
skylize, line 12: This should return a disposable, that can be pushed into `context.subcriptions`, to restore built in `showErrorMessage`.

src/overrides.ts
bpringe: Anything coming through with that message is coming from clojure-lsp / Calva. I don't think this can have any effect on other extensions or other lsp servers, if that's what you're concerned about.

bpringe: With the above in mind, I don't think that check is necessary.

src/overrides.ts
bpringe: Are you suggesting we create a class that implements `Disposable`, and then create a dispose method that restores the built in `showErrorMessage`?

Considering I don't think this affects any other extensions, I don't see the need for that, but if we did want to restore the built-in `showErrorMessage`, I'd rather add a `deactivate` function. So in essence, we're activating and deactivating overrides.

src/overrides.ts
skylize: That is not a safe assumption. CodeActions are part of LSP spec, and not limited to Clojure. While the odds of some other LSP is using the exact same error message is low, it is definitely not zero.

src/overrides.ts
skylize: This not about potentially making this change optional for users. It about being good citizens, when for example, someone uninstalls the extension. 

All of the numerous calls to `context.subcriptions.push` in `extension.ts` are to provide disposables to VS Code. When/if it needs to, Code can call `dispose` method of each disposable in that collection, to break down all of the Calva's infrastructure, and release things for garbage collection or undo global changes (like hijacking the error printer).

It's just one more line in this function...
```typescript
  return {dispose: () => vscode.window.showErrorMessage = vscodeShowErrorMessage}
```
... and then in `extension.ts` wrap in a push to subscriptions.
```
context.subscriptions.push(overrides.activate())
```

src/overrides.ts
skylize: Also the way this is written implies readiness to filter additional errors in the future. Without verifying Clojure, we need to be very careful and precise with future regex patterns.

bpringe: Maybe we should verify this, but I really don't think this would effect anything from any other extension. This override is specific to the runtime of the Calva extension alone.

@PEZ What's an easy way to test this? Maybe we could use a vsix from joyride that calls `showErrorMessage` and verify that they aren't intercepted by this override. I think that would be pretty crazy if they were.

Yes, this is written to make it easier to filter additional messages in the future if needed, however unlikely that may be.

Even if we verified the language is Clojure, if this override were to effect other extensions, it probably wouldn't be a good idea to use in the first place.

src/overrides.ts
bpringe: > This not about potentially making this change optional for users.

I wasn't suggesting optionality for users, just to be clear.

>  It about being good citizens, when for example, someone uninstalls the extension.

If I thought this would affect other extensions, I wouldn't make this change, at least not as it is, and maybe not at all.

> All of the numerous calls to context.subcriptions.push in extension.ts are to provide disposables to VS Code.

I'm aware of this. I've been a maintainer of Calva for quite some time now 😃. I appreciate your reviewing and concerns, though.

Also, thanks for reminding me that we don't need a class and can just use an object literal with a `dispose` property.

src/overrides.ts
bpringe: I verified that this override does not affect calls to `showErroMessage` from other extensions by using a vsix of joyride that I made call that function, and a vsix of this PR that I made log when a message is intercepted. 🎉

skylize: Awesome. Thanks for checking. And bigger thanks for finding and implementing this.

PEZ: I'm curious about the way to verify. Wouldn't it have worked to just use a the command **Joyride: Run Clojure Code** to pop up a matching error message? Sounds like you did something more involved, @bpringe? Super happy you checked it, just curious about if I missed something, or maybe if there is info we could add to the Joyride docs.

bpringe: To be honest, I haven't used joyride yet, and instead of learning the ropes I just did what I knew 😄. But maybe testing code that was compiled into the extension was a better test to be sure there's not some difference in how that command runs code? I don't know.

PEZ: That's good to know. I think we have some work to do with Joyride to relay just how easy it is to start using it. 😄 

If the test was to use `vscode/window.showErrorMessage` then this would have been tested just as well using Joyride. Everything that's done in a Joyride script is done in the context of the Joyride extension. FYI.

bpringe: Good to know! Thanks. Also, having not tried to get into Joyride yet, I don't know how easy it might be to get started, so I'm by no means suggesting it's not easy.

PR #1925, opened 2022-10-29
Use on-type settings for post paredit formatting

### LINKED ISSUE #1922: Calva Paredit: Kill/Delete Right deletes also the space before the form
If I have the following row and | marks the cursor:

```clojure
(count |[:a :b])
```

... and then I give command `Calva Paredit: Kill/Delete Right` ... I get:

```clojure
(count|)
```

and not:

```clojure
(count |)
```

(i.e. I'd like to have the space preserved) ... and therefore if I "paste" the form back I get:

```clojure
(count[:a :b])
```

and not:

```clojure
(count [:a :b])
```

We had some discussion about this in the Clojurians slack:
https://clojurians.slack.com/archives/CBE668G4R/p1666803713170629

I copy-paste here some information regarding that discussion that might help the person working with this ticket:

If I understood correctly the discussion with `pez` and `skylize` the `:remove-surrounding-whitespace?` rule in cljfmt causes this behaviour. 
`skylize` said: "A config to disable that rule would make this go away, but would also break the many cases where applying this rule is desirable."
`pez` said: "Calva already distinguishes between format-on-demand and format-as-you-type, relaxing the formatting for the latter case. It could be that this strategy should be expanded on, or some similar approach should be taken for Paredit-induced formatting (which counts as format-as-you-type)."

### LINKED ISSUE #1924: Use format-as-you-type for Paredit
When formatting, Calva has two basic ”modes”:

1. **On demand**: Strictly follow the users settings
2. **On type**: The users settings, but relax trimming surrounding and trailing whitespace, and, consecutive blank lines.

The formatting that happens with Paredit is, as I see it, format-as-you-type. But Calva doesn't treat it that way, and instead uses on-demand formatting. Why? I don't remember why. Could be an honest mistake. Could be some good reason. Although, I really can't think of the reason.

As things are, we get issues like this one:

* #1922

I think we should try format-as-you-type for Paredit, even for a release, and see what the feedback is. We might find some cases where we miss some whitespace trimming, but I am guessing that we will be able to fix those, case by case, rather than going back to on-demand format.

PEZ:
## What has changed?

What the title says, and what #1922 suggests

* Fixes #1922
* Fixes #1924

I have:

- [ ] Tests


bpringe (approved):
LGTM

PR #1948, opened 2022-11-06
Some VIM co-existence fixes

### LINKED ISSUE #1947: Make Calva a bit more VIM friendly
User of the VIM extension often run into troubles with some incompatibilities with Calva. Specifically these two issues are reported:

* The cursor jumps to the beginning of the line when entering new lines
* The `escape` key is bound to dismiss inline results

Users get rid of the issue with the cursor positioning by configuring Calva to not use the ”New indent engine”. I have been thinking this is a matter of the old engine being slower and thus being later than the VIM extension to the party. But today I realized that another difference is that the old indent engine actively sets the cursor position after re-indenting. So maybe we can do the same with the new indenter and solve this issue...

The `escape` key problem could be made much less impactful if we only bind it when we are actually showing inline results.

PEZ:
## What has changed?

* Introduce a context state for when inline results are displayed or not
* Using the new context state to only bind `escape` when inline results are showing
* Explicitly set the new cursor position after re-indenting with the new indent engine
* Some bug fixes with how we make sure the new indent engine gets a clean slate to start from

In my testing, the explicit cursor positioning fixes the problem with the cursor ending up at the start of te line when the VIM extension is enabled. 

Fixes #1947

## My Calva PR Checklist

I have:

- [ ] Tests

### DIFF
--- a/src/calva-fmt/src/extension.ts
+++ b/src/calva-fmt/src/extension.ts
@@ -79,7 +79,7 @@ export function activate(context: vscode.ExtensionContext) {
     )
   );
   vscode.window.onDidChangeActiveTextEditor(inferer.updateState);
-  vscode.workspace.onDidChangeConfiguration(async (e) => {
+  vscode.workspace.onDidChangeConfiguration((e) => {
     if (e.affectsConfiguration('calva.fmt.formatAsYouType')) {
       vscode.languages.setLanguageConfiguration('clojure', getLanguageConfiguration());
     }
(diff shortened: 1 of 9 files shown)


src/calva-fmt/src/extension.ts
```diff
@@ -82,4 +79,9 @@ export async function activate(context: vscode.ExtensionContext) {
     )
   );
   vscode.window.onDidChangeActiveTextEditor(inferer.updateState);
+  vscode.workspace.onDidChangeConfiguration((e) => {
+    if (e.affectsConfiguration('calva.fmt.formatAsYouType')) {
```
bpringe, line 83: Isn't this setting removed now?

PEZ: Might be a merge thing that I missed... I vaguely remember something with a strange merge when doing something recently.

Issue #1969, opened 2022-11-19
Make our nrepl client behave better


PEZ:
Currently Calva does just hammer the nREPL server with messages, even though the server has informed about which ops are supported. Calva starts the each session with asking about this (the `describe` op), but then we disregard this.

I think we should do two things, where the first one is more important.

1. Update the op methods in our client so that they do not bother the server with ops it doesn't support.
2. Update the call sites so that we can take proper action on missing ops.

PR #1970, opened 2022-11-19
Polite nrepl server

PEZ:
## What has changed?

Actually using the nREPL `describe` results to not spam the server with ops it doesn't handle anyway. Now the ops calls resolve to (or return, depending) `undefined` if they are not supported.

In some places I have added checks for this undefined. Not sure why that wasn't needed before, so I am pretty sure I have missed some places. But at least now I don't see any errors thrown in the console and things seem to work. Actually, they work better than before. E.g. Babashka info lookups were not succeeding because a `list-ns` call is not supported.

Fixes
* #1969

## My Calva PR Checklist

I have:

- [x] Tests

### DIFF
--- a/src/connector.ts
+++ b/src/connector.ts
@@ -635,19 +635,25 @@ export async function connect(
     console.error(e);
   }
   if (!['babashka', 'nbb', 'joyride', 'generic'].includes(connectSequence.projectType)) {
-    await nClient.session.info('clojure.core', 'map').catch((e) => {
-      void vscode.window
-        .showWarningMessage(
-          "Calva failed to perform a basic nREPL 'info' call. You need to start the REPL with cider-nrepl dependencies met",
-          'Show Calva Connect Docs'
-        )
-        .then((choice) => {
-          if (choice === 'Show Calva Connect Docs') {
-            void vscode.commands.executeCommand('simpleBrowser.show', 'https://calva.io/connect/');
-          }
-        });
-      console.error(`cider-nrepl dependencies not met: `, e);
-    });
+    // Give the describe a chance to run
+    setTimeout(() => {
+      if (!nClient.session.supports('info')) {
+        void vscode.window
+          .showWarningMessage(
+            'The nREPL server does not support cider-nrepl `info` op, which indicates troubles ahead. You need to start the REPL with cider-nrepl dependencies met',
+            'Show Calva Connect Docs'
+          )
+          .then((choice) => {
+            if (choice === 'Show Calva Connect Docs') {
+              void vscode.commands.executeCommand(
+                'simpleBrowser.show',
+                'https://calva.io/connect/'
+              );
+            }
+          });
+        console.error(`Basic cider-nrepl dependencies not met (no 'info' op)`);
+      }
+    }, 1000);
   }
   return true;
 }
--- a/src/connector.ts
+++ b/src/connector.ts
@@ -639,7 +639,7 @@ export async function connect(
       if (!nClient.session.supports('info')) {
         void vscode.window
           .showWarningMessage(
-            'The nREPL server does not support cider-nrepl `info` op, which indicates troubles ahead. You need to start the REPL with cider-nrepl dependencies met',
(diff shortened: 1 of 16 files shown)


PEZ:
Noticing that the testRunner now runs namespace tests twice if the tests lives in a `-test` namespace and you run them while the non-test namespace is the active file. I don't understand why this wasn't happening before, though. The `test-var-query` calls look different, which is also strange.

PEZ:
Still not sure what was going on, but I think we were compensating for some bug by running tests for both the ns and for it's `<ns>-test` namespace. Now `run namespace tests` runs the tests for the current ns, loading `<ns>-test` first.

src/connector.ts
```diff
@@ -634,21 +632,28 @@ export async function connect(
...
-    });
-  }
+  // Give the describe a chance to run
+  setTimeout(() => {
+    initializeDebugger(nClient.session);
+    if (!['babashka', 'nbb', 'joyride', 'generic'].includes(connectSequence.projectType)) {
+      if (!nClient.session.supports('info')) {
+        void vscode.window
+          .showWarningMessage(
+            'The nREPL server does not support cider-nrepl `info` op, which indicates troubles ahead. You need to start the REPL with cider-nrepl dependencies met',
```
bpringe:
```suggestion
            'The nREPL server does not support cider-nrepl `info` op, which indicates troubles ahead. You need to start the REPL with cider-nrepl dependencies met.',
```

src/namespace.ts
```diff
@@ -75,7 +75,7 @@ export async function createNamespaceFromDocumentIfNotExists(doc) {
       const client = replSession.getSession(utilities.getFileType(document));
       if (client) {
         const nsList = await client.listNamespaces([]);
-        if (nsList['ns-list'] && nsList['ns-list'].includes(ns)) {
+        if (nsList && nsList['ns-list'] && nsList['ns-list'].includes(ns)) {
```
bpringe, line 78: I think the above two lines could be a bit more simplified with optional chaining (I saw another case where this could be used in this PR too).

```typescript
        const nsList = await client.listNamespaces([])?.nsList;
        if (nsList?.includes(ns)) {
```

bpringe: Though this is arguably more clear, I suppose - just a bit less code.

src/nrepl/index.ts
```diff
@@ -393,21 +431,32 @@ export class NReplSession {
...
 
   stdin(message: string) {
-    this.client.write({
+    const msg = {
       op: 'stdin',
       stdin: message,
       session: this.sessionId,
-    });
+    };
+    if (this.supports('stdin')) {
```
bpringe: I think it would be better to check for support inside of the `write` function, rather than having to check in every function that calls `write`. Then this if check + log statement (if the op isn't supported), is only in one place.

bpringe: Maybe make `write` return `true` if the op is supported and was called or `undefined` if it's not and was not, then return the result of `write` (or pass the result to `resolve`) in these op calling functions.

src/namespace.ts
PEZ: It looked like that at first, but then I changed it for reasons I don't remember. =)

src/nrepl/index.ts
PEZ: I had it in the write function at first. But then it looked wrong to do a bunch of maybe-writes, and I also had some problems with it. The functions need to resolve with the result of the op, so I still had logic in the functions checking the result of the write...

But I reverted to this way without investigating the options to press on with the centralized approach. I guess if we pass on the resultResolver to write, it can handle this and we would have even less duplicated code. I'll have a look, but maybe we need to write it up on the todo list.

bpringe: Ah maybe it's not as simple as I was thinking. Also I see in some cases logic specific to the function occurs conditionally based on support for the op before write is called. There's probably a way to refactor it so put the support check in `write` but yeah... a todo item works for now.

PEZ: Yeah, I'm pretty sure it can be done, but all my attempts get very messy so far. Someone with better command over JavaScript than I have might see what should be done.

PEZ:
I need to update the publish script to stop me from releasing from a feature branch...

PR #2007, opened 2023-01-06
Handle completion response failures gracefully

### LINKED ISSUE #2006: Provider completions not handling errors gracefully
The provider completions subsystem requests results from the `lsp` and `nrepl` providers sequentially and then merges the results together. If either of these providers fail to report completion items and throws an error then the entire flow fails.

Ideally if one of the providers is working and is able to report results then the completions should continue to work.

I ran into this while dealing with an obscure issue which was causing my nrepl-based completions to fail. This issue isn't reproducible on my setup anymore, however the above problem can be reproduced simply by stopping the LSP server (and leaving the nrepl connected). I would expect completions to continue to work if I have an active `nrepl` connection but have stopped the LSP.

The fix should be to simply handle provider response failures more gracefully. I have an incoming PR which addresses this.

julienvincent:
The provider completions subsystem requests results from the `lsp` and `nrepl` providers sequentially and then merges the results together. If either of these providers fail to report completion items and throws an error then the entire flow fails.

Ideally if one of the providers is working and is able to report results then the completions should continue to work.

This commit fixes this issue by catching errors from the completions provider and quietly logging them to console instead of bubbling the error upstream.

Fixes #2006

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Changes to Calva.
 
 ## [Unreleased]
 
+- Fix: [Provider completions not handling errors gracefully](https://github.com/BetterThanTomorrow/calva/issues/2006)
+
 ## [2.0.322] - 2022-12-14
 
 - Fix: [Clojure notebooks don't seem to work on MS-Windows](https://github.com/BetterThanTomorrow/calva/issues/1994)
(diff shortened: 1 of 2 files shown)


PEZ (approved):
Awesome. Many thanks! 🙏

PR #2013, opened 2023-01-09
Don't block on snippet evaluation

### LINKED ISSUE #2012: Evaluating blocking snippets deadlocks the editor
If a snippet is evaluated with `runCustomREPLCommand` and it blocks (due to a bug or some such) then the editor almost completely deadlocks.

I found that running the `calva.interruptAllEvaluations` resolves the issue but it is still rather unexpected.

julienvincent:
If a snippet is evaluated with runCustomREPLCommand and it blocks (due to a bug or some such) then the editor almost completely deadlocks.

This commit changes the command handler for `runCustomREPLCommand` to not block on snippet evaluation.

Fixes #2012

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Changes to Calva.
 
 ## [Unreleased]
 
+- Fix: [Evaluating blocking snippets deadlocks the editor](https://github.com/BetterThanTomorrow/calva/issues/2012)
+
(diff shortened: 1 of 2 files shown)


PEZ:
Cool! Many thanks for your quality focus!

I will have a look later tonight, I hope.

PEZ (approved):
Thanks! LGTM.

PR #2017, opened 2023-01-11
Fix test on save

### LINKED ISSUE #2005: Setting calva.testOnSave broken: no tests found
It can't find tests anymore when saving the main code file. Saving the test file itself runs the tests fine, though. This worked in 2.0.319; it might have broken in https://github.com/BetterThanTomorrow/calva/commit/07add0b4f3dde630931c38c33fe286fad87e4db4.

Since the use case for this is to automatically run tests when changes are made to the code that needs testing, this should work.

bpringe:
## What has changed?

Fixed the test-on-save feature.

I made Calva run tests for the source namespace too because tests could be defined there, such as:

```clojure
(defn my-function
  "this function adds two numbers"
  {:test #(do
            (assert (= (my-function 2 3) 5))
            (assert (= (my-function 99 4) 8)))}
  ([x y] (+ x y)))
```

~However, with the above, I get an error in the output window - `; TypeError: Cannot read properties of undefined (reading 'filter')` - so I need to look into that when I get back to this.~

The above is now fixed by the check for `undefined` in `saveStacktrace`.

~I think we should probably also allow `testNs` to take an array of namespaces (and change its name to something like `testNamespaces`) since the `ns-query`'s `exactly` prop can take a collection of namespaces. Then we can avoid mapping over namespaces and dealing with an array of promises.~

Maybe the above can wait until later and we can keep the changes smaller in this PR?

I updated the test reporting for running namespace tests. We now say list the namespaces being tested.

I also fixed an issue which caused an error to be printed with `: ` in front of it. See the filtering related to `typeof m === 'string'`. In the problem case, the first item in `msg` was an empty array, for some reason, and the second was a string. This problem happened when I had a failing assertion in the metadata of a function in the source namespace (see the pic below which shows the case after the fix).

Example of running tests from a source namespace which also has test assertions in the metadata of a function, one of which fails.

image

Fixes #2005 

## My Calva PR Checklist

- [x] Tests
  - ~[ ] Tested the VSIX built from the PR (so, after you've submitted the PR). You'll find the artifacts by clicking _Show all checks_ in the CI section of the PR page, and then _Details_ on the `ci/circleci: build` test.~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,3 +91,6 @@ src/calva-fmt/atom-language-clojure/.bash_history
 
 # Leiningen
 .lein-repl-history
+
+# Portal
+.portal
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Changes to Calva.
 
 ## [Unreleased]
 
+- Fix: [Setting calva.testOnSave broken: no tests found](https://github.com/BetterThanTomorrow/calva/issues/2005)
+
 ## [2.0.324] - 2023-01-15
 
 - Fix: [Evaluating blocking snippets deadlocks the editor](https://github.com/BetterThanTomorrow/calva/issues/2012)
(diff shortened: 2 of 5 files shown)


PEZ:
> Maybe the above can wait until later and we can keep the changes smaller in this PR?

Sounds great. Issue please. 😄

PEZ (approved):
Looks great!

PR #2020, opened 2023-01-12
LSP support for multi-project and multi-workspace

julienvincent:
Right now Calva only supports opening clojure projects from the root of the project. This will be where the LSP process is started and the clojure LSP needs a deps.edn/project.clj file to exist at it's root in order to provide proper completions.

This also means that multi-workspace setups in VSCode is not really supported.

This commit refactors the LSP implementation to allow provisioning multiple LSP clients at different roots. Dependent systems can then request the LSP client governing a particular file/directory.

This change simultaneously allows opening directories that contain multiple projects as well as setting up multi-workspace vscode configurations.

The LSP initialization logic is also altered such that the server is started when a clojure file is opened rather than when Calva initializes. This prevents poluting non-clojure projects with `.lsp` and `.clj-kondo` directories.

---

### Commands

The LSP lifecycle commands have been reworked to make a bit more sense within the new architecture:

The `start` command now list all discovered clojure project roots in the workspace, allowing the user to pick where they want to start an LSP client.

This is especially useful in mono-repos where a user might want to pick if they want to start the LSP at the project root or within a specific sub-package (or maybe a combination of both).

The `restart/stop` commands now list the directories mapping to running LSP clients and allows the user to select the client they want to restart/stop.

### Status Bar and Menu

The lsp status bar item now tracks the currently active editor. As the user changes the active window the status bar will show the status of the lsp governing the file they have open.

When the user clicks the status bar item we now show a single type of LSP menu instead of dynamically changing it based on the status of the client. This is because there are now multiple LSP clients running.

## Checklist

- [x] Read [How to Contribute](https://github.com/BetterThanTomorrow/calva/wiki/How-to-Contribute#before-sending-pull-requests).
- [x] Directed this pull request at the `dev` branch. (Or have specific reasons to target some other branch.)
- [x] Made sure I have changed the PR base branch, so that it is not `published`. (Sorry for the nagging.)
- [ ] Updated the `[Unreleased]` entry in `CHANGELOG.md`, linking the issue(s) that the PR is addressing.
- [x] Figured if **anything** about the fix warrants tests on Mac/Linux/Windows/Remote/Whatever, and either tested it there if so, or mentioned it in the PR.
- [x] Added to or updated docs in this branch, if appropriate
- [ ] Tests
  - [ ] Tested the particular change
  - [ ] Figured if the change might have some side effects and tested those as well.
  - [ ] Smoke tested the extension as such.
[...]

PR #2046, opened 2023-01-30
Fix clojure-lsp regressions from #2020

julienvincent:
This is a collection of fixes which should resolve the regressions and issues reported as a result of #2020.

- When using the `"when-workspace-opened-use-workspace-root"` setting the clojure-lsp server is only started in valid clojure workspaces.
- Ensures that operations like `Organise Imports` continue to work. This was an unforeseen  side-effect of disabling the dynamic lsp command registrations the client was performing. This default behaviour is still disabled, but now we manually register the commands (once) and augment them with support for selecting the correct lsp client to forward the request to.
- The project selection menus have been cleaned up and had some behavioural tweaks. See below:

### Project Menus

There were a few changes made to the project menus:

- The second line containing `details` on why the project was considered has been removed.
- Projects are now grouped by their workspace root which should help make it easier to sort through the list visually.
- Project folders are now sorted within their groups to make the list more deterministic.
- Project folders show the path relative to the workspace root instead of the absolute path. This should help make scanning the list easier by removing unnecessary/duplicate information.

When sorting projects the behaviour is as follows:

- The workspace group containing the preselected project is sorted to the top.
- The projects within each group are sorted such that the workspace root is first, the preselected item is second and then the remaining items are sorted by path length.

image

# Checklist

- [ ] Tests

### DIFF
--- a/src/lsp/commands/lsp-commands.ts
+++ b/src/lsp/commands/lsp-commands.ts
@@ -221,7 +221,7 @@ export function registerLspCommands(clients: defs.LSPClientMap) {
     ...clojureLspCommands.map((command) => registerUserspaceLspCommand(clients, command)),
 
     /**
-     * The clojure-lsp server previously used to dynamically register all of these top-level commands however that behaviour was
+     * The clojure-lsp server previously used to dynamically register all of these top-level commands. However, that behaviour was
      * disabled to add support for provisioning multiple lsp clients in the same vscode window. The built-in behaviour resulted
      * in command registration conflicts.
      *
--- a/src/lsp/commands/lsp-commands.ts
+++ b/src/lsp/commands/lsp-commands.ts
@@ -1,3 +1,4 @@
+import * as vscode_lsp from 'vscode-languageclient/node';
 import * as calva_utils from '../../utilities';
 import * as defs from '../definitions';
 import * as vscode from 'vscode';
@@ -53,16 +54,37 @@ function sendCommandRequest(
   }
 
   client
-    .sendRequest('workspace/executeCommand', {
+    .sendRequest(vscode_lsp.ExecuteCommandRequest.type, {
       command,
       arguments: args,
+    })
+    .catch((error) => {
+      return client.handleFailedRequest(
+        vscode_lsp.ExecuteCommandRequest.type,
+        undefined,
+        error,
+        undefined
+      );
     })
     .catch((e) => {
       console.error('Failed to execute lsp command', e);
     });
 }
 
-function registerLspCommand(
+const getLSPCommandParams = () => {
+  const editor = calva_utils.getActiveTextEditor();
+  const document = calva_utils.tryToGetDocument(editor.document);
+  if (!document || document.languageId !== 'clojure') {
+    return;
+  }
+
+  const line = editor.selection.start.line;
+  const column = editor.selection.start.character;
+  const doc_uri = `${document.uri.scheme}://${document.uri.path}`;
+  return [doc_uri, line, column];
+};
+
+function registerUserspaceLspCommand(
   clients: defs.LSPClientMap,
   command: ClojureLspCommand
 ): vscode.Disposable {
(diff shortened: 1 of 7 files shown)


PEZ:
Thanks! To my testing this works as you advertise.

I think that tucking the project file to the end of the path can get a bit confusing. E.g. here the *shadow-w-backend* project has several project files, `deps.edn`, `project.clj`, and `shadow-cljs.edn`. As a user I wonder what does it mean that `deps.edn` is included in the path? 

image

julienvincent:
> I think that tucking the project file to the end of the path can get a bit confusing. E.g. here the _shadow-w-backend_ project has several project files, `deps.edn`, `project.clj`, and `shadow-cljs.edn`. As a user I wonder what does it mean that `deps.edn` is included in the path?

Oops this is a bug :) will fix.

Resolved in 13d539fc

PEZ:
> * When using the `"when-workspace-opened-use-workspace-root"` setting the clojure-lsp server is only started in valid clojure workspaces.

This changes how it behaves today when opening a folder w/o a Clojure project at the root. Not sure we should necessarily keep that behaviour (ought to be pretty uncommon, and clojure-lsp help gets pretty limited). WDYT, @julienvincent @bpringe @ericdallo?

julienvincent:
I also expect it would be pretty uncommon. As a fallback users can always explicitly start the clojure-lsp server in the root if they want as well.

julienvincent:
@PEZ Are we concerned about this/anything else? Anything outstanding before this can be ready for merge?

src/lsp/commands/lsp-commands.ts
```diff
@@ -193,7 +218,19 @@ export function registerLspCommands(clients: defs.LSPClientMap) {
       });
     }),
 
-    ...clojureLspCommands.map((command) => registerLspCommand(clients, command)),
+    ...clojureLspCommands.map((command) => registerUserspaceLspCommand(clients, command)),
+
+    /**
+     * The clojure-lsp server previously used to dynamically register all of these top-level commands however that behaviour was
```
bpringe:
```suggestion
     * The clojure-lsp server previously used to dynamically register all of these top-level commands. However, that behaviour was
```

src/project-root.ts
```diff
@@ -123,41 +136,100 @@ export function findFurthestParent(from: vscode.Uri, uris: vscode.Uri[]) {
...
+  }, []);
+};
+
+function sortPreSelectedFirst(groups: vscode.Uri[][], selected: vscode.Uri) {
+  // First sort the groups, bringing the group containing the preselected item to the top.
+  const sorted_groups = groups.sort((a, b) => {
+    if (!selected) {
+      return 0;
+    }
+    function groupContainsUri(uris: vscode.Uri[]) {
```
bpringe: I'm not sure if it matters much here (if at all), but couldn't this be defined outside the `sort` callback (but still inside `sortPreSelectedFirst`?

bpringe (commented):
Just left a suggestion and question.

The changes look good to me overall, though I haven't tested them. If @PEZ thinks it's good to go when he has a chance to look/test, then let's merge 🚀.

Thanks for the fixes!

src/project-root.ts
julienvincent: I moved it out to the top level and made it take `selected` as a second arg.

PEZ:
I think this is good to go. Will merge in a jiffy. Thanks again. I love working with you, @julienvincent!

A thing we can follow up with in a PR is to add a reason to the pre-selected project. In the jack-in/connect project menu, it is pulled out of, and breaks, the sort order. For good reasons, but the user isn't told those. 

image

In the screenshot the user is told a ”reason” for the first entry, it's the Workspace root. The selected item could say something like *Closest Project*. Yeah, that's the reason why it is pre-selected, not why it is sorted where it is, but anyway.

I like that the workspace root is sorted to the top, btw. It makes sense and chimes with the rest of the ordering.

src/project-root.ts
bpringe: Great!

PR #2062, opened 2023-02-04
Move docs PR checklist to a separate template

PEZ:
Updating the PR template with some of the things we discussed.

Mainly I decided to move the docs stuff to its own template anyway, bad Ux or not. It is just cleaner this way.

WDYT?

### DIFF
--- a/docs/PULL_REQUEST_TEMPLATE.md
+++ b/docs/PULL_REQUEST_TEMPLATE.md
@@ -1,4 +1,13 @@
-<!-- ❤️ Thanks for filing a Pull Request on Calva! You are contributing to a better Clojure coding experience. ❤️
+<!--
+❤️ Thanks for filing a Pull Request on Calva! You are contributing to a better Clojure coding experience. ❤️
+
+PLEASE NOTE:
+If you want to file a Pull Request on the documentation of Calva (calva.io),
+then use the Documentation PR template by adding 'template=docs.md' to the
+query parameters of the URL of this page.
+
+The rest of this template is about changes to the Calva source code.
+
 ## What you can expect:
 
 Here are some things we consider before we merge:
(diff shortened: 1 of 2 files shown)


docs/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,4 +1,13 @@
-<!-- ❤️ Thanks for filing a Pull Request on Calva! You are contributing to a better Clojure coding experience. ❤️
+<!--
+❤️ Thanks for filing a Pull Request on Calva! You are contributing to a better Clojure coding experience. ❤️
+
+PLEASE NOTE:
+If you want to file a Pull Request on the documentation of Calva (calva.io),
+then use the Documentation PR template by adding 'template=docs.md' to the
+query parameters of the URL of this page.
```
bpringe, line 7: I wonder about this adding more friction, even if it's a small ask, but we'll see if it's problematic. I like having a separate template, anyway.

docs/PULL_REQUEST_TEMPLATE.md
```diff
@@ -55,8 +57,7 @@ If this PR involves code changes, I have:
 - [ ] Tests
   - [ ] Tested the particular change
   - [ ] Figured if the change might have some side effects and tested those as well.
-  - [ ] Smoke tested the extension as such.
-  - [ ] Tested the VSIX built from the PR (so, after you've submitted the PR). You'll find the artifacts by clicking _Show all checks_ in the CI section of the PR page, and then _Details_ on the `ci/circleci: build` test.
+  - [ ] Sanity tested the extension with my changes.
```
bpringe: Can we add a bit more clarity/guidance here? I can imagine people wondering what exactly this means.

If sanity testing is different depending on the changes made, then we can state that and/or state an example. Like "For example, if the changes are related to jack-in or could affect it, test that jack-in works."

If it's more about testing a set of basic features, maybe we can list them out.

PEZ: We could. I think it's OK if people interpret it differently, though. (But, I don't want them to get confused.)

PEZ: I decided to remove it. It is sort of covered by the remaining items, and CI makes some smoke testing anyway. It's better we ask for specific tests if we start to worry about something in particular in a PR.

PR #2073, opened 2023-02-09
Jack-in without prompt

### LINKED ISSUE #2049: Establish REPL connection without being prompted for the kind of project every time
On every jack in, Calva prompts the user for the kind of project being used (`deps.edn`, `joyride`, etc.).
The suggestion is to introduce a mechanism that allows this to be configured once, potentially via a project-level setting. 

If this feature were to support selecting custom REPL sequences as the kind of project, it would allow users to jack in without being prompted at all. For that, they would have to use this setting in combination with the `menuSelections` attribute from the custom REPL sequence.

SillyCoon:
## What has changed?

Fixes #2049 

Now you can set up the default project for Jack-in via config or after the Jack-in run by click on `Yes`.
I'm not sure that it's great UI/UX but it's pretty simple and the information window doesn't block other input.
Снимок экрана_20230213_203329
Снимок экрана_20230213_203538
Снимок экрана_20230213_203621

## My Calva PR Checklist

I have:

- [x] Tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -375,10 +375,10 @@
             "default": {},
             "description": "Specifies any environment variables your project needs. (Probably mostly for your Workspace Settings.)"
           },
-          "calva.jackInProjectType": {
+          "calva.autoSelectReplConnectProjectType": {
             "type": "string",
             "default": null,
-            "description": "Specifies the project type for Jack-in by default. (Probably mostly for your Workspace Settings.)"
+            "description": "When set Calva will not prompt for a project type, at Jack-in or Connect. Instead the project name specified here will be used. (Probably mostly for your Workspace Settings.)"
           },
           "calva.jackInDependencyVersions": {
             "type": "object",
--- a/package.json
+++ b/package.json
@@ -375,6 +375,11 @@
             "default": {},
             "description": "Specifies any environment variables your project needs. (Probably mostly for your Workspace Settings.)"
           },
+          "calva.jackInProjectType": {
+            "type": "string",
+            "default": null,
+            "description": "Specifies the project type for Jack-in by default. (Probably mostly for your Workspace Settings.)"
+          },
           "calva.jackInDependencyVersions": {
             "type": "object",
             "description": "Versions of the dependencies injected by Calva Jack-in",
--- a/src/nrepl/connectSequence.ts
+++ b/src/nrepl/connectSequence.ts
@@ -368,7 +368,7 @@ function getDefaultSequence(
       return defaultSequence;
     } else {
       void vscode.window.showInformationMessage(
-        `No such project "${defaultProject}" available for Jack-in, please check your config.`
+        `No such project type - "${defaultProject}" - available for jack-in. Please check your settings.`
       );
     }
   }
--- a/src/nrepl/connectSequence.ts
+++ b/src/nrepl/connectSequence.ts
@@ -358,7 +358,7 @@ function getDefaultSequence(
 
     if (defaultSequence) {
       void vscode.window.showInformationMessage(
-        `Used "${defaultSequence.name}" as a default project for Jack-in.`
+        `Used "${defaultSequence.name}" as a default project type for jack-in.`
       );
       void state.extensionContext.workspaceState.update(
         `d-${saveAsPath}`,
--- a/src/nrepl/connectSequence.ts
+++ b/src/nrepl/connectSequence.ts
@@ -328,7 +328,8 @@ function getConnectSequences(projectTypes: string[]): ReplConnectSequence[] {
 
 async function askToSetDefaultProjectForJackIn(project: string) {
   const setAsDefault = await utilities.showBooleanInformationMessage(
-    `Do you want to set "${project}" as the jack-in project type for the current workspace?`
+    `Do you want to set "${project}" as the jack-in project type for the current workspace?`,
+    'autoSelectReplConnectProjectTypeNotShow'
   );
 
   if (setAsDefault) {
--- a/src/nrepl/connectSequence.ts
+++ b/src/nrepl/connectSequence.ts
@@ -288,12 +288,15 @@ const defaultCljsTypes: { [id: string]: CljsTypeConfig } = {
 const autoSelectProjectTypeSetting: `calva.${keyof Config}` =
   'calva.autoSelectReplConnectProjectType';
 
+const autoSelectDocLink = `  - See https://calva.io/connect/#auto-select-project-type`;
+
 const defaultProjectSettingMsg = (project: string) =>
   formatAsLineComments(
     [
-      `If you want to set the selected project type as the default for the current workspace`,
-      `please add this setting to your .vscode/settings.json file:`,
-      `  "${autoSelectProjectTypeSetting}": "${project}"`,
+      `Connecting using "${project}" project type.`,
+      `You can make Calva auto-select this:`,
+      `"${autoSelectProjectTypeSetting}": "${project}"`,
+      autoSelectDocLink,
       '\n',
     ].join('\n')
   );
@@ -369,9 +372,9 @@ function getUserSpecifiedSequence(
       outputWindow.appendLine(
         formatAsLineComments(
           [
-            `Used "${defaultSequence.name}" as a default for jack-in.`,
-            `If you want to change it, please remove "${autoSelectProjectTypeSetting}"`,
-            `  from settings.json and re-run the jack-in command.`,
+            `Auto-selecting project type "${defaultSequence.name}".`,
+            `You can change this from settings:`,
+            autoSelectDocLink,
             '\n',
           ].join('\n')
         )
@@ -386,8 +389,9 @@ function getUserSpecifiedSequence(
       outputWindow.appendLine(
         formatAsLineComments(
           [
-            `No such project type - "${userSpecifiedProjectType}" - available for jack-in.`,
-            `Please check your "${autoSelectProjectTypeSetting}" settings.`,
+            `Project type "${userSpecifiedProjectType}" not found.`,
+            `You need to update the auto-select setting.`,
+            autoSelectDocLink,
             '\n',
           ].join('\n')
         )
--- a/src/nrepl/connectSequence.ts
+++ b/src/nrepl/connectSequence.ts
@@ -327,7 +327,7 @@ function getConnectSequences(projectTypes: string[]): ReplConnectSequence[] {
 
 async function askToSetDefaultProjectForJackIn(project: string) {
   const setAsDefault = await utilities.showBooleanInformationMessage(
-    `Do you want to set "${project}" as a default Jack-in project type for the current workspace?`
+    `Do you want to set "${project}" as the jack-in project type for the current workspace?`
   );
 
   if (setAsDefault) {
--- a/src/nrepl/connectSequence.ts
+++ b/src/nrepl/connectSequence.ts
@@ -325,6 +325,16 @@ function getConnectSequences(projectTypes: string[]): ReplConnectSequence[] {
   return sequences;
 }
 
+async function askToSetDefaultProjectForJackIn(project: string) {
+  const setAsDefault = await utilities.showBooleanInformationMessage(
+    `Do you want to set "${project}" as a default Jack-in project type for the current workspace?`
+  );
+
+  if (setAsDefault) {
+    await vscode.workspace.getConfiguration('calva').update('jackInProjectType', project);
+  }
+}
+
 /**
  * Returns the CLJS-Type description of one of the build-in.
  * @param cljsType Build-in cljsType
@@ -335,6 +345,35 @@ function getDefaultCljsType(cljsType: string): CljsTypeConfig {
   return defaultCljsTypes[cljsType];
 }
 
+function getDefaultSequence(
+  sequences: ReplConnectSequence[],
+  saveAsPath: string
+): ReplConnectSequence | undefined {
+  const defaultProject = getConfig().jackInProjectType;
+
+  if (defaultProject) {
+    const defaultSequence = sequences.find(
+      (s) => s.name.toLocaleLowerCase() === defaultProject.toLocaleLowerCase()
+    );
+
+    if (defaultSequence) {
+      void vscode.window.showInformationMessage(
+        `Used "${defaultSequence.name}" as a default project for Jack-in.`
+      );
+      void state.extensionContext.workspaceState.update(
+        `d-${saveAsPath}`,
+        defaultSequence.projectType
+      );
+
+      return defaultSequence;
+    } else {
+      void vscode.window.showInformationMessage(
+        `No such project "${defaultProject}" available for Jack-in, please check your config.`
+      );
+    }
+  }
+}
+
 async function askForConnectSequence(
   cljTypes: string[],
   saveAs: string,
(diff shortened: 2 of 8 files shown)


SillyCoon:
Hi @PEZ! Please review this PR. I also added an integration test for it and I guess I need to edit the documentation. I wanted to record the video, but actually I was not able to do it today, so I can incude it somewhere tomorrow.

PEZ:
Cool! I hadn't thought about this way to do it. I was stuck in how this used to work. I think I like this way better, because it is so straightforward.

We need to be a bit careful with pop-ups. This one is very helpful, but I can see how it could get annoying for someone closing it without selecting an option. Some alternatives:

1. We could include information about the setting in the pop-up + add a button for ”Don't ask again”, which would mean ”Don't ask again for this workspace”. 
1. Include settings info in the pop-up and always only ask once per workspace.
1. Don't pop-up anything. We inform about it in the docs.
1. Don't pop-up anything. We inform about it in the docs, plus in the Output/REPL Window.

I haven't tested this myself yet, so here questions:

1. Does it work with Custom REPL Connect Sequences?
1. Does it also suppress the menu when connecting to a running REPL? I think it should, and that the setting probably should be renamed.

I requested review by @bpringe now, so that we can get his opinions on the Ux.

This is great work! 🙏

SillyCoon:
> Cool! I hadn't thought about this way to do it. I was stuck in how this used to work. I think I like this way better, because it is so straightforward.
> 
> We need to be a bit careful with pop-ups. This one is very helpful, but I can see how it could get annoying for someone closing it without selecting an option. Some alternatives:
> 
> 1. We could include information about the setting in the pop-up + add a button for ”Don't ask again”, which would mean ”Don't ask again for this workspace”.
> 2. Include settings info in the pop-up and always only ask once per workspace.
> 3. Don't pop-up anything. We inform about it in the docs.
> 4. Don't pop-up anything. We inform about it in the docs, plus in the Output/REPL Window.
> 
> I haven't tested this myself yet, so here questions:
> 
> 1. Does it work with Custom REPL Connect Sequences?
> 2. Does it also suppress the menu when connecting to a running REPL? I think it should, and that the setting probably should be renamed.
> 
> I requested review by @bpringe now, so that we can get his opinions on the Ux.
> 
> This is great work! 🙏

Thank you for the great detailed answer!

I agree about pop-ups, it's a kinda slippery road. What do you mean by not pop-up, do you mean to set the selected project as a default for this workspace if we call Jack-in?

About testing. It should work with custom sequences and I guess it doesn't suppress the menu, but I will check it twice in a day or two and will let you know.
Thanks!

PEZ:
Now testing this quickly:

* Connecting I get this:
    image
    I don't think we should pop it up there. We can write it in the Output Window instead.
* When pressing escape out of the Project Type menu, jack-in is aborted, but the question to save the default project type still pops up.



> 1. Does it work with Custom REPL Connect Sequences?

Yes it does. ✅ 


> 2. Does it also suppress the menu when connecting to a running REPL? I think it should, and that the setting probably should be renamed.

Yes, this works for Connect as well as for Jack-in. As it should. I suggest the setting be named `calva.autoSelectReplConnectProjectType` and the description could be: _When set Calva will not prompt for a project type, at Jack-in or Connect. Instead the project name used here will be used. (Probably mostly for your Workspace Settings.)_

> What do you mean by not pop-up, do you mean to set the selected project as a default for this workspace if we call Jack-in?

I mean to leave it up to the user to add the setting. If we do this we probably want to inform about the possibility in the Output window when connecting the REPL.

SillyCoon:
> Now testing this quickly:
> 
> * Connecting I get this:
>   image
>   I don't think we should pop it up there. We can write it in the Output Window instead.
> * When pressing escape out of the Project Type menu, jack-in is aborted, but the question to save the default project type still pops up.
> 
> > 1. Does it work with Custom REPL Connect Sequences?
> 
> Yes it does. ✅
> 
> > 2. Does it also suppress the menu when connecting to a running REPL? I think it should, and that the setting probably should be renamed.
> 
> Yes, this works for Connect as well as for Jack-in. As it should. I suggest the setting be named `calva.autoSelectReplConnectProjectType` and the description could be: _When set Calva will not prompt for a project type, at Jack-in or Connect. Instead the project name used here will be used. (Probably mostly for your Workspace Settings.)_
> 
> > What do you mean by not pop-up, do you mean to set the selected project as a default for this workspace if we call Jack-in?
> 
> I mean to leave it up to the user to add the setting. If we do this we probably want to inform about the possibility in the Output window when connecting the REPL.

Thank you very much for the testing and quick reply! I will fix the issues soon. 

About UX. Personally, I prefer the option with pop-up suppression (Don't show it again for the current workspace), but it would be nice to hear other contributors' opinions as well.  

The problem with the manual setting is that we can't suggest only available sequences for the project to choose from. My original thought was to provide the dropdown with the options in the config, but it seems impossible to do since VS Code doesn't support changing the list of available options in settings dropdowns in runtime (or I didn't find a way to do it).

package.json
```diff
@@ -375,6 +375,11 @@
             "default": {},
             "description": "Specifies any environment variables your project needs. (Probably mostly for your Workspace Settings.)"
           },
+          "calva.jackInProjectType": {
+            "type": "string",
+            "default": null,
+            "description": "Specifies the project type for Jack-in by default. (Probably mostly for your Workspace Settings.)"
```
bpringe: > (Probably mostly for your Workspace Settings.)

Can you elaborate on what this means?

src/nrepl/connectSequence.ts
```diff
@@ -325,6 +325,16 @@ function getConnectSequences(projectTypes: string[]): ReplConnectSequence[] {
   return sequences;
 }
 
+async function askToSetDefaultProjectForJackIn(project: string) {
+  const setAsDefault = await utilities.showBooleanInformationMessage(
+    `Do you want to set "${project}" as a default Jack-in project type for the current workspace?`
```
bpringe:
```suggestion
    `Do you want to set "${project}" as the jack-in project type for the current workspace?`
```

I think maybe we should avoid saying "default" since once it's set there's no picking another one from a menu - it's just the project type that's used.

src/nrepl/connectSequence.ts
```diff
@@ -335,24 +345,62 @@ function getDefaultCljsType(cljsType: string): CljsTypeConfig {
...
+  const defaultProject = getConfig().jackInProjectType;
+
+  if (defaultProject) {
+    const defaultSequence = sequences.find(
+      (s) => s.name.toLocaleLowerCase() === defaultProject.toLocaleLowerCase()
+    );
+
+    if (defaultSequence) {
+      void vscode.window.showInformationMessage(
+        `Used "${defaultSequence.name}" as a default project for Jack-in.`
```
bpringe:
```suggestion
        `Used "${defaultSequence.name}" as a default project type for jack-in.`
```

src/nrepl/connectSequence.ts
```diff
@@ -335,24 +345,62 @@ function getDefaultCljsType(cljsType: string): CljsTypeConfig {
...
+      );
+      void state.extensionContext.workspaceState.update(
+        `d-${saveAsPath}`,
+        defaultSequence.projectType
+      );
+
+      return defaultSequence;
+    } else {
+      void vscode.window.showInformationMessage(
+        `No such project "${defaultProject}" available for Jack-in, please check your config.`
```
bpringe:
```suggestion
        `No such project type - "${defaultProject}" - available for jack-in. Please check your settings.`
```

package.json
bpringe: I think we should specify that this allows bypassing the prompt for the project type at jack-in. Maybe we can say "Specifies the project type for jack-in, which causes the project type menu at jack-in to be bypassed."

src/nrepl/connectSequence.ts
```diff
@@ -335,24 +345,62 @@ function getDefaultCljsType(cljsType: string): CljsTypeConfig {
   return defaultCljsTypes[cljsType];
 }
 
+function getDefaultSequence(
```
bpringe: Considering my other comment about using the term "default", maybe we can change this to something like `getUserSpecifiedSequence`.

src/nrepl/connectSequence.ts
```diff
@@ -335,24 +345,62 @@ function getDefaultCljsType(cljsType: string): CljsTypeConfig {
...
+  sequences: ReplConnectSequence[],
+  saveAsPath: string
+): ReplConnectSequence | undefined {
+  const defaultProject = getConfig().jackInProjectType;
```
bpringe: Considering my other comments, maybe we should name this and similarly named variables added in this PR to something like `userSpecifiedProjectType`.

package.json
bpringe: @PEZ Do we need to list the possible values here somehow?

@SillyCoon I saw your comment about using a dropdown, but maybe we can use a dropdown and just list all the possible project types. It provides a more specific schema for the setting. In either case, they can put something that won't work for the current project, but at least with a dropdown the setting isn't as susceptible to a spelling mistake, even if they edit it in the json and not via the settings UI (VS Code will suggest the possible string values if it's a dropdown, I think).

bpringe:
@SillyCoon Thanks for the PR!

> I requested review by @bpringe now, so that we can get his opinions on the Ux.

@PEZ I think we can go with this:

> We could include information about the setting in the pop-up + add a button for ”Don't ask again”, which would mean ”Don't ask again for this workspace”.

It seems fairly simple and very little code to maintain overall, plus this PR is already geared toward this.

package.json
SillyCoon: Thank you for the comment! 
- (Probably mostly for your Workspace Settings.) - I copied it from calva.jackInEnv. In my understanding, it means that this option is useful only in your Workspace `settings.json`, and not as a setting for the entire VS Code
- I picked up the naming from here https://github.com/BetterThanTomorrow/calva/pull/2073#issuecomment-1425411118
- In this case, I'm afraid that dropdown doesn't support custom sequences defined by user

PEZ:
I'm still hesitating about the popup. Reasons:

* General popup-avoidance, in the past we have had complaints on Calva that it pops open too much.
* The jack-in and connect often prompts for more things after the project type and the popup then makes us prompt for two things at the same time. It gets confusing.
* For new users it is sometimes they don't really know which project type they should select.
  * Then this question adds cognitive load.
  * They risk accepting a setting that is not the one they should use. This can put them in a strange state where they don't get the menu again, and might not know how to get it back, since they didn't do it themselves.

I think that for now it is better we print a message in the Output window about adding the setting. We can print the value there, and the user can copy it.

bpringe:
@PEZ I agree with your reasons. I think we should avoid the pop-up as well.

package.json
bpringe: > I picked up the naming from here https://github.com/BetterThanTomorrow/calva/pull/2073#issuecomment-1425411118

That works.

> In this case, I'm afraid that dropdown doesn't support custom sequences defined by user

What do you mean here? I think the dropdown can just list all project types that Calva supports, so it wouldn't be defined by the user.

bpringe: I'd like @PEZ's opinion here though. If we add a new project type later we'd have to be sure to add to this dropdown, but if the dropdown pulls from the same list of project types that are used elsewhere, then adding a new one would automatically add the option to this dropdown. I don't know if the options are centralized like that at the moment, though.

PEZ: I'm not sure what is discussed here. I have tried the PR and custom sequences seem to work just fine.

bpringe: The user has to type out the string themselves for the project type. Typos would lead to an error, right? That's probably not a big deal though. I think we should at least list out the possible values for this setting in the docs.

bpringe: And of course some might not apply to certain projects, but we should state that as well.

PEZ: The possible values are not known, it depends on the project and what custom REPL sequences and stuff. We can mention in the description that the value to fill in will be printed in the Output/REPL window when they connect to the REPL.

SillyCoon:
Can't argue with that, thanks for your help! I will move the message to Output Window then, should be pretty easy

package.json
SillyCoon: Yeah, that's the main problem, we don't know all the possible projects to suggest until the runtime, and we cannot change the dropdown for the settings on runtime too, it seems to be one of the limitations of the VS code itself 😅 So the only simple way to be sure that we're putting the correct project to the setting is to use the `jack-in` dropdown, and to notify the user in the Output Window that the setting is incorrect if it was specified with a typo.

SillyCoon:
@PEZ @bpringe I've replaced a pop-up with the Output Window. Please check the new screenshots and correct my descriptions of the settings.

PEZ:
Thanks! I like your suggestions for Output window texts.

It is rather few people who read the output in this window, I think. In some other PR we can consider cleaning all that text it up and rather link to calva.io.

We might want to move the instructions for this setting there. Then we can elaborate a bit more on how it works. Not sure where, but I think that for now, just before this section on the **Connect the REPL** page: https://calva.io/connect/#monorepos-multiple-clojure-projects-in-one-workspace will do. Something like:

```markdown
## Auto-select Project Type

You can make both Jack-in and Connect stop prompting you for project type in projects where you always want to use the same. The setting `calva.autoSelectReplConnectProjectType` is used for this. Here's how to use it:

1. Connect the REPL and manually select the project type/connect sequence you want to use.
1. In the Output/REPL window Calva will print about auto-selecting project type, including the value you need. **Copy this value**
1. In VS Code Settings, search for ”Calva auto” and you will find the setting
1. Make sure you are editing Workspace settings
1. **Paste the project type value you copied above**

Next time you connect the REPL, Calva will use this value and the project type prompt will not be shown.

To make Calva prompt for the project type, reset this setting (there's a cog icon there giving you an option to do this).

!!! Or edit Workspace Settings JSON
    You can of course also edit this setting manually in `.vscode/settings.json` for the workspace. ”Reset” is then equivalent to removing the setting from this file.
```

Then in the Output REPL window we can link to https://calva.io/connect/#auto-select-project-type, with short prompts like:

> Connecting using "Leiningen” project type. You can make Calva auto-select this:
> * See \<link\>

> Auto-selecting project type "Leiningen". You can change this from settings
> * See \<link\>

> Project type "Bla bla” not found. You need to update the auto-select setting.
> * See \<link\>

For the last case there I think we should show the menu, which will then make us print the correct settings value.

WDYT?

package.json
bpringe: Oh, I forgot about custom repl sequences being in that list.

> We can mention in the description that the value to fill in will be printed in the Output/REPL window when they connect to the REPL.

This sounds good. 

> So the only simple way to be sure that we're putting the correct project to the setting is to use the jack-in dropdown, and to notify the user in the Output Window that the setting is incorrect if it was specified with a typo.

I'm wondering if the output window is the best place to inform them of an issue with the setting though, since the output window may not be open at that time. I think in most cases it will open when they start to jack-in, but some people may have the setting to not open the output window at jack-in enabled. Maybe _this_ is a place to use a pop-up.

bpringe: I just saw this comment from @PEZ:

> For the last case there I think we should show the menu, which will then make us print the correct settings value.

That makes sense to me.

SillyCoon:
Sounds great! I'll create the PR with the updated documentation later then. 
But I don't understand the last point. If the project type is incorrect, you should pick the correct one like there is no setting in the config. You can copy the new value to the config instead of the wrong one, I guess it works as you wrote but maybe I'm wrong.

PEZ:
> If the project type is incorrect, you should pick the correct one like there is no setting in the config. You can copy the new value to the config instead of the wrong one, I guess it works as you wrote but maybe I'm wrong.

Sounds exactly like what I had in mind.

PEZ:
I think the docs should be updated in this PR, btw.

SillyCoon:
> I think the docs should be updated in this PR, btw.

sorry, I forgot that the code and the docs are in the same repo, will do it in this PR 😌

SillyCoon:
I've updated the docs and output. 
As for me, it's useful to leave the setting string in the output directly to simplify copying, so it would be like this. What do you think?

image

PEZ:
Thanks so much for this work, @SillyCoon! It's interesting this with settings. I start with ”no” when someone suggests a setting, because they add more work than I often first calculate with. They add cognitive load for the users too. And in supporting users, a layer of ”which settings for X are you using?” to all other permeations of things. As we can see with this little setting there is a lot to consider right off the bat.

That said, I know we made the right call here making this configurable. When testing this I noticed how nice jack-in got. Let's go for that connect prompt as well, which asks for port with a pre-populated text input.

I noticed a problem with this setting, as well as almost all Calva settings, when testing this PR. I tested in a workspace where I had several root folders. When I have set a default for auto-selecting project type, that was used regardless of which folder I was jacking in to. Because I had set it with the **Workspace** settings... Trying to set it on **Folder** settings didn't work because we lack `"scope": "resource"` on the settings objects in the manifest. What's more. Even when I added that it still didn't work! Because we don't read folder settings... We should fix this. Connect Sequences and a lot of other settings make sense to add here. I think maybe all settings deserve this treatment. CC: @bpringe @julienvincent

PR #2077, opened 2023-02-10
Rollback clojure-lsp auto-start behaviour default

julienvincent:
This reverts the auto-start behaviour back to using the first workspace folder by default.

This is achieved by modifying the `calva.enableClojureLspOnStart` setting to include `"always-use-first-workspace-root"` and sets this as the default. The previous settings and associated behaviours still exist and can be enabled by changing the default.

This new default setting will only ever start a single clojure-lsp server, which will be started in the first workspace folder (or in an os tmp directory if the first workspace folder is not a valid clojure project).

This PR currently bases off of #2076 (and therefore depends on it) in order to take advantage of the tmp directory support, but if we want to inverse the order I can change that. As such, only the last commit here is relevant as the first few commits comes from #2076.

Long term I don't think this should remain the default, instead using the `"when-workspace-opened-use-workspace-root"` as the default - but we can discus changing the default in a future PR.

### Checklist

- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Changes to Calva.
 
 ## [Unreleased]
 
+- [Rollback clojure-lsp auto-start behaviour default](https://github.com/BetterThanTomorrow/calva/pull/2077)
 - Fix: [clojure-lsp does not start in a workspace without project files](https://github.com/BetterThanTomorrow/calva/issues/2069)
 
 ## [2.0.332] - 2023-02-15
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Changes to Calva.
 
 ## [Unreleased]
 
-- [Rollback clojure-lsp auto-start behaviour default](https://github.com/BetterThanTomorrow/calva/pull/2077)
+- Fix: [Default clojure-lsp startup behaviour is changed since v2.0.327](https://github.com/BetterThanTomorrow/calva/issues/2084)
 - Fix: [clojure-lsp does not start in a workspace without project files](https://github.com/BetterThanTomorrow/calva/issues/2069)
 
 ## [2.0.332] - 2023-02-15
--- a/docs/site/clojure-lsp.md
+++ b/docs/site/clojure-lsp.md
@@ -16,15 +16,20 @@ By default you won't need to install/setup anything as Calva handles that for yo
 
 Calva is able to automatically start the clojure-lsp server for you and can be configured to start the server under various different conditions. These behaviours can be configured through the `calva.enableClojureLspOnStart` setting, which takes the following options:
 
++ "always-use-first-workspace-root"
 + "when-workspace-opened-use-workspace-root"
 + "when-file-opened-use-furthest-project"
 + "never"
 
-#### "when-workspace-opened-use-workspace-root" [default]
+#### "always-use-first-workspace-root" [default]
 
-When set to `"when-workspace-opened-use-workspace-root"` Calva will start the clojure-lsp in the root of all opened vscode workspaces. All Clojure files in a workspace will be serviced by the clojure-lsp server running in that workspace. This behavior requires that you are opening workspaces with a valid Clojure project in the root (the directory must contain a `deps.edn`, `project.clj` or `shadow-cljs.edn` file).
+When set to `"always-use-first-workspace-root"` Calva will attempt to start the clojure-lsp in the root of the first workspace folder if it is a valid clojure project. If it is not a valid clojure project it will fall back to starting the [fallback server](#fallback-server).
 
 This is the default auto-start behaviour.
+
+#### "when-workspace-opened-use-workspace-root"
+
+When set to `"when-workspace-opened-use-workspace-root"` Calva will start the clojure-lsp in the root of all opened vscode workspaces. All Clojure files in a workspace will be serviced by the clojure-lsp server running in that workspace. This behavior requires that you are opening workspaces with a valid Clojure project in the root (the directory must contain a `deps.edn`, `project.clj` or `shadow-cljs.edn` file).
 
 #### "when-file-opened-use-furthest-project"
 
--- a/package.json
+++ b/package.json
@@ -173,14 +173,16 @@
             "enum": [
               "when-workspace-opened-use-workspace-root",
               "when-file-opened-use-furthest-project",
+              "always-use-first-workspace-root",
               "never"
             ],
             "enumDescriptions": [
               "Start clojure-lsp automatically when a workspace containing a Clojure project is opened. The server will be started with its project root set to the workspace root. Best option for many single-Clojure-project workspaces and Polylith-like projects. It is recommended to configure your project so that you can use this option and have only one clojure-lsp server running.",
               "Start clojure-lsp automatically when a Clojure file is opened. The server will be started with its project root set to the project found furthest from the opened file. Use this option if your monorepo has several Clojure projects, but lacks a governing project file at the workspace root.",
+              "Start clojure-lsp automatically in the first workspace root. Use this option to avoid starting multiple lsp servers when multiple independent clojure projects are added as workspace folders",
               "Don't start clojure-lsp automatically. You can always start clojure-lsp at will via the command palette or from the statusbar. Good option in multi-project workspaces where you want to start a clojure-lsp for a particular project. Also useful if you are running on a machine with limited resources, like a Raspberry Pi."
             ],
-            "default": "when-workspace-opened-use-workspace-root"
+            "default": "always-use-first-workspace-root"
           },
           "calva.prettyPrintingOptions": {
             "type": "object",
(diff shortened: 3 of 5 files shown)


PEZ:
Before pulling #2075 and #2076, what's the status of this one? Is it updated with those PRs? I'd like for this one to be ready for merge once we merge the other two, because otherwise we'll be in a state where we can't release `dev`.

julienvincent:
@PEZ Both #2075 and #2076 can be merged and released independently of this PR (and independently of each other). This PR shouldn't ever need to block `dev` from being released.

Regardless I'll pull #2076 into this PR in the meantime.

PEZ:
This is the most important of the changes, since it is dealing with a regression. I think we should have had this one out first and then worked on the improvements. It's blocked because I don't want to release changes to the clojure-lsp auto-start behaviour without restoring the default Calva used to have. It get's too weird for the users.

julienvincent:
> I think we should have had this one out first and then worked on the improvements

We can release this independently of #2076 if we want to, fyi.

> This is the most important of the changes, since it is dealing with a regression

To be honest I don't agree that this PR is resolving a regression. I feel quite strongly that this PR is actually a bad idea and is taking a step backwards. It's bothering me so I would like to motivate:

I believe the work in #2075 and #2076 are the real fixes for regressions we have seen as a result of #2020. The intent behind #2020 was to try keep the existing Calva behaviour as similar as possible for existing users while adding support (by default) for users with multiple projects.

I failed to achieve this in the PR as the behaviour for existing users was affected more than intended. I do think however that once #2075 and #2076 are merged the behaviour of Calva should be nearly identical to what it was before (when used in the same way) but should additionally work as expected when opening multiple projects, which is what the original intention was. I don't think existing users should notice much of a difference, multi-workspace setups should work, it feels like a win-win to me.

By merging this PR we are effectively dropping support again (by default) for multi-project workspaces.

Part of the motivation for this PR was to avoid spawning multiple processes when a user has an excessive number of projects open. I don't think however that this is the right way to address this concern.

- If the user has opened multiple independent clojure projects I think that we should err on the side of trying to support what the user is doing instead of preventing them.
- I think it is quite unlikely that users would be opening an excessively high number of independent projects. It's probably a lot more likely they have 3 or 4 open, which should be fine.
- We don't have the same prejudice against multiple opened vscode windows. If a user opened 20+ projects in independent vscode windows that would anyway spawn 20+ lsp servers - producing the same effect.

I think a better way to address this particular concern is to have a warning in Calva when spawning more than X processes, allowing the user to confirm they want to continue.

---

I think the fact that we have been able to address the issues that have been reported by users so far, without needing to change the default again, is quite promising and motivates that there is a path forward without us needing to land this change.

PEZ:
I appreciate that you try to see beyond the immediate symptoms and that you come up with better solutions than an immediate fix. However, this is the Calva project, it works in a special way. We dare make quite big changes without paralysis out of fear of some breakage, because we are ready to roll back, or quickly fix regressions. It is so far unprecedented that our users have waited over a two days for a fix of a regression of this importance. I just don't want that to happen. Among all the users there are tons of them who have been having issues with the change, that we have not heard from.

With Calva we fix/revert as fast as we can, even when we know of better solutions. That buys us time to get the better solutions ready. I am disagreeing with the order you are doing things here, not the things you are doing.

Let's do things the Calva project's way. It has served our users well so far.

We also disagree of what is a good default. Until we reach agreement on that we should not change defaults.

I now tried in one of the workspaces I have. It's for a client building a rather big application. Even with #2075 Calva now starts over 30 clojure-lsp servers, and it takes almost that many minutes for all servers to have started. I think the workspace is a bit crazy. Which is the reason I haven't seen the issue until i now got reminded to test it. I usually work with that project with one folder at a time (just opening the folders I am working with in separate windows, without any workspace file).

The #2075 default behaviour is great for some projects as an opt in, but the way Calva used to behave serves most project great. To get clojure-lsp to behave better in many monorepos is most often to add config so that one clojure-lsp server can serve it all, Polylith style. That will open up for one REPL serving all projects in the repo as well, Polylith style.

It could happen that I am wrong and you are right. I am often wrong. All I am asking is that we first reach agreement, then change. Without agreement, we don't change.

PEZ:
Looks like so.

image

Good news is that all of them go away when I close the workspace.

PEZ:
Some other good news is that this PR fixes it.

So, if we were to merge the PRs now. How do they depend on each other? Is this one building on the others? I noticed that I got a tmp server with this one for a workspace I tested with. But I don't see that change in the changelog.

julienvincent:
> We dare make quite big changes without paralysis out of fear of some breakage, because we are ready to roll back, or quickly fix regressions. It is so far unprecedented that our users have waited over a two days for a fix of a regression of this importance.

> I am disagreeing with the order you are doing things here, not the things you are doing.

I appreciate this approach to making changes and agree with the sentiment of needing to rollback undesirable changes as soon as possible.

I think I did not clarify well enough my intent behind these three PR's. I opened all three on the same day and tried to keep them independent from each other so that the order in which they are merged and released does not matter.

This particular PR, in it's current state, basis off of #2076 but that is not a requirement - it was just presented as such as I think they go well together. The top commit can be very easily shifted onto `dev` and released in isolation. I guess I wasn't clear enough that I was offering both as an option.

I am happy to defer to you with regard to the order in which any of these PR's get merged and released.

julienvincent:
> I now tried in one of the workspaces I have. It's for a client building a rather big application. Even with #2075 Calva now starts over 30 clojure-lsp servers, and it takes almost that many minutes for all servers to have started. I think the workspace is a bit crazy. Which is the reason I haven't seen the issue until i now got reminded to test it. I usually work with that project with one folder at a time (just opening the folders I am working with in separate windows, without any workspace file).

This workspace does sound pretty crazy haha. Can you tell me more about this project? Is it a mono-repo? Why are the folders not part of some root `deps.edn`? Also, what is the reason for needing to open all 30+ projects at once (especially if they are independent, as it seems)?

> ... and it takes almost that many minutes for all servers to have started.

This particular issue should actually be resolved in #2076 - prior to that PR lsp servers would be started sequentially, but they are now started in parallel.

PEZ:
I don't know much about the rationale behind the workspace. It's a monorepo with a lot of sub projects/libraries. And then this workspace makes all sub projects workspace roots. I only work in two of these sub projects and sometimes in the main project. Haven't been using the workspace, just got the idea that it might be good to see how it behaves in this multi-clojure-lsp environment.

As for merging. I want to merge all PRs. Just not sure how I do it, but if I merge one, then you can update the next and then the next, making them ready to be merged? This one has zero checklist items ticked, so maybe it isn't ready?

julienvincent:
> As for merging. I want to merge all PRs. Just not sure how I do it, but if I merge one, then you can update the next and then the next, making them ready to be merged? This one has zero checklist items ticked, so maybe it isn't ready?

Ok lets get these merged today then. I would say we start by merging #2076 and this one, and then I'll resolve any conflicts in #2075 and we can merge that too.

I have updated the site documentation here to include the new default setting enum.

PEZ:
We're not quite ready to merge this one yet. 😄 

> * [ ] Updated the `[Unreleased]` entry in `CHANGELOG.md`, linking the issue(s) that the PR is addressing.
> * [ ] Referenced the issue I am fixing/addressing _in a commit message for the pull request_.
>     * [ ]  If I am fixing the issue, I have used [GitHub's fixes/closes syntax](https://help.github.com/en/articles/closing-issues-using-keywords)

What this means is partly that the PR lacks a problem statement. I should have posted an issue earlier, but now fixed that:

* https://github.com/BetterThanTomorrow/calva/issues/2084

This is what should be linked in the Changelog and mentioned with closing syntax in a commit.

There also seems to be a lack of testing:

> * [ ] Figured if **anything** about the fix warrants tests on Mac/Linux/Windows/Remote/Whatever, and either tested it there if so, or mentioned it in the PR.
> * [ ] Figured if the change might have some side effects and tested those as well.

PEZ:
As for the Windows/cross-platform question. It can be hard to test. Most often a PR does not risk being sensitive to platform differences, but it happens, and that checklist item is there to have us pause and consider it. And if we see a possible problem that we think should be tested we need to ask each other for help about it. (This could be either testing or someone checking the relevant part of the code for a second opinion.)

julienvincent:
I don't think anything in this PR specifically requires testing on windows/cross-platform.

Perhaps the logic from #2076 does. I see if I can spin up that test windows VM again.

PEZ:
> I don't think anything in this PR specifically requires testing on windows/cross-platform.

You know where to update that status. 😄 

> Perhaps the logic from #2076 does. I see if I can spin up that test windows VM again.

I had some success with provisioning a Windows machine on Azure. But I'll have to put in some more work there before we can use it for these purposes.

Maybe see if we can get help from Calva users in the slack channel?

julienvincent:
I'll ask in the slack channel. If possible can you give this PR a try as well and confirm the behaviour is as you expect?

PEZ:
> can you give this PR a try as well and confirm the behaviour is as you expect?

Sorry, missed this question. I have tested this and it worked as I expected. Including only one server started in that crazy workspace. 😄

PR #2099, opened 2023-03-07
Windows CI testing

PEZ:
This PR adds build and test jobs running with Windows executors on Circle CI.

image

It was a lot of work figuring out how to do it and there is still work needed to be done to make it work fully. What works:

- ✅ Calva is checked out and built on Windows
- ✅ cljs unit tests run and pass
- ✅ ts unit tests run and pass
- 🛑 The intergration test environment starts, and the job is marked as succeeding, but I don't think the tests are actually run
- 🛑 Same with the two e2e tests.

I will not continue with this and don't think we should aim for merging. Reasons:

- All steps also take very, very long time, compared to their Linux counterparts. It's 20+ minutes to wait for feedback.
- Windows runs count towards our Free Plan 30K credits and they are depleted now, because it took me so many test runs to even get to where it is now.
- The feedback loop is slow and I think it will take a very long time to figure out why the integration and e2e tests do not actually run.

We will need to figure out other ways to ensure Calva works on Windows.

## My Calva PR Checklist

I have:

- ~[ ] Updated the `[Unreleased]` entry in `CHANGELOG.md`, linking the issue(s) that the PR is addressing.~
- [x] Tests


PEZ:
Closing it directly. This PR is for documenting that it even happened. 😄

bpringe:
It was worth a try. At least we know now what method not to use. 😄

PR #2102, opened 2023-03-09
Promptless jack-in and connect

### LINKED ISSUE #1908: AutoConnect setting?
I'm trying to create a REPL connect sequence where I don't have to hit enter on the `localhost:port` prompt.  I see in the `connect` function that if `isAutoConnect` is true is set then it'll automatically connect, but I'm not sure where that is being set.

I tried setting `"calva.autoConnect": true` in preferences but that didn't seem to do it.

### LINKED ISSUE #1984: Expose calva.connect as VSCode API
With Joyride, I could create my own quick-tasks that connect to project-specific REPLs e.g. a production REPL via an SSH tunnel that is always at the same port, or picking a shadow-cljs build all the time, without needing to discover a CLJS configuration. (I always pick `:main`).

Imagined API:

```
await (vscode/commands.executeCommand "calva.connect" "localhost:6666")
;; or perhaps to auto-discover? But I would probably expose the auto-discovery in its own API
await (vscode/commands.executeCommand "calva.connect" #js {:type "shadow-cljs" :build ":main"}})
(vscode/commands.executeCommand "calva.connect" #js {:type "shadow-cljs" :build ":main"}})
```

Of course it could be a part of Calva's Extension API proper, perhaps at the `repl` module?

```
(calva/repl.connect "localhost:6666")
```

I'm not sure I understand how to connect to a CLJS repl so please ignore my API suggestions :)

### LINKED ISSUE #2094: Option for auto selecting a project
### Motivation 💪
Starting a REPL with a single command, without any configuration to learn, is an excellent value for someone jumping on an existing project, exploring an example project for a concept or library, or getting started with Clojure.

### Background 👍
The `replConnectSequences` option allows us to define everything needed to start a REPL on a specific project so that the developer can just run the Jack-In command in Calva and get a running REPL immediately.

Additionally, Calva allows auto-selecting the custom sequences defined within the `replConnectSequences` with the `calva.autoSelectReplConnectProjectType` option. However, if you work in a mono repo, there is another step before project type selection: `Project Selection`.

### Suggestion 💡
It would be great to auto-select a project for the Jack-In command. It could follow the same naming scheme as the project type selection: `calva.autoSelectReplConnectProjectSelection`. Then, I could set it to the `Workspace Root` (or any other project within the mono repo), and just running one single command, the Jack-In command, will be enough to get a REPL!

```jsonc
{
    "calva.autoOpenJackInTerminal": true,
    "calva.autoSelectReplConnectProjectType": "My Favorite Seq",
    "calva.autoSelectReplConnectProjectSelection": "Workspace Root", // <-- Suggested option
        {
            "name": "My Favorite Seq",
            ...   
        }
    ]
}
```

> PS: Having this option would let me remove the last step [here](https://github.com/furkan3ayraktar/clojure-polylith-realworld-example-app#start-a-repl-in-vscode--calva) and allow people quickly get a Polylith REPL.

### LINKED ISSUE #2101: Select connect host:port automatically from port file, without prompting
Calva insists on asking for `host:port` with a prefilled prompt. While it is just an `enter` key press away, it is still quite unecessary in the vast majority of cases, I think.

Let's make it the default to not prompt for this and add a config for changing this default.

PEZ:
## What has changed?

This is a collection of changes making fully promptless jack-in and connect possible.

- Added a deprecation message to `calva.autoSelectReplConnectProjectType`
- Added a REPL Connect Sequence config `autoSelectForJackIn`
- Added a REPL Connect Sequence config `autoSelectForConnect`
- Added a config `calva.autoSelectNReplPortFromPortFile` and made it default true
- Made the command `calva.connect` take an options object with `host`, `port`, `connectSequence`, and `disableAutoSelect` arguments
- Made the command `calva.jackIn` take an options object with `connectSequence`, and `disableAutoSelect` arguments
- Added a config `calva.autoConnectRepl` that when true will make Calva automatically connect to a running REPL when a project is opened.

When a connect sequence has both `projectRootPath` and `autoSelect` configured, both the project root menu and the project type menues will be suppressed. Unless the connect/jackIn command provides the optuon `disableAutoSelect`

* Fixes #2094
* Fixes #2101
* Fixes #1984
* Fixes #1908 

## My Calva PR Checklist

I have:

- [x] Tests


PEZ:
@Cyrik The missing piece here is the auto-connect feature you wanted:

* #1908 

I think it makes sense to add it, just is a bit involved so I deemed it ”do later”. The tricky part is to figure out if auto-connect should happen. I think it is a matter of picking up the `nReplPortFile` from all connect sequences (built in and custom) and then look for those. If anyone of them exists, fire the `calva.connect` command.

Cyrik:
Yeah, I think looking for that file was the way I wanted to go as well.
But this is already pretty close to autoconnect anyway, so I'd say that the ticket actually fell in urgency, even if it's now easier to implement :)

PEZ:
Ah, wth, I added the auto-connect option now, @Cyrik 😄 When I described above how it should be done, I realized it wasn't much anyway. Please give it a spin!

docs/site/connect-sequences.md
```diff
@@ -20,6 +20,8 @@ A connect sequence configures the following:
 
 * `name`: (required) This will show up in the Jack-in quick-pick menu when you start Jack-in (see above).
 * `projectType`: (required) This is either "Leiningen”, ”deps.edn”, ”shadow-cljs”, ”lein-shadow”, "Gradle", or ”generic".
+* `autoSelect`: A boolean.If true, this sequence will be automatically selected, suppressing the Project Type menu when connecting to a REPL or Jacking in. Use together with `projectRootPath` to also suppress the Project Root menu. Add usage of `menuSelections` to go for a prompt-less REPL connection. If you have more than one sequence with `autoSelect` set to true, the first one will be used.
```
bpringe:
```suggestion
* `autoSelect`: A boolean. If true, this sequence will be automatically selected, suppressing the Project Type menu when connecting to a REPL or Jacking in. Use together with `projectRootPath` to also suppress the Project Root menu. Add usage of `menuSelections` to go for a prompt-less REPL connection. If you have more than one sequence with `autoSelect` set to true, the first one will be used.
```

docs/site/connect-sequences.md
```diff
@@ -20,6 +20,8 @@ A connect sequence configures the following:
...
+* `projectRootPath`: An array of path segments leading to the root of the project for which this connect sequence should go with. Use together with `autoSelect` to suppress the Project Root menu. The path can be absolute or relative to the workspace root. If there are several Workspace Folders, the workspace root is the path of the first folder. So relative paths will only work for this first folder.
```
bpringe:
```suggestion
* `projectRootPath`: An array of path segments leading to the root of the project to which this connect sequence corresponds. Use together with `autoSelect` to suppress the Project Root menu. The path can be absolute or relative to the workspace root. If there are several Workspace Folders, the workspace root is the path of the first folder, so relative paths will only work for this first folder.
```

docs/site/connect-sequences.md
```diff
@@ -43,12 +45,12 @@ A connect sequence configures the following:
 
 The [Calva built-in sequences](https://github.com/BetterThanTomorrow/calva/blob/published/src/nrepl/connectSequence.ts) also use this format, check them out to get a clearer picture of how these settings work.
 
-!!! Note "Apropos the **ClojureScript nREPL Server** built-in sequence"
-    Because of ancient decisions in the design of Calva session management, the current implementation of the ClojureScript nREPL Server connect sequence is a workaround. Calva will still indicate that it has a Clojure session available, which is nonsense in a pure ClojureScript nREPL environment. This session is also a ClojureScript session.
+!!! Note "Path segments"
+    `projectRootPath` and `nReplPortFile` both take an array of path segments. This is to make the paths wor cross-platform. If you can't be bothered splitting up the path in segments, put the whole path in the first segment, though please note that if you use Windows path separators, these will not work for users with Linux or macOS.
```
bpringe:
```suggestion
    `projectRootPath` and `nReplPortFile` both take an array of path segments. This is to make the paths work cross-platform. If you can't be bothered splitting up the path in segments, put the whole path in the first segment, though please note that if you use Windows path separators, these will not work for users with Linux or macOS.
```

docs/site/connect-sequences.md
```diff
@@ -43,12 +45,12 @@ A connect sequence configures the following:
...
 
 ## Example Sequences
 
-Wether you hust want to speed up your workflow, or encode some workflow/mechanics into it, there's often that you can create a custyom sequence that helps.
+Wether you just want to speed up your workflow, or encode some workflow/mechanics into it, there's often that you can create a custom sequence that helps.
```
bpringe:
```suggestion
Whether you just want to speed up your workflow or encode some workflow/mechanics into it, it's often the case that you can create a custom sequence that helps.
```

docs/site/connect-sequences.md
```diff
@@ -71,6 +73,30 @@ Minimize the amount of selecting from the Jack-in/Connect menu when working with
 
 See [shadow-cljs + Clojure with Calva: The basics](https://blog.agical.se/en/posts/shadow-cljs-clojure-cljurescript-calva-nrepl-basics/) for how Calva and nREPL work with ClojureScript.
 
+
+### Polylith
+
+This is the connect sequence used in the [Polylith Real World App](https://github.com/furkan3ayraktar/clojure-polylith-realworld-example-app). It let's you jack-in to the project, and starts the Real World App without any prompts.
```
bpringe:
```suggestion
This is the connect sequence used in the [Polylith Real World App](https://github.com/furkan3ayraktar/clojure-polylith-realworld-example-app). It lets you jack-in to the project, and starts the Real World App without any prompts.
```

docs/site/connect-sequences.md
```diff
@@ -71,6 +73,30 @@ Minimize the amount of selecting from the Jack-in/Connect menu when working with
...
+            "cljsType": "none",
+            "menuSelections": {
+                "cljAliases": ["dev", "test"]
+            }
+        }
+    ],
+    "calva.autoConnectRepl": true,
+```
+
+The `calva.autoConnectRepl` config makes Calva, at project open, look for the nRepl port file and automatically connect the repl if the file exists. Meaning you can leave the app running when you close the project in VS Code, and Calva will reconnect when you open the project again. (Or maybe you just need to reload the VS Code window, and not loose the REPL state.)
```
bpringe:
```suggestion
The `calva.autoConnectRepl`, when set to `true`, makes Calva, at project open, look for the nRepl port file and automatically connect the repl if the file exists. Therefore, you can leave the app running when you close the project in VS Code, and Calva will reconnect when you open the project again. (Alternatively, maybe you just need to reload the VS Code window and not lose the REPL state.)
```

docs/site/connect.md
```diff
@@ -66,26 +66,27 @@ All this said, I still recommend you challenge the conclusion that you can't use
...
 
-### Starting the REPL from application code?
+### Customizing Connect
 
-If your project is setup so that the REPL server is started by the application code, you will need to get the cider-nrepl middleware in place. See the cider-nrepl docs about [embedding nREPL in your application](https://docs.cider.mx/cider-nrepl/usage.html#via-embedding-nrepl-in-your-application).
+If there is an nRepl port file, Calva will use it and not prompt for `host:port` when connecting. You can make Calva prompt for this by setting the boolean config `calva.autoSelectNReplPortFromPortFile` to `false`.
+
+With the setting `calva.autoConnectRepl` you can make Calva automatically connect the REPL if there is an nRepl port file present when the project is opened.
+
+With this and the below mentioned auto-select options you can make connect a promptless experience. See: [Connect Sequences](connect-sequences.md).
```
bpringe:
```suggestion
With this and the below mentioned auto-select options you can make connect a prompt-less experience. See: [Connect Sequences](connect-sequences.md).
```

package.json
```diff
@@ -552,6 +563,17 @@
...
+                  "type": "boolean",
+                  "markdownDescription": "If true, this sequence will be automatically selected, suppressing the Project Type menu when connecting to a REPL or Jacking in. Use together with `projectRootPath` to also suppress the Project Root menu. Add usage of `menuSelections` to go for a prompt-less REPL connection. If you have more than one sequence with `autoSelect` set to true, the first one will be used."
+                },
+                "projectRootPath": {
+                  "type": "array",
+                  "markdownDescription": "An array of path segments leading to the root of the project for which this connect sequence should go with. Use together with `autoSelect` to suppress the Project Root menu. The path can be absolute or relative to the workspace root. If there are several Workspace Folders, the workspace root is the path of the first folder. So relative paths will only work for this first folder. If you can't be bothered splitting up the path in segments, put the whole path in the first segment, though please note that if you use Windows path separators, these will not work for users with Linux or macOS.",
```
bpringe:
```suggestion
                  "markdownDescription": "An array of path segments leading to the root of the project to which this connect sequence corresponds. Use together with `autoSelect` to suppress the Project Root menu. The path can be absolute or relative to the workspace root. If there are several Workspace Folders, the workspace root is the path of the first folder, so relative paths will only work for this first folder. If you can't be bothered splitting up the path in segments, put the whole path in the first segment, though please note that if you use Windows path separators, these will not work for users with Linux or macOS.",
```

package.json
```diff
@@ -552,6 +563,17 @@
...
+                  "items": {
+                    "type": "string"
+                  }
+                },
                 "nReplPortFile": {
                   "type": "array",
                   "description": "An array of path segments with the project root-relative path to the nREPL port file for this connect sequence. E.g. For shadow-cljs this would be [\".shadow-cljs\", \"nrepl.port\"]",
```
bpringe:
```suggestion
                  "description": "An array of path segments for the relative path from the project root to the nREPL port file for this connect sequence. E.g. For shadow-cljs this would be [\".shadow-cljs\", \"nrepl.port\"]",
```

src/connector.ts
```diff
@@ -612,12 +613,14 @@ export async function connect(
         await vscode.workspace.fs.stat(portFile);
         const bytes = await vscode.workspace.fs.readFile(portFile);
         port = new TextDecoder('utf-8').decode(bytes);
+        outputWindow.appendLine(`; Read port file: ${portFile}`);
```
bpringe: To be consistent with the "; Using host:post ..." statement below, maybe this could be placed before the file is read and say "; Reading port file: ..."?

src/connector.ts
```diff
@@ -662,14 +669,34 @@ async function standaloneConnect(connectSequence: ReplConnectSequence) {
...
       .logEvent('REPL', 'StandaloneConnect', `${connectSequence.name} + ${cljsTypeName}`)
       .send();
-    await connect(connectSequence, false).catch(() => {
+    return connect(
+      connectSequence,
+      getConfig().autoSelectNReplPortFromPortFile,
+      hostname,
+      port
+    ).catch(() => {
       // do nothing
```
bpringe: I know this was like this before this PR, but don't we want to do something like log or alert the user if connecting fails?

src/extension.ts
```diff
@@ -481,6 +481,10 @@ async function activate(context: vscode.ExtensionContext) {
     );
   }
 
+  if (await connector.shouldAutoConnect()) {
```
bpringe, line 484: There might be an issue here. `shouldAutoConnect` returns a promise that resolves to `nReplPortFileExists()` if the `autoConnectRepl` config is true. `nReplPortFileExists` returns a promise that resolves to an array of true and false values, so this `if` can evaluate to something like `if ([false, false, false]) ...` if no port file exists, which would be a true statement. Then the connect command would be run even though there's no port file.

image

bpringe: I'm also not sure if the call to `nReplPortFileExists` inside `shouldAutoConnect` should be awaited in order for the promise returned from `shouldAutoConnect` to resolve to an actual array and not to another promise that resolves to an array.

src/nrepl/jack-in.ts
```diff
@@ -262,9 +257,8 @@ export async function jackIn(connectSequence: ReplConnectSequence, cb?: () => un
 
   let projectConnectSequence: ReplConnectSequence = connectSequence;
   if (!projectConnectSequence) {
-    try {
-      projectConnectSequence = await getProjectConnectSequence();
-    } catch (e) {
+    projectConnectSequence = await getProjectConnectSequence();
```
bpringe: `getProjectConnectSequence` can possibly throw an exception. Do we not still want to catch that here like we were?

bpringe (commented):
Looks great overall! Found some things to improve doc-wise and commented about some concerns.

src/connector.ts
PEZ: I agree. Will fix!

src/connector.ts
PEZ: I'll have a look at the call sites. Might be that we should return the promise without this catching.

src/extension.ts
PEZ: I'll look again at this.

src/nrepl/jack-in.ts
PEZ: Right now this will show an error alert to the user with the error message being thrown, instead of continuing with jack-in. I'll see if it is that we shuold show our own message instead and return early.

src/extension.ts
PEZ: I think that this should be taken care of now.

src/nrepl/jack-in.ts
PEZ: I now remember why I choose to do it like this. For some reason the error message is never shown when I try to do vscode.window.errorMessage. I now choose to disregard that and output the error to the output window as well as (failing to) show the error message. It works from the connect call site, so there's that.

src/connector.ts
PEZ: Very odd, I think the reason we did nothing here was that there is something funny with the call stack. I now choose to remove the `catch` and somewhat handle the error, cleaning up state and such, at the call site.

src/extension.ts
bpringe: I don't see a change to this code. It still looks potentially problematic as stated above. Maybe I'm missing something?

src/nrepl/jack-in.ts
```diff
@@ -265,6 +260,12 @@ export async function jackIn(connectSequence: ReplConnectSequence, cb?: () => un
     try {
       projectConnectSequence = await getProjectConnectSequence();
     } catch (e) {
+      outputWindow.appendLine(`; ${e}\n; Aborting jack-in.`);
+      // TODO: Figure out why this is not shown to the user.
+      void vscode.window.showErrorMessage(e, 'OK');
```
bpringe, line 275: I wonder if awaiting this would cause it to be shown to the user.

src/extension.ts
PEZ: It doesn't resolve to an array any longer.

src/nrepl/jack-in.ts
PEZ: It doesn't. I've tried that and some 100 other things. =)

src/extension.ts
bpringe: Oh, gotcha!

PR #2118, opened 2023-03-16
Use tree-kill again for the jack-in process

### LINKED ISSUE #2116: Stopping Jacked-in REPL process doesn't kill Unix Java process
Enviroment:
```
VSCode: latest
WSL2: Ubuntu 18.04
Calva: v2.0.338
```

Checked on a minimal project:
Run Jack-in process via `deps.end` project-type.
`ps -a`  shows java process.
Then `ctrl + c` in Calva terminal followed with `ps -a` and java process is still there.

PEZ:
We shifted to just calling `.kill()` on the process in 2.0.337, which fixed an issue with us not being able to detect if we had killed the process or not. But it also caused us to fail to kill any REPL processes that have child processes. So now we are doing both tree-kill and calling the `.kill()` method.

* Fixes #2116

## My Calva PR Checklist

I have:

- [x] Tests

### DIFF
--- a/src/nrepl/jack-in-terminal.ts
+++ b/src/nrepl/jack-in-terminal.ts
@@ -116,8 +116,16 @@ export class JackInTerminal implements vscode.Pseudoterminal {
       this.writeEmitter.fire('🛑 Stopping/killing the Jacked-in REPL process... 🛑\r\n');
       console.log('Jack-in terminal killProcess(): Closing any ongoing stdin event');
       this.process.stdin.end(() => {
-        console.log('Jack-in terminal killProcess(): Closing any ongoing stdin event');
-        this.process.kill();
+        // On some machines we need to use tree-kill to kill the process, so we do it always
+        // https://github.com/BetterThanTomorrow/calva/issues/2116
+        console.log('Jack-in terminal killProcess(): Killing process using tree-kill');
+        kill(this.process.pid, 'SIGTERM', (err) => {
+          if (err) {
+            console.log('Jack-in terminal killProcess(): Error killing process', err);
+          }
+          // The test for this.process.killed above needs this to have happened too
+          this.process.kill();
+        });
       });
     } else if (this.process && this.process.killed) {
       this.writeEmitter.fire(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ Changes to Calva.
 
 ## [Unreleased]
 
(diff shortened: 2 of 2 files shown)


src/nrepl/jack-in-terminal.ts
```diff
@@ -116,8 +116,16 @@ export class JackInTerminal implements vscode.Pseudoterminal {
...
       console.log('Jack-in terminal killProcess(): Closing any ongoing stdin event');
       this.process.stdin.end(() => {
-        console.log('Jack-in terminal killProcess(): Closing any ongoing stdin event');
-        this.process.kill();
+        // On some machines we need to use tree-kill to kill the process, so we do it always
+        // https://github.com/BetterThanTomorrow/calva/issues/2116
+        console.log('Jack-in terminal killProcess(): Killing process using tree-kill');
+        kill(this.process.pid, 'SIGTERM', (err) => {
+          if (err) {
+            console.log('Jack-in terminal killProcess(): Error killing process', err);
```
bpringe, line 124: Should we only show this message if `this.process.kill()` (below) fails to kill the process?

I see in the doc for that method says it returns false for failure. We could check that.

bpringe: I'm assuming the below call is sort of a second / further try to kill the process if the first time fails, but maybe I misunderstand.

PEZ: It's complicated. 😄 What happpened here was that I had forgotten why we needed `tree-kill`. I noticed that using `process.kill()` fixed the bug where we clould not detect that the process was killed (`process.killed` returned `false`). So I changed to using `process.kill()` instead of `tree-kill`. Which led to the 337 bug where we failed to kill the process if it had spawned sub processes.

We have every reason to believe that `tree-kill` actually kills the process. The call to `process.kill()` is there only to get the `process.killed` check to work.

Did that make sense? I can hardly unpack it myself...

Anyway, all that said, as I almost expect the `process.kill()` call to "fail” (since we have killed the process with `tree-kill`), I think we might create noice if we log that.

The real fix is probably to figure out how to check if the process is dead without using `process.killed`.

bpringe: Ahh.

> We have every reason to believe that tree-kill actually kills the process. The call to process.kill() is there only to get the process.killed check to work.

This is where I'm confused, since the check of process.killed happens _before_ the call to process.kill(), so I don't see how the latter enables the former to work.

PR #2167, opened 2023-04-21
Only log cljfmt+clojure-lsp error if clojure-lsp has started

### LINKED ISSUE #2078: cljfmt isn't found unless project is started
`Fetching formatting settings from clojure-lsp failed. Check that you are running a version of clojure-lsp that provides "cljfmt-raw" in serverInfo.`

If I launch vscode I get the above message, and cljfmt wont format correctly.  If I then start my project (lein repl), which happens to include lein-cljfmt, then cljfmt starts formatting correctly.  More interestingly, if I stop the project but don't close vscode, cljfmt continues to format correctly.  

This is a new behavior as of a few days (I was off 2/4 - 2/9)

Im not sure where clojure-lsp is looking for cljfmt, is there somewhere else besides project.clj / plugins I need to be updating?

I think cljfmt had a pretty big changed merged in recently, but can't confirm the timeline.

PEZ:
## What has changed?

- Log error to console instead of alerting about it
- Only log error if the server has started

The trade-off here is that the user might get default formatting without having been noticed. But it should be rare, and not the worlds biggest deal.

* Fixes #2078

## My Calva PR Checklist

I have:

- [ ] Tests

### DIFF
--- a/src/formatter-config.ts
+++ b/src/formatter-config.ts
@@ -51,15 +51,12 @@ export async function getConfig(
   if (configPath === LSP_CONFIG_KEY && document) {
     const clientProvider = lsp.getClientProvider();
     const client = clientProvider.getClientForDocumentUri(document.uri);
-    if (client) {
+    if (client && client.isRunning) {
       lspFormatConfig = await lsp.api.getCljFmtConfig(client);
     }
-  }
-  if (configPath === LSP_CONFIG_KEY && !lspFormatConfig) {
-    void vscode.window.showErrorMessage(
-      'Fetching formatting settings from clojure-lsp failed. Check that you are running a version of clojure-lsp that provides "cljfmt-raw" in serverInfo.',
-      'Roger that'
-    );
+    if (client.isRunning && !lspFormatConfig) {
+      console.error('Fetching formatting settings from clojure-lsp failed. Check that you are running a version of clojure-lsp that provides "cljfmt-raw" in serverInfo.');
+    }
   }
   const cljfmtContent: string | undefined =
     configPath === LSP_CONFIG_KEY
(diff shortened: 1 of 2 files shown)


src/formatter-config.ts
```diff
@@ -51,15 +51,14 @@ export async function getConfig(
...
+    if (client && client.isRunning) {
       lspFormatConfig = await lsp.api.getCljFmtConfig(client);
     }
-  }
-  if (configPath === LSP_CONFIG_KEY && !lspFormatConfig) {
-    void vscode.window.showErrorMessage(
-      'Fetching formatting settings from clojure-lsp failed. Check that you are running a version of clojure-lsp that provides "cljfmt-raw" in serverInfo.',
-      'Roger that'
-    );
+    if (client.isRunning && !lspFormatConfig) {
```
bpringe: Would this error get logged even if `configPath !== LSP_CONFIG_KEY`? Does that mean this error will get logged always by default, until someone sets `configPath === LSP_CONFIG_KEY`?

bpringe: On another note, I think we need a check here for the client existing like you added in the `if` condition above -- `client && client.isRunning`. See [this comment](https://github.com/BetterThanTomorrow/calva/issues/2078#issuecomment-1518326258). Better yet though, you can nest the check for `lspFormatConfig` in the first `if` block. Something like:

```javascript
if (client && client.isRunning && configPath === LSP_CONFIG_KEY) { // see question above on that last check
  lspFormatConfig = await lsp.api.getCljFmtConfig(client);
  if (!lspFormatConfig) {
    console.error(...
  }
```

PEZ: Not sure I understand the first concern. It is all protected with a check for `configPath === LSP_CONFIG_KEY`, right?

I fixed the second concern in the latest commit. Please have a look. Maybe your first concern still needs to be addressed.

bpringe: Oh, I didn't see that check before. Never mind the first concern.

PR #2285, opened 2023-08-15
Use array instead of object for indent rules

juiceghost:
## What has changed?

- Use array instead of object for indent rules. The earlier method of spreading objects and then combining them overwrote existing keys in the provided configuration.

* Fixes #2284 

## My Calva PR Checklist

I have:

- [X] Read [How to Contribute](https://github.com/BetterThanTomorrow/calva/wiki/How-to-Contribute#before-sending-pull-requests).
- [X] Directed this pull request at the `dev` branch. (Or have specific reasons to target some other branch.)
- [X] Made sure I have changed the PR base branch, so that it is not `published`. (Sorry for the nagging.)
- [X] Made sure there is an issue registered with a clear problem statement that this PR addresses, (created the issue if it was not present).
    - [X] Updated the `[Unreleased]` entry in `CHANGELOG.md`, linking the issue(s) that the PR is addressing.
    - [X] Referenced the issue I am fixing/addressing _in a commit message for the pull request_.
        - [X] If I am fixing the issue, I have used [GitHub's fixes/closes syntax](https://help.github.com/en/articles/closing-issues-using-keywords)
- [X] Figured if **anything** about the fix warrants tests on Mac/Linux/Windows/Remote/Whatever, and either tested it there if so, or mentioned it in the PR.
- ~[ ] Added to or updated docs in this branch, if appropriate~
- [X] Tests
  - [X] Tested the particular change
  - [X] Figured if the change might have some side effects and tested those as well.
- [X] Formatted all JavaScript and TypeScript code that was changed. (use the [prettier extension](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) or run `npm run prettier-format`)
- [X] Confirmed that there are no linter warnings or errors (use the [eslint extension](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint), run `npm run eslint` before creating your PR, or run `npm run eslint-watch` to eslint as you go).

Ping @pez, @bpringe, @corasaurus-hex, @Cyrik

PR #2286, opened 2023-08-15
cljfmt 11.0.x extra indents fix

### LINKED ISSUE #2280: Update the indenter to handle the cljfmt `0.11.x` breaking config update since version 2.0.383
From  https://github.com/weavejester/cljfmt/tree/master#breaking-changes-in-011x:

> The `:indents` key has been split into `:indents` and `:extra-indents`. The `:indents` key replaces all default indents, while the :`extra-indents` key will append to the default indents.

This probably should be mentioned in the Calva docs. Perhaps it calls for an update to the docs more generally to include the `:extra-indents`, since that is probably the most common configuration from now on.

It also calls for an update of how the Calva indenter reads the configuration. We probably need to recreate the logic now used in **cljfmt** for this, including the `:legacy/merge-indents?` escape hatch.

PEZ:
## What has changed?

* Makes the indenter read `:extra-indents` and `:indents` from the cljfmt configuration, and merge in cljfmt/default-indent:
    * if `:indents` is not used, or
    * if `:indents` are used and `:legacy/merge-indents?` is true.
* Fixes a bug with #2285 (not released yet) where the lack of an `:indent` map in the config made the indenter fail to parse the config completely.

This

* Fixes #2280 

## My Calva PR Checklist

I have:

- [x] Tests

### DIFF
--- a/src/extension-test/unit/cursor-doc/indent-test.ts
+++ b/src/extension-test/unit/cursor-doc/indent-test.ts
@@ -402,9 +402,7 @@ describe('indent', () => {
 function mkConfig(extraRules: indent.IndentRules, replaceRules?: indent.IndentRules) {
   return {
     'cljfmt-options': {
-      ...{
-        'extra-indents': extraRules,
-      },
+      'extra-indents': extraRules,
       ...(replaceRules ? { indents: replaceRules } : {}),
     },
   };
--- a/src/extension-test/unit/cursor-doc/indent-test.ts
+++ b/src/extension-test/unit/cursor-doc/indent-test.ts
@@ -250,6 +250,24 @@ describe('indent', () => {
         expect(indent.getIndent(defndoc.model, p, letBlockConfig)).toEqual(2);
       });
     });
+    describe('replacing cljfmt defaults', () => {
+      const doc = docFromTextNotation('(let []\n|x)');
+      const defndoc = docFromTextNotation('(defn []\n|x)');
+      const p = textAndSelection(doc)[1][0];
+      const emptyConfig = mkConfig({}, {});
+      it('with empty replace config, does not use the built-in rule for the `let` body', () => {
+        expect(indent.getIndent(doc.model, p, emptyConfig)).toEqual(5);
+      });
+      const someConfig = mkConfig(
+        {
+          '/foo+/': [['inner', 0]],
+        },
+        {}
+      );
+      it('with some config, still uses the built-in the built-in rule for the `let` body', () => {
+        expect(indent.getIndent(doc.model, p, someConfig)).toEqual(5);
+      });
+    });
   });
 
   describe('collectIndents', () => {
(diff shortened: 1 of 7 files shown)


src/extension-test/unit/cursor-doc/indent-test.ts
```diff
@@ -250,6 +250,24 @@ describe('indent', () => {
...
+      it('with empty replace config, does not use the built-in rule for the `let` body', () => {
+        expect(indent.getIndent(doc.model, p, emptyConfig)).toEqual(5);
+      });
+      const someConfig = mkConfig(
+        {
+          '/foo+/': [['inner', 0]],
+        },
+        {}
+      );
+      it('with some config, still uses the built-in the built-in rule for the `let` body', () => {
```
juiceghost, line 268: Is the built-in doubled here unintentionally?

src/extension-test/unit/cursor-doc/indent-test.ts
```diff
@@ -380,10 +398,13 @@ describe('indent', () => {
...
 });
 
-function mkConfig(rules: indent.IndentRules) {
+function mkConfig(extraRules: indent.IndentRules, replaceRules?: indent.IndentRules) {
   return {
     'cljfmt-options': {
-      indents: rules,
+      ...{
+        'extra-indents': extraRules,
+      },
```
juiceghost: I think this destructuring of a single key/value pair inside an object could be superfluous?

So instead of
```
...{
        'extra-indents': extraRules,
      },
```
We should be fine with
`'extra-indents': extraRules,`

src/formatter-config.ts
```diff
@@ -14,18 +14,8 @@ const defaultCljfmtContent =
...
-    ? options
-    : {
-        ...options,
-        indents: options['indents'],
-      };
-}
-
 function configuration(workspaceConfig: vscode.WorkspaceConfiguration, cljfmt: string) {
-  const cljfmtOptions = cljfmtOptionsFromString(cljfmt);
+  const cljfmtOptions = cljsLib.cljfmtOptionsFromString(cljfmt);
```
juiceghost, line 18: Does this change mean that in case of error, we would get the defaultIndents rather than just the indents from options?

Looking at
```
(defn- read-cljfmt
  [s]
  (try
    (as-> s $
      (parse-clj-edn $))
    (catch js/Error e
      (merge default-fmt
             {:error (.-message e)
              :indents cljfmt/default-indents}))))
```

in formatter.cljs

src/extension-test/unit/cursor-doc/indent-test.ts
PEZ: Wouldn't that risk create a `'extra-indents': undefined` entry?

src/formatter-config.ts
PEZ: Yes. “just the indents from options” are not available if the parsing fails.

src/extension-test/unit/cursor-doc/indent-test.ts
PEZ: Ah, I see what you mean now. Fixed it. Thanks!

PR #2297, opened 2023-08-23
readme: make sponsor directive easier to find

dustingetz:
**change**: Readme change only: made sponsorship call to action stand out.

**problem**: I was scanning for the sponsor button and had trouble finding it (had to read a bunch of text that i didn't want to read, was about to give up :) )

**test impact**: zero

Checklist:
* I targeted `published` as it is a readme change only, if this is incorrect should be easy to cherry pick
* Ping @PEZ


PEZ:
Thanks!

PR #2302, opened 2023-08-29
Subscribe repl process stdout

### LINKED ISSUE #2300: stdout from sub threads is not printed in the terminal
When spawning off a thread and printing to stdout from the thread, the output is printed from the repl/Clojure process and not picked up by Calva's nREPL client. It can create a bit of confusion, like with this [missionary](https://github.com/leonoel/missionary) code where only `Hello` is printed in the REPL/Output window and the other two strings are printed in the jack-in pseudo terminal:

```clojure
  (def slowmo-hello-world
    (m/sp (println "Hello")
          (m/? nap)
          (println "World")
          (m/? nap)
          (println "!")))

  (let [begin (System/currentTimeMillis)]
    (m/? slowmo-hello-world)
    (println "Took: " (- (System/currentTimeMillis) begin) " ms"))
Hello
nil
clj:hello.example:=>
```

And in the jack-in terminal this is printed:
```
World
!
```

PEZ:
## What has changed?

As part of the Clojure REPL connect process we now check the setting `calva.redirectServerOutputToRepl`, and if it is `true` (the default) we send the `out-subscribe` message to the nREPL server. This makes the server rebind stuff in there catching output that would otherwise be printed by the repl server, sending it as `out` and `err` messages to the the Calva nREPL client.

* Fixes #2300

## My Calva PR Checklist

I have:

- [x] Tests

### DIFF
--- a/docs/site/output.md
+++ b/docs/site/output.md
@@ -136,6 +136,10 @@ In full stack projects, you will probably use the window as a REPL for both `clj
 
 ![Toggle CLJC](images/howto/cljc-toggle-button.png)
 
+## REPL process output (stdout and stderr)
+
+When evaluating code, `stdout`, and `stderr` stemming from `(println ...)` et al is captured by Calva and sent to the Output window. By default this includes output from threads spawned by the evaluated code as well as a lot of other output from the repl process. You can opt out of the “this includes” behaviour with the `calva.redirectServerOutputToRepl` setting.
+
 ## Known Quirks
 
 Due to limitations in the VS Code API it is hard for Calva to know if the output file is opened, and also if it is opened more than once. Make it a habit to leave this window opened. And if it is opened in several tabs, expect evaluation printouts to be a bit unpredictable.
--- a/docs/site/output.md
+++ b/docs/site/output.md
@@ -138,7 +138,15 @@ In full stack projects, you will probably use the window as a REPL for both `clj
 
 ## REPL process output (stdout and stderr)
 
-When evaluating code, `stdout`, and `stderr` stemming from `(println ...)` et al is captured by Calva and sent to the Output window. By default this includes output from threads spawned by the evaluated code as well as a lot of other output from the repl process. You can opt out of this behavior by setting the `calva.redirectServerOutputToRepl` setting to `false`.
+When Calva is connected to the REPL, the Output window will by default print not only results of evaluations, but also:
+
+1. Things printed to `stdout` and `stderr` in the **main thread** of the evaluations
+2. Things printed to `stdout` and `stderr` from **child threads** of the evaluations
+3. Anything printed to `stdout` and `stderr` by the REPL process
+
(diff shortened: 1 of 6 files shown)


docs/site/output.md
```diff
@@ -136,6 +136,10 @@ In full stack projects, you will probably use the window as a REPL for both `clj
 
 ![Toggle CLJC](images/howto/cljc-toggle-button.png)
 
+## REPL process output (stdout and stderr)
+
+When evaluating code, `stdout`, and `stderr` stemming from `(println ...)` et al is captured by Calva and sent to the Output window. By default this includes output from threads spawned by the evaluated code as well as a lot of other output from the repl process. You can opt out of the “this includes” behaviour with the `calva.redirectServerOutputToRepl` setting.
```
bpringe:
```suggestion
When evaluating code, `stdout`, and `stderr` stemming from `(println ...)` et al is captured by Calva and sent to the Output window. By default this includes output from threads spawned by the evaluated code as well as a lot of other output from the repl process. You can opt out of this behavior by setting the `calva.redirectServerOutputToRepl` setting to `false`.
```

I think it's fine to just say "this behavior" as it implies the previously stated behavior. I also think it's clearer to say what value they should give that setting to opt-out.

package.json
```diff
@@ -906,6 +906,11 @@
                 "cljs": null
               }
             }
+          },
+          "calva.redirectServerOutputToRepl": {
+            "type": "boolean",
+            "markdownDescription": "Redirect the output from the REPL server process to the REPL/Output window. With this disabled output from child threads of evaluations will not be shown in the REPL/Output window, but instead in the terminal where the repl is started. (The Jack-in terminal if Calva started the repl).",
```
bpringe:
```suggestion
            "markdownDescription": "Redirect the output from the REPL server process to the REPL/Output window. With this disabled output from child threads of evaluations will not be shown in the REPL/Output window, but instead in the terminal where the REPL is started (the jack-in terminal if Calva started the REPL).",
```

bpringe (approved):
Just left a couple of suggestions.

docs/site/output.md
PEZ: Maybe the whole section is a bit unclear. What I am trying to say is that the setting only affects part of the output. That's why I tried to qualify which behaviour. I'm merging your suggestion, and will look at the section again to see if I can clarify what it all is about.

package.json
PEZ: Thanks! I now ran a search on the repo and see we use `repl` about 20% of the time in strings like this. Dunno, but I think I prefer lowercase. But in any case (haha) we should be consistent, I think.

docs/site/output.md
PEZ: I've pushed an updated version of the copy. Please have a look and see if you think it's an improvement.

package.json
bpringe: Yeah, I tend to use lowercase in conversations.

docs/site/output.md
bpringe: Looks good to me!

PR #2326, opened 2023-10-11
Fix for ClojureScript Quick Start path bug

### LINKED ISSUE #2325: [Windows] The ClojureScript Quick Start REPL Experiences are broken
Hello! This is the follow up issue to our investigation in the Clojure slack. I will also make a PR with a proposed fix to resolve this issue after. 

## Problem 
On Windows, running the "Quick Start ClojureScript" commands for either Node and Browser does not work.

## Expectation
When running a quick start command, Calva should boot up its own environment to immediately allow the user to start using some of Calva's neat features in a core `.cljs` file. 

## Outcome
When running the command, nothing happens. There's no feedback. Not in the output window, the terminals, or any popups. Some sort of behavior is happening in the background, and isn't being reflected to the user.

AlbertSnows:
## What has changed?

- I have removed a catch that was silently hiding an exception that can cause calva to not display an error to the user
- I have added a fix to the code where an exception is thrown via a problematic file path join. Instead of doing path.join on an array spread, we remove that join and do joinPath on the array that we were trying to join on. 

More explanation is provided below the checklist. 

Fixes #2325

## My Calva PR Checklist

I have:

- [ ] Tests

## Why is this happening? 

This issue was uncovered in a huddle session less than an hour ago at time of writing. Explanations to follow. 

### Why isn't it showing an error to the user? 

It's not showing an error to the user because the exception that is (presumably) thrown is being caught and silently logged to a console that (I think) isn't exposed to the user. Specifically, this line. 

https://github.com/BetterThanTomorrow/calva/blob/57a24c859b07a0ed6786dbe285a1add7f0727ddb/src/nrepl/repl-start.ts#L91

### Why isn't the command behaving as expected?

The command isn't working because of nuances with javascript's join behavior. Look at this line. 
https://github.com/BetterThanTomorrow/calva/blob/57a24c859b07a0ed6786dbe285a1add7f0727ddb/src/nrepl/repl-start.ts#L72

When javascript joins these lines it results in code such as `src\\hello_world` instead of the desired `src/hello_world`. This is an invalid directory format, so the directory cannot be created! Thus when trying to act on these assumed created files, the function will throw an exception that is silently caught. 

## What is the solution? 

The solution is twofold. First, remove the catch so that this error bubbles up and notifies the user. Second, remove the join since it's not needed to build the path connection for the file creation location; as follows. 

```
  const directoryPath = path.dirname(filePath).split(/\//);
  const dirUri = vscode.Uri.joinPath(storageUri, ...directoryPath);
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 Changes to Calva.
 
+## [Unreleased]
+
 ## [2.0.391] - 2023-10-11
 
 - Fix: [Allowing dram download to fail vocally for user. Fixing a JS join path error](https://github.com/BetterThanTomorrow/calva/issues/2325)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 Changes to Calva.
 
-## [Unreleased]
+## [2.0.391] - 2023-10-11
+
+- Fix: [Allowing dram download to fail vocally for user. Fixing a JS join path error](https://github.com/BetterThanTomorrow/calva/issues/2325)
 
 ## [2.0.390] - 2023-10-11
 
--- a/src/nrepl/repl-start.ts
+++ b/src/nrepl/repl-start.ts
@@ -69,7 +69,8 @@ async function fetchConfig(configName: string): Promise<DramConfig> {
 
 async function downloadDram(storageUri: vscode.Uri, configPath: string, filePath: string) {
   const downloadUrl = `${dramsBaseUrl()}/${configPath}/${filePath}`;
-  const dirUri = vscode.Uri.joinPath(storageUri, path.join(...path.dirname(filePath).split(/\//)));
+  const directoryPath = path.dirname(filePath).split(/\//);
+  const dirUri = vscode.Uri.joinPath(storageUri, ...directoryPath);
   await vscode.workspace.fs.createDirectory(dirUri);
   const storeFileUri = vscode.Uri.joinPath(storageUri, path.join(...filePath.split(/\//)));
   return await utilities.downloadFromUrl(downloadUrl, storeFileUri.fsPath).catch((err) => {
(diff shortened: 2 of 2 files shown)


AlbertSnows:
@PEZ One thing I want to highlight is the outcome of side effects. I've tested on windows, but since I don't have the domain knowledge to assess the impact on either different windows platforms or other operating systems, I would suggest ensuring that this change doesn't break other assumptions from said platforms. 
Fortunately, this function is only used in one spot. As such, I think the only question that comes to mind in terms of possible side effects of this change is the possible states of the parameters. There are three of them. 
- storageUri (of type vscode.Uri)
- configPath (of type string)
- filePath (of type string)

We know the types, but the remaining question in my mind is, "Are there any formats that these types come in that worked previously but may be broken with these new changes?" My assumption is no, but I will defer to you experience to make that judgement.

src/nrepl/repl-start.ts
```diff
@@ -69,7 +69,8 @@ async function fetchConfig(configName: string): Promise<DramConfig> {
 
 async function downloadDram(storageUri: vscode.Uri, configPath: string, filePath: string) {
   const downloadUrl = `${dramsBaseUrl()}/${configPath}/${filePath}`;
-  const dirUri = vscode.Uri.joinPath(storageUri, path.join(...path.dirname(filePath).split(/\//)));
```
PEZ, line 72: We have had a lot of problems with this category of mistakes. Most oft them we have found, but this one was still lurking in there! (I think we may still have some places undiscovered.)

src/nrepl/repl-start.ts
```diff
@@ -84,13 +85,9 @@ export async function downloadDrams(
 ) {
   await Promise.all(
```
PEZ, line 86: Note to self: Consider returning the promise instead of awaiting it. Similar to how we shouldn't have catched the error here, I think it makes sense to push the promises out and up as close to the user as possible.

PEZ:
Thanks for helping with improving Calva! 🙏 ❤️. Code looks great. I will try it on my Mac just to make sure I am understanding the issue correctly.

We could consider adding this to the end-to-end tests. What speaks against it is that the tests are already dependent on a lot of external things and this would add more of that.

PEZ:
It still works on my Mac. (Or course it does!)

Good points about validation of the inputs there. At the moment we have full control of the whole chain, but my ambition is that more people will contribute drams (the configurations behind those Quick Start experiences) and then it will be very good to remember to help with that by validating inputs and generating good error messages.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ Changes to Calva.
 
 ## [Unreleased]
 
+## [2.0.391] - 2023-10-11
+
+- Fix: [Allowing dram download to fail vocally for user. Fixing a JS join path error](https://github.com/BetterThanTomorrow/calva/issues/2325)
```
PEZ, line 9: I was reviewing a bit too fast. This entry should not have a date, nor a version. It should remain just **Unreleased**. But no worries, I fixed it on the `dev` branch.

PR #2329, opened 2023-10-22
Fix structural edit hang

### LINKED ISSUE #2299: Bug: backspace (structural editing) occasionally "hangs"
I'm editing code and start typing a symbol or I'm typing into a string, realize I've mistyped and try to backspace but nothing happens -- I can type more characters but cannot backspace -- and then after a brief period my backspaces suddenly take effect at wherever the cursor happens to be at that point.

It's very odd. I can click elsewhere in the file. I can move the cursor with the arrow keys. Typing anything except backspace seems to work fine. And then those "missing" backspaces get actioned at the (new) cursor position.

It's a fairly recent phenomenon and it happens randomly between a few and maybe half a dozen times a day.

PEZ:
## What has changed?

I seems it was my changes to how we find the `ns` from the current cursor position that caused a stack overflow in unbalanced structure and thus #2299.

I've added a check for this kind of unbalance and stop the recursion by letting the search for the ns continue from the start of the document in these cases.

I've also added some tests for the unbalance-check, and while doing so notice that it doesn't actually find unbalance if it is from a lack of closing brackets, only lack of opening brackets. I couldn't immediately fix it, so left some disabled tests, and added TODOs. We've been living with this bug for long, (I don't even remember when I added the code for unbalance check). Besides, I think the real way to check for unbalance is to piggy back on the highlighter's work, which does a much better job at it, and also, since it is doing the job, we shouldn't be doing it over and over again at other places.

* Fixes #2299

## My Calva PR Checklist

I have:

- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ Changes to Calva.
 ## [Unreleased]
 
 - Bump bundled deps.clj to v1.11.1.1413
+- Add timestamps to nREPL message log diagnostics
+- Fix: [Backspace (structural editing) occasionally hangs](https://github.com/BetterThanTomorrow/calva/issues/2299)
 
 ## [2.0.391] - 2023-10-12
 
--- a/src/cursor-doc/paredit.ts
+++ b/src/cursor-doc/paredit.ts
@@ -778,15 +778,6 @@ export function open(
   }
 }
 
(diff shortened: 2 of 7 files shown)


PEZ:
Hi there everyone! I am hoping that the latest release, v2.0.392, fixes this. Please let me know if you agree or if the issue is still there.

PR #2344, opened 2023-11-11
2343 squint nrepl

### LINKED ISSUE #2343: Make Calva's nrepl client squint-cljs compatible
There's a new nREPL server in town: for [squint-cljs](https://github.com/squint-cljs/squint)

Calva fails to connect to it, and fixing that I bump into one error after another preventing me from evaluating anything. CIDER can do it, so I think it highlights things we need to fix in our client.

TODO: Document here what errors I get. (Or I might mention those in the PR that is hopefully coming up).

PEZ:
## What has changed?

The way we handle ClojureScript-only repls should really be fixed. I think the proper fix is to fix the way we handle nREPL sessions all together, and that it is not meaningful to do a fix that only addresses the ClojureScript-only question. Anyway, the way we handle it is to:
1. connect the ClojureScript REPL as a CLJ session,
2. clone the session as a CLJS-to-be session (although it already is)
3. Fake that we upgrade it to a CLJS session

When that is done there are two sessions, CLJ and CLJS, both connected to the same REPL. But the user will probably only interact with one, so it works suprisingly well.

However, the squint-cljs REPL evaluates keywords to string, so our fake upgrade evaluation failed. We are now evaluating the number `42` instead, which should stay consistent even with crazy future REPLs.

While at it I also realized that we were not using the `ns` field of the `eval` op, and that that was the reason we had to do all those switchNS shenanigans. I have therefore now removed the switchNS calls, and replaced it with an evaluation of the current file's `ns` form when the current ns changes.

For this I choose to change the return type of the function that finds the ns of the current file. Now it returns both the ns and the form text.

I've tested it with a bunch of different projects. Will ask in the Calva channel for help with more testing.

* Fixes #2343

## My Calva PR Checklist

I have:

- [x] Tests

### DIFF
--- a/src/nrepl/index.ts
+++ b/src/nrepl/index.ts
@@ -372,10 +372,9 @@ export class NReplSession {
     });
   }
 
-  async evaluateInCurrentNs(nsForm: string) {
+  async evaluateInNs(nsForm: string, ns: string) {
     try {
-      console.log('Evaluating in current ns: ', nsForm, this.client.ns);
-      await this.eval(nsForm, this.client.ns).value;
+      await this.eval(nsForm, ns).value;
     } catch (e) {
       console.error(e);
     }
--- a/src/nrepl/index.ts
+++ b/src/nrepl/index.ts
@@ -372,6 +372,15 @@ export class NReplSession {
     });
   }
 
+  async evaluateInCurrentNs(nsForm: string) {
+    try {
+      console.log('Evaluating in current ns: ', nsForm, this.client.ns);
+      await this.eval(nsForm, this.client.ns).value;
+    } catch (e) {
+      console.error(e);
+    }
+  }
+
   async requireREPLUtilities() {
     const CLJS_FORM = `(try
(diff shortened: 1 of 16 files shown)


src/nrepl/index.ts
```diff
@@ -372,8 +372,12 @@ export class NReplSession {
     });
   }
 
-  async switchNS(ns: any) {
-    await this.eval(`(in-ns '${ns})`, this.client.ns).value;
+  async evaluateInNs(nsForm: string, ns: string) {
+    try {
+      await this.eval(nsForm, ns).value;
```
bpringe, line 377: Why is the first arg called `nsForm`? This, generically, could be any code form, right?

PEZ: Good catch. I renamed and changed this function quite a bit while working on this. Maybe the best is to rename the function to indicate that it is only really meant to evaluate the ns form. For other cases, we have the eval function already.

PR #2397, opened 2024-02-11
html2hiccup Dedicated option to prevent Hiccup-like class shortcuts option

### LINKED ISSUE #2390: Provide option to not have Hiccup conversion convert class names to Hiccup dot shortcut class names
When using the popular [Tailwind CSS](https://tailwindcss.com/) it not uncommon to have class names like this in your HTML:
```
<a href="#" class="inline-block bg-[#5890FF] min-w ...">
  ...
</a>
```
When pasting this as Hiccup I get:
```
[:a.inline-block.bg-[#5890FF].min-w.... {:href "#"}
 "..."]
```
... which is not valid Clojure.

And so I propose an option to not bother massaging class names into the Hiccup shortcut-style, instead just leaving it as a normal class attribute like so:
```
[:a {:class "inline-block bg-[#5890FF] min-w ..." :href "#"}}
 "..."]
```

### LINKED ISSUE #2398: Provide a dedicated option to not have Hiccup conversion convert class names to Hiccup dot shortcut class names
This is a continuation of issue #2390 which was mostly addressed by @PEZ in PR #2390, save the actual explicit option for always leaving classes in the attribute.

gerdint:
## What has changed?

- Adds :add-classes-to-tag-keyword? option to HTML to Hiccup converter.
- I don't find the code worse than before the option is added (though it would be nice to make the remaining-attrs let-binding clearer, I haven't touched it in this PR)

This option is particularly useful when using Tailwind CSS which involves huge amounts of classes, so putting them in the tag position makes it less readable (but I do tend to use the dot syntax with some main class names)

Fixes #2398

## My Calva PR Checklist

I have:

~~If I am fixing just part of the issue, I have just referenced it w/o any of the "fixes” keywords. ~~
~~Figured if **anything** about the fix warrants tests on Mac/Linux/Windows/Remote/Whatever, and either tested it there if so, or mentioned it in the PR. ~~
- [x] Tests
~~Figured if the change might have some side effects and tested those as well.~~


gerdint:
Hmm, not sure why there are 3 extra commits in the PR not authored by me, will investigate.

PEZ:
Thanks for doing this! ❤️ 🙏

PEZ:
> * I don't find the code worse than before the option is added

It's not really about wether the code is worse or not. The change is clear enough, but it adds yet another code path to a function where I don't think the code is super clear. This was why I didn't add this in my last PR. It really would be nice to try clear the code up a bit, making room for the extra branching. But maybe it's OK as it is for now, and we can return to the tidying up later. Please add the docs and the settings and we will have this merged.

gerdint:
OK!

I noticed that you also decided to subject the id attribute to the same cautionary valid-as-hiccup-kw? treatment. FWIW I do not think this is necassary in that I have never seen HTML code with a similarly wretched id attribute. Defensive programming never hurts, but I would argue perhaps uncalled for here and that slightly added complexity could perhaps be traded for this one! :)

Will update docs and settings!

PEZ:
Well, I had never seen those weird classnames before. Awareness can be a curse. =)

gerdint:
Ah yes, I wish I could give the sensible CSS class names blue pill! ;)

docs/site/hiccup.md
```diff
@@ -54,6 +54,7 @@ The Hiccup converstion can be tweaked with two options using the setting `calva.
 
 * `mapify-style`: boolean, default `false`. When `true` any `style` attribute will be converted to a map ([Reagent](https://reagent-project.github.io/) supports this)
 * `kebab-attrs?`: boolean, default `false`. When `true` attribute names will be converted from *camelCase*, or *snake_case/SNAKE_CASE* to *kebab-case*. (Reagent wants most attribute names like this.)
+* `no-class-shortcuts?`: boolean, default `false`. When `true` all class names will remain in the class   attribute (as opposed to being converted to Hiccup's CSS-style `[:tag.clz1.clz2]` shortcuts). This may be preferable in the event that elements tend to have a lot of class names, such as when using Tailwind CSS.
```
PEZ: I think an option saying something like `:add-classes-to-tag-keyword?` that defaults to `true` makes sense. My brain doesn't handle double negations very well. 😄

package.json
```diff
@@ -860,6 +860,11 @@
                 "type": "boolean",
                 "markdownDescription": "Should the attribute names be converted to a kebab-case?",
                 "default": false
+              },
+              "no-class-shortcuts?": {
+                "type": "boolean",
+                "markdownDescription": "Should all class names remain in the class attrbute?",
```
PEZ: Consider reusing more of the description from the docs here. Maybe a combo of reusing and linking to the docs makes the sweet spot.

docs/site/hiccup.md
gerdint: Good idea

gerdint:
@PEZ So I changed behaviour and name of the option to match your suggestion. But when I test the default `true` value does not seem to take effect. If I don't set it in my settings file the default value appears to be `false`, which obviously breaks havoc on tests.

I suspect the settings logic in converters.ts need to something in order for default true values to work. Do you have an idea on the top of your head?

PEZ:
I think that the defaults have never worked, but that they evaluate to false because `undefined` or something.

The defaults should be provided as a map, sibling to `"properties"`. Something like so:

```json
            "default": {
              "mapify-style?": false,
              "kebab-attrs?": false,
              "add-classes-to-tag-keyword?": true
            }
```

gerdint:
@PEZ Adding that object as a sibling of "properties" did not seem to make a difference for me. any other advice? I guess this can be worked around closer to the html2hiccup ns, but would be good to know the intended approach.

PEZ:
I'll have a look at it. It shouldn't be worked around. 😄

PEZ:
Works on my machine! 😄 

When I place the defaults as a separate map, and run without custom settings the classes are tucked on the tag keword, and if I add a setting the defaults are inserted as they should, and if if I change the new setting to false the classes are instead moved to the props map.

I'm thinking that maybe when you added defaults as a separate map you also removed the defaults from where they shouldn't be? And maybe when you did that you left some trailing comma, and then the manifest doesn't parse and you will not be testing with your change. Check the formatter task output for errors. Iirc it will complain about these things.

PEZ:
With defaults:
```clojure
[:tbody.d-block
 [:tr.d-block
  [:td.d-block.comment-body.markdown-body.js-comment-body
   [:p {:dir "auto"}
    [:a.user-mention.notranslate {:data-hovercard-type "user"
                                  :data-hovercard-url "/users/PEZ/hovercard"
                                  :data-octo-click "hovercard-link-click"
                                  :data-octo-dimensions "link_type:self"
                                  :href "https://github.com/PEZ"}
     "@PEZ"]
    "Adding that object as a sibling of \"properties\" did not seem to make a difference for me. any other advice? I guess this can be worked around closer to the html2hiccup ns, but would be good to know the intended approach."]]]]
```

`"add-classes-to-tag-keyword?": false`:

```clojure
[:tbody {:class ["d-block"]}
 [:tr {:class ["d-block"]}
  [:td {:class ["d-block" "comment-body" "markdown-body" "js-comment-body"]}
   [:p {:dir "auto"}
    [:a {:data-hovercard-type "user"
         :data-hovercard-url "/users/PEZ/hovercard"
         :data-octo-click "hovercard-link-click"
         :data-octo-dimensions "link_type:self"
         :href "https://github.com/PEZ"
         :class ["user-mention" "notranslate"]}
     "@PEZ"]
    "Adding that object as a sibling of \"properties\" did not seem to make a difference for me. any other advice? I guess this can be worked around closer to the html2hiccup ns, but would be good to know the intended approach."]]]]
```

Passing with flying colors!

gerdint:
Well indeed, it works when I invoke the command manually.
My concern was that the change still causes 9 tests to fail, and that seems to be because the true default value is not sticking:

```
npm run calva-lib-test 
...
========= Running Tests =======================
[:calva-lib] Build completed. (268 files, 1 compiled, 0 warnings, 8,31s)
...

Ran 118 tests containing 306 assertions.
9 failures, 0 errors.
===============================================
[:test] Build completed. (233 files, 3 compiled, 0 warnings, 3,31s)

```
Now that I inspect it I see that `{:add-classes-to-tag-keyword? true}` probably has to be added to all test cases in html2hiccup_test.cljs. I will come up with some helper function to try to minimize the noise!

gerdint:
Now I did the least intrusive way, but perhaps it would be clearer to do like I suggested above, to introduce a helper fun in html2hiccup_test.cljs that the tests call instead of sut/html->hiccup?

gerdint:
@PEZ How about making the options a dynamic binding?

PEZ:
Not sure I fully understand the problem. But anyway. Would it work with using a default value in `html->hiccup?`. It spreads the knowledge about what is the default, but I think that's ok for this particular config.

Or we could accept the noise.

gerdint:
Well, that's what the above latest commit does. I do not have a good grasp of the whole config system, and so I think consistency is most important. But I perhaps we don't have an other options that defaults to true prior to this, and the situation novel? That said, I think the having the default inside html->hiccup is quite common practice and OK. It's the syncing with the VS Code settings system that is the question I guess.

PEZ:
As long as the VS Code settings override the defaults, we're not in such a bad place. And as long as we have tests demonstrating that the feature can be toggled from the outside.

What we do for the formatter is that we read the cljfmt defaults from CLJS land. But that's a lot of config, and the defaults are not determined by us, so it is simply not an option to have defaults in two places.

I think we're good. If you are happy, it is time to merge. I'll let you fix the conflicts, lest we step on each other's work.

gerdint:
@PEZ OK I'm done

PEZ:
Thanks for helping with making Calva better, @gerdint! ❤️ 🙏  And thanks for making it a pleasure working with you.

PR #2411, opened 2024-02-27
Automatically show rename UI after refactors that add names

### LINKED ISSUE #2388: Feature request: make it easy to pick the name for the new binding created after running one of the "extract-" code actions
Currently when you run one of the code actions that creates a new binding (move to let, extract fn, etc.), it picks the name of the binding for you, and you have go to change both uses of it manually. Pylance, the Python language extension helpfully prompts for a name for the binding:

https://github.com/BetterThanTomorrow/calva/assets/2607086/756c968c-316b-488e-91d5-032100c77feb

Upon further investigation, it appears that the way this is accomplished is Pylance consists of _both_ a VSCode extension _and_  a LSP server. When an action that creates a new binding is sent to the LSP, the extension immediately follows the resolution of that action with a "rename symbol" action.

It would be nice if Calva were to do something similar to help devs rename bindings newly created by code actions, either by using Pylance's rename-symbol approach, or by highlighting all instances of the new symbol with multiple cursors. Personally I kind of like the multiple cursors approach because it doesn't put anything in front of the code, but either approach would be an improvement.

maxrothman:
Fixes #2388

Most language extensions automatically trigger the rename UI after completing a refactor command that introduces new names, such as Javascript's "Extract to constant" commands. Previously, for such commands run from the command palette, Calva prompted for the new name in a palette, and did not support auto-renaming commands run from quick-fix actions (under the lightbulb icon), whatever default name was returned from the LSP would be used. Now, Calva uses the Rename Symbol feature in both cases.

To support this, introduce two new properties of ClojureLspCommand, `defaultName` and `afterCommandFn`. `afterCommandFn` is called after the initial command completes when it is present on the command object. The only implementation of an `afterCommandFn` presently is `renameAfterRefactor`, which looks for the hardcoded `defaultName` in the edit returned from the LSP and triggers Rename Symbol on the last place where it appears, leaving the user's cursor near where it started.

Also extract clojureLspCommands and moved it above sendCommandRequest because sendCommandRequest uses it to retrieve the afterCommandFn from the original command object, which is not otherwise present in scope.

I have not touched anything to do with tests or documentation, please let me know what if anything needs to be done for those.

## My Calva PR Checklist

I have:

- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Changes to Calva.
 
 ## [Unreleased]
 
+- [Automatically do Rename Symbol after refactors that add new names](https://github.com/BetterThanTomorrow/calva/issues/2388)
+
 ## [2.0.414] - 2024-02-24
 
 - [HTML->Hiccup: Add option to always keep classes as attributes if so desired](https://github.com/BetterThanTomorrow/calva/issues/2398) 
--- a/src/lsp/commands/lsp-commands.ts
+++ b/src/lsp/commands/lsp-commands.ts
@@ -3,14 +3,30 @@ import * as calva_utils from '../../utilities';
 import * as defs from '../definitions';
 import * as vscode from 'vscode';
 import * as api from '../api';
+import * as _ from 'lodash';
 
 const RESOLVE_MACRO_AS_COMMAND = 'resolve-macro-as';
 
-type ClojureLspCommand = {
+type BaseLspCommand = {
   command: string;
-  extraParamFn?: () => Thenable<string>;
   category?: string;
 };
+
+// If afterCommandFn is defined, then defaultName must be too
+type ClojureLspCommand = BaseLspCommand &
+  (
+    | {
+        afterCommandFn: (
+          commandArgs: Array<string | number>,
+          commandResponse: Record<string, unknown>
+        ) => Thenable<any>;
+        defaultName: string;
+      }
+    | {
+        defaultName?: string;
+        afterCommandFn?: never;
+      }
+  );
 
 type LSPCommandHandlerParams = {
   clients: defs.LspClientStore;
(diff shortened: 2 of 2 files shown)


PEZ:
Thanks! 🙏 This great. Even though it only saves me one step, I really enjoy the feeling. And for those that haven't figured out that step it will be even more amazing.

On the first use it seems to miss a bit, even though the character index is correct.

Move to let (cursor before the expression):

image

Move to let (cursor after the expression):

image

(I think clojure-lsp does weird things here too, but want to try on a branch without these changes to judge that better.)

Extract function (cursor before or after the expression, same result):

image

The expression I am able to repro this with is `(first (string/upper-case c))` on line 27, in 
`test-data/projects/pirate-lang/src/pez/pirate_lang.clj`.

With subsequent runs it works perfectly. Don't know if we can do anything about it. Also, editing the file in any way seems to also snap things into working mode, so it only affects someone who hits a quick-fix first thing after opening a project in VS Code. I'm ready to ship if you figure it is too weird/out of reach to figure out a way to avoid it.

maxrothman:
I saw this issue while I was developing too, that it only happens until the file has been edited is a good insight. Unfortunately I don't known what I can do to fix this issue, my instinct is that it has to be coming from clojure-lsp since the editor position is correct. Should we page the LSP team?

maxrothman:
Upon further reflection, this _has_ to be an LSP issue. All that the rename commands takes is a Position and a Uri, the range is chosen from the position, presumably by the LSP.

PEZ:
@ericdallo, what do you make of this? I was thinking it could be VS Code. I can't reproduce it without the changes in this PR. Is there something I can test, provide, whatever to shed more light on it?

@maxrothman, I think we can merge this without having solved this. It is a minor nuisance, and it only happens in a very rare situation (unless you are testing this PR). Let me know when you feel done with the PR checklist.

maxrothman:
Sounds good @PEZ , the only remaining checklist items have to do with docs and tests. Are any doc/test changes necessary for you to feel good about merging this PR? Otherwise I think it's good to go.

ericdallo:
hum, I think I know what's happening, when a code action is applied, the editor send the changes to clojure-lsp to then re-lint the file and generate new analysis for the file, the problem is that we have a debouncer for changes and they run async, so what I think it's happening is that we are renaming it to fast after the code action apply, so clojure-lsp is retunring to rename the position of the old symbol before the code action is applied 😔 Not sure if there is a good fix for that besides sleeping some ms maybe

maxrothman:
Thanks @ericdallo, I’ll try adding a short delay. Any idea why it would only happen when the file was freshly opened, and not after any edits have occurred?

ericdallo:
Not really

PEZ:
In latest VS Code:

https://github.com/BetterThanTomorrow/calva/assets/30010/289af93f-18e8-4178-ba1b-420fc3b4e46d

😄

PEZ:
I have Calva time today, @maxrothman, so wanna merge and release this. We can fix that glitch later. (Or attempt to, I think it looks really hard to fix.)

maxrothman:
🎉 Thanks to both of you for helping my through my first Calva PR! It was fun to contribute. 🎉 

I think I might still try to add a delay to fix that glitch. @ericdallo what's the delay on the debounce in the LSP, so that I can make sure my delay is longer than it?

ericdallo:
@maxrothman it's 300ms, but there is the analysis time on clj-kondo that takes some more considerable ms and it may vary

maxrothman:
I finally had a chance to circle back on this today. @ericdallo your theory appears to be correct: adding a 400ms delay removes the issue, reducing the delay to 200ms makes it reappear.

I'm hesitant to penalize the performance of all refactors by a half-second just to remove an issue only seen if a refactor happens before any edits. At this point unless anyone objects, I'll just leave it as-is. If there were going to be a fix, I feel like we'd have to figure out why the issue only happens before any edits, and that investigation would have to be coordinated with clojure-lsp.

PEZ:
I think it is fine as it is. The behaviour of extracting something first thing when opening a project is probably mostly for us who are testing the feature. And also not the end of the world with the glitch, since it then starts to work.

ericdallo:
Yeah, it there was a option to wait 400ms only for a few specific actions that would be better

maxrothman:
Even then, we'd be penalizing the vast majority of requests to resolve an issue with the first one. If I could do a quick fix I'd wait 400ms on the first request only (or any request in a file that happens before an edit) but that'd be way too clunky.

PR #2474, opened 2024-03-30
Paredit Multicursor - Rewrap

riotrah:
## What has changed?

- Implement multicursor for paredit rewrap
- Like my other multicursor work this time, extracted command logic into a commands.ts file, adding some indirection and opt-in multicursor based on user setting etc
- Add tests

Originally this was going to include multicursor wrap (now a WIP at #2472 ), but a couple of test cases keeping failing, so here's rewrap alone!

Fixes #2473 

## My Calva PR Checklist

I have:

- [x] Read [How to Contribute](https://github.com/BetterThanTomorrow/calva/wiki/How-to-Contribute#before-sending-pull-requests).
- [x] Directed this pull request at the `dev` branch. (Or have specific reasons to target some other branch.)
- [x] Made sure I have changed the PR base branch, so that it is not `published`. (Sorry for the nagging.)
- [x] Made sure there is an issue registered with a clear problem statement that this PR addresses, (created the issue if it was not present).
    - [x] Updated the `[Unreleased]` entry in `CHANGELOG.md`, linking the issue(s) that the PR is addressing.
- [x] Figured if **anything** about the fix warrants tests on Mac/Linux/Windows/Remote/Whatever, and either tested it there if so, or mentioned it in the PR.
- [x] Added to or updated docs in this branch, if appropriate
- [x] Tests
  - [x] Tested the particular change
  - [x] Figured if the change might have some side effects and tested those as well.
- [x] Formatted all JavaScript and TypeScript code that was changed. (use the [prettier extension](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) or run `npm run prettier-format`)
- [x] Confirmed that there are no linter warnings or errors (use the [eslint extension](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint), run `npm run eslint` before creating your PR, or run `npm run eslint-watch` to eslint as you go).

Ping @pez, @bpringe, @corasaurus-hex, @Cyrik

PR #2477, opened 2024-03-30
Paredit Multicursor - SelectCurrentForm

### LINKED ISSUE #2476: Paredit Multicursor - SelectCurrentForm

riotrah:
## What has changed?

~~This PR is targeting my other PR #2474 's branch, such that once that gets merged, this should automatically target `dev`, but I'll leave the 2nd branch target checkbox unmarked just in case for now~~

- Implement multicursor `selectCurrentForm`
- Removes its previous implementation in `src/select.ts` and exposure under root ~~`calva.selectCurrentForm`~~ commands, into a rewritten version in `src/cursor-doc/paredit.ts`/`commands.ts`/`src/paredit/extension.ts` like the other paredit multicursor rewrites exposed as `paredit.selectCurrentForm`.
  - **[EDIT]**: **just keeps the previous command id for backwards compat, thanks @PEZ** 
- Adds tests for single and multi cursor

Fixes #2476 

## My Calva PR Checklist

I have:

- [x] Tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -1562,8 +1562,8 @@
         "enablement": "editorLangId == clojure"
       },
       {
-        "category": "Calva",
-        "command": "paredit.selectCurrentForm",
+        "category": "Calva Paredit",
+        "command": "calva.selectCurrentForm", // legacy id for backward compat
         "title": "Select Current Form",
         "enablement": "editorLangId == clojure"
       },
@@ -2151,7 +2151,7 @@
         "when": "calva:keybindingsEnabled"
       },
       {
-        "command": "calva.selectCurrentForm",
+        "command": "calva.selectCurrentForm", // legacy id for backward compat
         "key": "ctrl+alt+c ctrl+s",
         "when": "calva:keybindingsEnabled && editorLangId == clojure && editorTextFocus"
       },
@@ -3078,7 +3078,7 @@
         },
         {
           "when": "editorLangId == clojure",
-          "command": "calva.selectCurrentForm",
+          "command": "calva.selectCurrentForm", // legacy id for backward compat
           "group": "calva/a-structural-editing"
         },
         {
--- a/package.json
+++ b/package.json
@@ -1344,12 +1344,6 @@
         "category": "Calva"
       },
       {
-        "command": "calva.selectCurrentForm",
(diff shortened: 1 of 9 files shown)


riotrah:
@PEZ both this pr and the `rewrap` one both seem to fail the netlify deploy process, I'm guessing from the paredit docsite changes - but I can't derive any useful info from the deploy logs.

package.json
```diff
@@ -1343,12 +1343,6 @@
         "enablement": "calva:connected",
         "category": "Calva"
       },
-      {
-        "command": "calva.selectCurrentForm",
-        "title": "Select Current Form",
-        "category": "Calva",
-        "enablement": "editorLangId == clojure"
-      },
```
PEZ, line 1351: I don't think we should retire command ids. They can be part of peoples keybindings configuration and Joyride scripts. Our options here is to either let it reside in the wrong category, or have two command ids point to the same function.

riotrah: Ah of course, I even discovered this missing command in the first place because I tried my custom `selectCurrentForm` shortcut with multicursor enabled and noticed it didn't do it! Thanks for catching this

riotrah: 1. If we make a duplicate reference command:
   - 👍🏽 users can easily find it amongst `paredit.*` commands in the palette or the keyboard shortcuts UI or as part of autocomplete in the shortcuts JSON
   - 👍🏽 we can (maybe?) rename the existing command's title to "Select Current Form (legacy)" or something, to make it clearer as users discover commands that there's a "legacy" one and a "normal" one, even if they're really the same.
   - 👎🏽 this creates 2 duplicate commands with no obvious mechanism to migrate users' existing shortcuts (unless there _is_ a vsc system for that??)

2. If we keep the old, single command:
   - 👍🏽 simpler, no migration question even arises
   - 👎🏽 less obviously discoverable or grouped as part of `paredit`

riotrah: After thinking about it, I'm basically complete impartial

riotrah: wish we had analytics for command invocations haha

riotrah: i'll keep it at 1 cmd, but change the title

PR #2481, opened 2024-03-31
Add webview output experience

### LINKED ISSUE #2480: Add webview for REPL output
The idea behind this output location is for it to be a read-only webview. The initial release will likely not come with any additional mechanism for input beyond using `comment` forms in files, which is fine by me, because I never type / eval code in the output window.

If desired in the future, we can add an additional input mechanism, which can even live inside the webview, but would be separate from the output area. Mixing input and output in the same container adds complexity that we want to avoid going forward.

bpringe:
## What has changed?

A new webview for REPL output is added. It's a new output location in settings. 

Closes #2480 

## My Calva PR Checklist

I have:

- [x] Tests


package.json
```diff
@@ -3254,6 +3254,7 @@
     "open": "^6.3.0",
     "parinfer": "^3.12.0",
     "posthtml-parser": "^0.11.0",
+    "react-dom": "^18.2.0",
```
PEZ: I don't know if we need React for something in particular. If we do, then of course we should use it. Otherwise, it may be better to use something with less bloat, like Dumdom, or maybe Replicant.

bpringe: It's for reagent, which I consider very lightweight and should be used lightly for this. If you think there's any performance issue or other issues with it that would be significant for our use-case, we should discuss that, but switching it out later should be pretty simple.

bpringe: I'm open to alternatives if there's a good reason to use something else.

bpringe: Dumdom looks interesting. Using a library that supports react gives us reach into the react ecosystem though, which could be nice in the future for more rich features and interactivity (making it less likely we'll need to recreate the wheel).

PEZ:
Shall we try to release this behind an opt-in setting as quickly as we can?

package.json
PEZ: My experience with React is that it changes all the time in breaking ways, making it high maintenance. With Dumdom we would avoid this, as the vdom library it uses is super tiny and stable. With Replicant we would get a pure ClojureScript vdom as well, and have even more of the familiar Clojure stability (though I am not sure if the Replicant API is marked as stable yet).

However, I am not opposed to React if we have a good reason to use it.

bpringe: I say we move forward with it for a first-release, and maybe be conscious of our usage of it (keeping it light), and if we come across issues we can switch it out later.

I'll continue to keep other options in mind and think about the React usage for now...

bpringe: > I say we move forward with it for a first-release

I may reconsider this. I need to look at Dumdom more closely.

bpringe:
> Shall we try to release this behind an opt-in setting as quickly as we can?

Sorry, I'm moving slowly with this so far due to being busy with other life things. It's not quite ready but there's not that much more to do before getting it to a very basic usable state. I've started dedicating a little focused dev time on weekends for this.

package.json
PEZ: Having had a closer look at Replicant, In think it's better to use that instead of Dumdom. They are quite similar, but Replicant drops the dependency on Snabbdom, as well as some legacy API requirements that Dumdom carries.

mkdocs.yml
```diff
@@ -45,74 +45,75 @@ markdown_extensions:
...
-    - rich-comments.md
-    - hiccup.md
+      - commands-top10.md
+      - evaluation.md
+      - formatting.md
+      - syntax-highlighting.md
+      - paredit.md
+      - REPL:
+          - repl-window.md
+          - output-webview.md
```
bpringe: This was added. Just noting that since there are whitespace changes which might make that not obvious.

package.json
```diff
@@ -3311,16 +3331,16 @@
   },
   "scripts": {
     "watch-docs": "mkdocs serve",
-    "clean": "rimraf ./out && rimraf ./tsconfig.tsbuildinfo && rimraf ./cljs-out",
```
bpringe, line 3314: rimraf takes multiple paths

src/cljs-lib/src/calva/repl/webview/core.cljs
```diff
@@ -0,0 +1,230 @@
...
+    <div id=\"output\"></div>
+
+    <script src=\"" js-source "\"></script>
+  </body>
+</html>"))
+
+(defn get-js-source
+  [{:keys [vscode/vscode]
+    vscode-context :vscode/context}
+   {:keys [^js webview-panel]}]
```
bpringe, line 90: Following an arg pattern of `[context fn-specific-args]`

src/cljs-lib/src/calva/repl/webview/core.cljs
```diff
@@ -0,0 +1,229 @@
...
+    (.. ^js webview-panel (onDidDispose (fn [] (dispose-repl-output-webview-panel repl-output-webview-panel))))
+    (set-webview-html! context {:webview-panel webview-panel})
+    (add-subscriptions! context {:webview-panel webview-panel})
+    webview-panel))
+
+(defn ^:export show-repl-output-webview-panel
+  [preserve-focus?]
+  (let [context {:env/is-debug (:is-debug util/env)
+                 :vscode/vscode @util/vscode
+                 :vscode/context @util/vscode-context}
```
bpringe, line 174: The pattern I followed is constructing a context in top-level functions using dereferenced values of the atoms storing the vscode API and the vscode context. This makes testing a little simpler because instead of redef'ing or resetting atoms for testing of downstream functions, we just construct a context with whatever mocks we want. This also makes downstream functions more pure.

bpringe: Also, being able to mock the vscode API and test that certain functions in it are being called correctly is pretty neat 😃.

src/cljs-lib/src/calva/repl/webview/core.cljs
```diff
@@ -0,0 +1,229 @@
...
+    (set-code-theme! context {:color-theme-kind active-code-theme-kind
+                              :webview-panel webview-panel})))
+
+(def output-category->command-name
+  {"otherOut" "show-stdout"
+   "evalOut" "show-stdout"
+   "evalResults" "show-result"
+   "evalErr" "show-stdout"
+   "otherErr" "show-stdout"
+   "clojure" "show-result"})
```
bpringe, line 188: We currently use "show-stdout" for all non-code output, but we can easily change that to make different output types look more distinct. This is just a starting point.

src/cljs-lib/src/calva/repl/webview/core.cljs
```diff
@@ -0,0 +1,229 @@
...
+
+(defn stacktrace->message
+  "Takes a clojurefied stacktrace and returns a string representation of it for printing."
+  [stacktrace]
+  (->> stacktrace
+       (filter (fn [{:keys [flags class]}]
+                 (and (not (some #{"dup"} flags))
+                      (not (contains? stacktrace-classes-to-ignore class)))))
+       (map stacktrace-entry->string)
+       (str/join "\n")))
```
bpringe, line 218: This logic was basically copied from the TS code so the stack traces look the same.

package.json
PEZ: TIL

src/cljs-lib/src/calva/repl/webview/core.cljs
PEZ: In my experience VS Code complains if I try to deref the API object. You are not running in to that? If so I should have a look at what I do differently...

deps.edn
```diff
@@ -8,6 +8,11 @@
         org.clojars.liverm0r/dartclojure {:mvn/version "0.2.23-SNAPSHOT"}
         vvvvalvalval/supdate {:mvn/version "0.2.3"}
         camel-snake-kebab/camel-snake-kebab {:mvn/version "0.4.3"}
+        no.cjohansen/replicant {:git/url "https://github.com/cjohansen/replicant.git"
+                                :git/sha "ef2fecbe301cafa8449b74a1e2cbe4fc3ddc18ac"}
```
PEZ: We should be using the latest stable Replicant, I think:

`:mvn/version "2025.03.27"`

deps.edn
```diff
@@ -8,6 +8,11 @@
...
         #_#_org.clojars.liverm0r/dartclojure {:local/root "../DartClojure"}}
+ :aliases {:dev {:extra-deps {org.clojars.abhinav/snitch {:mvn/version "0.1.14"}}}
```
PEZ, line 13: Not sure this is needed. I add it via a keykombo. But I also don't object to having it there. 😄

docs/site/output-webview.md
```diff
@@ -0,0 +1,37 @@
+---
+title: The Output Webview
+description: A read-only webview for REPL output.
```
PEZ: Not sure we need to say it is read-only?

docs/site/output-webview.md
```diff
@@ -0,0 +1,37 @@
...
+
+!!! Tip "Quickly create Rich Comments with a Calva command"
+    Use the Calva command "Add Rich Comment," which will add a `comment` form right below the form your cursor is currently at. Memorize the keyboard shortcut for this, and you can very easily create Rich Comments near the code you're working on to try out code in the REPL before adding it to your actual code.
+
+!!! Note "Don't like adding scratch code to code files?"
+    If you prefer not to leave scratch code in your code files or prefer not to have to remove it later, then create a separate scratch code file. This file can be your own personal scratch file, which you can commit, giving you a versioned history of all your scratch work, which may come in handy later.
```
PEZ: I wonder if we need to mention this at all. It is an output view. The text about how to effectively send things to the repl is nice, so we should keep it. Maybe a REPL-input page and we can link to it from here as a **See also**?

PEZ: Calva has fiddle files for this. Which should probably at least be mentioned. It allows for some workflows that a random scratch file doesn't provide.

docs/site/output.md
```diff
@@ -11,11 +11,12 @@ Calva categorizes output into three types:
...
 * **other output**: Other messages, logs, etc
 
-With the setting `calva.outputDestinations`, you can configure where each category of output should go to:
+With the setting `calva.outputDestinations`, you can configure where each category of output should go to. These are the allowed values and their descriptions:
 
-* the [REPL Window](repl-window.md)
-* the _Calva Says_ Output Channel
-* the _Calva Output_ (pseudo) Terminal
+* `"repl-window"` - The [REPL Window](repl-window.md) (an editor-based read/write output view). This is the default value.
+* `"webview"` - The [output webview](output-webview.md) (a read-only webview that is much more performant than the REPL Window)
```
PEZ: I wonder... `webview` is an implementation detail. May lead some people to think that it is external to VS Code. How about we just call it `output-window`? It will create some confusion in the code, but we can clean that up in a later commit.

docs/site/output-webview.md
PEZ, line 1: I suggest Output Window, or Output View. Avoiding to bring in the implementation detail about how it is built.

package.json
```diff
@@ -1955,6 +1963,12 @@
         "command": "calva.showResultOutputDestination",
         "title": "Show/Open the result output destination"
       },
+      {
+        "category": "Calva",
+        "command": "calva.showReplOutputWebview",
+        "title": "Show/Open the REPL output webview",
+        "enablement": "calva:connected"
```
PEZ: I have started to think we shouldn't disable commands as often as we do. It can create confusion as to why something is not there, not doing anything, and so on. And opening the output view makes total sense when the repl isn't connected.

package.json
```diff
@@ -2730,6 +2744,12 @@
         "args": false,
         "when": "calva:keybindingsEnabled && !calva:outputWindowActive"
       },
+      {
+        "command": "calva.showReplOutputWebview",
+        "key": "ctrl+alt+o w",
```
PEZ: How about we claim `ctrl+alt+o o` for this one? I guess it is now taken by the output channel, but we can change that. I don't think many people use it anyway.

repl-output-ui/css/main.css
```diff
@@ -0,0 +1,2 @@
+/* This file is linked in the repl output webview. It's not currently used, but I'm leaving it so if we want
+   to add styles in the future it's easy to do so. */
```
PEZ, line 2: 👍

src/cljs-lib/src/calva/repl/webview/core.cljs
```diff
@@ -0,0 +1,228 @@
...
+                               " (when is-debug " 'unsafe-eval' ") csp-source ";
+                    " (when is-debug "connect-src ws://localhost:9630/api/remote-relay;") "
+                    base-uri 'none';
+                    form-action 'none';\">
+
+    <title>REPL Output</title>
+
+    <link rel=\"stylesheet\" href=\"" css-href "\" />
+
+    <!-- Should these stylesheets and scripts be saved and referenced locally so that if users are offline the webview still functions as expected? -->
```
PEZ, line 50: Probably. We can add an issue about it.

src/cljs-lib/src/calva/repl/webview/core.cljs
```diff
@@ -0,0 +1,228 @@
...
+    vscode-context :vscode/context}]
+  (let [extension-uri (.. ^js vscode-context -extensionUri)]
+    (.. ^js vscode -Uri (joinPath extension-uri "repl-output-ui" "css" "main.css"))))
+
+(defn set-webview-html!
+  [context
+   {:keys [^js webview-panel]}]
+  (let [js-source (get-js-source context {:webview-panel webview-panel})
+        css-path (get-css-path context)
+        css-href (.. ^js webview-panel -webview (asWebviewUri css-path))
```
PEZ, line 106: I will have to lean the `..` syntax. 😄

src/cljs-lib/src/calva/repl/webview/core.cljs
```diff
@@ -0,0 +1,228 @@
...
+    (.. webview-panel
+        -webview
+        (postMessage (pr-str (merge
+                              {:id (str (random-uuid))} ;; Provide an id if one wasn't provided by the caller
+                              message))))))
+
+;; The connect-src and unsafe-eval are only needed in development mode for the shadow-cljs
+;; dev workflow to function properly
+(defn get-webview-html
+  [{:env/keys [is-debug]} {:keys [js-source css-href csp-source]}]
```
PEZ, line 23: A thing I have started to do is to “namespace” all keys I get away with doing it. So here, for instance, I would invent a namespace, say `:html/` for the args. It gets you rename support and often the same maps are bing sent around and you can find references and such.

Just noting. Nothing we need to address in any hurry, me thinks. (I am not even sure it is a good idea, even if I have had pretty good experience from it.)

src/evaluate.ts
PEZ, line 1: What a horrible experience, shifting from that beautifully designed ClojureScript code to this horribly mess of TypeScript!

docs/site/output-webview.md
```diff
@@ -0,0 +1,37 @@
+---
+title: The Output Webview
+description: A read-only webview for REPL output.
+---
+
+The output webview is a read-only webview for REPL output. It's more performant than [the editor-based REPL window](repl-window.md).
```
PEZ: It has a ton of other benefits too. As well as things where it loses. (Such as paredit navigation and selection, which we probably should add later).

I think we should have a small list of features here (such as rich stack traces) and link to the output destinations page for a discussion about how the different destinations compare.

docs/site/output.md
PEZ, line 1: We can have a table here, comparing the destinations. A bit like those tables comparing price plans 😄 Something to help people make the right choices.

PEZ:
I'd like:

* Labels/headings for what ouput is of which kind. This is something I get confused about in the terminal output. In the screenshot below it is the evaluated code and then the result. Since it is a data structure, they look extra much the same. Labels would help tell me what I am looking at.
* Formatted evaluated code.
* If possible, the same highlighting as I have in my editor. I am using GitHub Dark (Default) so it is the same color scheme, but it makes different decisions about what should get what color.
* Copy buttons on every block of output.

image

PEZ:
I'd like to merge this ASAP. Which I think is now, but I'll wait for your input, @bpringe. Maybe some of the naming that I suggest in my comments should be addressed first, so that it is less likely that we will change settings names and such. Also something that marks this as WIP so that people can opt in to help test it, without expecting things to work perfectly yet.

PEZ:
Output is only added to the view as long as it is opened. I want output to be there when I open it if it was my destination setting. Right now it also clears when I close it and then open it. Which is not ideal.

PEZ:
Issue: The output clears when I move it to a different editor group.

PEZ:
We should support ANSI coloring (or at least strip it, but really, would be great if we supported it.)

image

PEZ:
Sometimes syntax highlighting is missing completely:

image

deps.edn
bpringe: Ah yes, I didn't think to upgrade. This PR's been a long time in the making. 😄

deps.edn
bpringe: Ah you mean with Calva Power Tools?

docs/site/output-webview.md
bpringe: I think it makes sense to clarify that, since it's a potentially significant difference (for some people) to the editor output window.

bpringe: If we add some input mechanism to the webview itself in the future, we can update the docs.

docs/site/output-webview.md
bpringe: I vote to keep it for now. As I mentioned in another comment, it's a big difference (for some people) to the editor-based REPL output window. Considering that many users of it probably will come to it familiar with the editor-based output window, we should address a potentially frequently asked question right in the page about the webview. That's my opinion, at least. I think updating this page in the future / removing this clarification in the future might make sense.

docs/site/output.md
bpringe: I like removing "webview" from the name. I'll do that.

package.json
bpringe: Sounds good. I actually think similarly.

src/cljs-lib/src/calva/repl/webview/core.cljs
bpringe: I prefer it 🤷

bpringe:
> Sometimes syntax highlighting is missing completely:

@pez Could you create an issue and provide repro steps? We should label these issues with a new label like "output-view".

bpringe:
> We should support ANSI coloring (or at least strip it, but really, would be great if we supported it.)

@PEZ Could you also create an issue with a repro steps for this too?

src/cljs-lib/src/calva/repl/webview/core.cljs
bpringe: I have not experienced that. What's the error?

bpringe:
> Copy buttons on every block of output.

This is a todo item already in my project for this. They're all draft issues but once we release this I'll change whatever aren't completed to real issues and started created real issues onward.

> If possible, the same highlighting as I have in my editor. I am using GitHub Dark (Default) so it is the same color scheme, but it makes different decisions about what should get what color.

We discussed this a while back, and it's not easy from what I could tell. One hightlightjs theme per VS Code color theme kind is a good first pass, at least.

bpringe:
> Output is only added to the view as long as it is opened. I want output to be there when I open it if it was my destination setting. Right now it also clears when I close it and then open it. Which is not ideal.

I have some draft issues for these problems and plan to address them. They are high in priority.

image

deps.edn
PEZ: Even before Power Tools I had this in my custom repl commands array:

```json
      {
        "name": "qol: Add Snitch dependency",
        "repl": "clj",
        "snippet": "(require '[clojure.repl.deps :refer [add-libs]])\n\n(add-libs '{ org.clojars.abhinav/snitch {:mvn/version \"0.1.16\"}})\n\n(require '[snitch.core :refer [defn* defmethod* *fn *let]])"
      },
```

package.json
bpringe: It's taken by `showResultOutputDestination`, which makes sense to me. I changed this one to `ctrl+alt+o v` (v for view).

When we allow multiple output destinations, the `showResultOutputDestination` command may no longer make sense. I don't think users would want to open _all configured_ destinations, so maybe we can change that one to something else?

deps.edn
bpringe: Leaving it for now but may remove it later

PR #2482, opened 2024-04-01
Allow paredit commands to accept args - and use them for multicursor & copying when killing

riotrah:
## What has changed?

The philosophy behind this was already described in the private calva team chat, but I can repost here if desired.

- Makes `/src/paredit/extension.ts` multicursor and kill cmds accept an arg, for now a map with a single key for each respective type of command.
- The multicursor paredit commands' args looks like `{multicursor: boolean}`
  - When this args map and entry are non-nil, override the workspace or user config for `calva.paredit.multicursor` for that particular keybinding.
- The kill commands' args looks like `{copy: boolean}`
  - When this args map and entry are non-nil, override the workspace or user config for `calva.paredit.killAlsoCutsToClipboard` for that particular keybinding.

This will only affect users that utilize these args. There have been no documentation additions so if merged as-is there should be no changes for anyone, but if we like this idea, we'll want to probably mention this somewhere.

I will do some more changes here, like fortifying the static typing and CHANGELOG entry, and aforementioned documentation, and perhaps some basic tests, if we like this idea overall.

I'll create an issue if we like this idea.

Fixes #

## My Calva PR Checklist

I have:

    - [] Updated the `[Unreleased]` entry in `CHANGELOG.md`, linking the issue(s) that the PR is addressing.
- [x] Tests

### DIFF
--- a/docs/site/paredit.md
+++ b/docs/site/paredit.md
@@ -199,3 +199,18 @@ There is an ongoing effort to support simultaneous multicursor editing with Pare
 - Movement
 - Selection (except for `Select Current Form` - coming soon!)
 - Rewrap
+
+### Toggling Multicursor per command
+
+The experimental multicursor-supported commands support an optional command arg - like `copy` for the `kill*` commands [mentioned above](#command-args) - to control whether multicursor is enabled for that command. This is an alternative to, or supports binding-specific overrides for, `calva.paredit.multicursor`.
+
+For example:
+
+```json
+{
+  "key": "ctrl+k",
+  "command": "paredit.sexpRangeExpansion",
+  "when": "... your when conditions ...",
+  "args": {"multicursor": false}
+}
+```
--- a/docs/site/paredit.md
+++ b/docs/site/paredit.md
@@ -161,6 +161,59 @@ There are some context keys you can utilize to configure keyboard shortcuts with
 
 In some instances built-in command defaults are the same as Paredit's defaults, and Paredit's functionality in a particular case is less than what the default is. This is true of *Expand Selection* and *Shrink Selection* for Windows/Linux when multiple lines are selected. In this particular case adding `!editorHasMultipleSelections` to the `when` clause of the binding makes for a better workflow. The point is that when the bindings overlap and default functionality is desired peaceful integration can be achieved with the right `when` clause. This is left out of Paredit's defaults to respect user preference, and ease of maintenance.
 
+### Command Args
+
+VSCode allows you to provide arguments to configured bindings, allowing, for example, different variants or behaviors for the same command, each bound to different shortcuts. Calva takes advantage of this by offering two commands with args:
+
+#### **1. `copy` for all `kill*` commands**
+
+When specified, will control whether killed text will be copied to the clipboard.
+This is an alternative to, or supports binding-specific overrides for, `calva.paredit.killAlsoCutsToClipboard`.
+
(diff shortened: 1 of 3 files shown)


PEZ:
This is so awesome. The exact right way to go with this.

PEZ:
I like this idea. Please go ahead with issue and documentation. I think a mention about the arguments could be added as a sub-header just before we mention that strings are treated as lists: https://calva.io/paredit/#strings-are-not-lists-but-anyway

Question. The kill commands support two options in their argument (`multi-cursor`, `copy`) right?

riotrah:
@PEZ 

1. I'm working on some documentation, will push, lmk what you think! I'll push in separate commits with the necessary changes for this PR first, then some edits of the rest of the About Keyboard Shortcuts section in `src/docs/paredit.md`.
2. Nope, the `kill*` commmands support multicursor behavior nor arg, but when I get to them, their arg will still be a map, but with both keys `copy` & `multicursor`.

riotrah:
My thought was to place them in a sub section underneath `About Keyboard Shortcuts`, since these command args are only applicable for shortcuts, not any other form of invocation (eg cmd palette, menus, etc).

But if you'd like to surface them higher, I think that could make sense - I'm not particularly partial to either.

PEZ:
> My thought was to place them in a sub section underneath `About Keyboard Shortcuts`, since these command args are only applicable for shortcuts, not any other form of invocation (eg cmd palette, menus, etc).

A thing with commands is that they are API. Even if keyboard shortcuts are an obvious place where this API surfaces it doesn't stop there. They may be used Joyride scripts, or from some other extension. That said, I'm totally fine with this going to the Keyboard Shortcuts section.

That commands are API also means that we need to be careful around them. Whatever we add we may need to support forever. Dunno what this says about the `multicursor` option, come to think of it. But if you flag that in the documentation as potentially temporary, we should be fine.

riotrah:
Ah yes, good point

riotrah:
Might be worth introducing the concept with kill+copy in the section you'd suggested, and allude to that in the experimental multicursor section at the bottom. Maybe we rename to `experimentalMulticursor`?

I mean, technically, if we kept `multicursor` as-is forever, we would "just" disable multicursor for a given cmd invocation

riotrah:
@PEZ I've pushed a couple commits, feel free to review by commit. Namely, the last two:

- last: an unrelated edit of the About Keyboard Shortcuts section
- 2nd last: the relevant command args additions

docs/site/paredit.md
```diff
@@ -46,6 +46,34 @@ The Paredit commands are sorted into **Navigation**, **Selection**, and **Edit**
 
 To make the command descriptions a bit clearer, each entry is animated. When you try to figure out what is going on in the GIFs, focus on where the cursor is at the start of the animation loop.
 
+### Command Args
+
+VSCode allows you to provide arguments to configured bindings, allowing, for example, different variants or behaviors for the same command, each bound to different shortcuts. Calva takes advantage of this by offering one official command with args:
```
PEZ: There are other Calva commands that accept arguments. Also, I tend to avoid making the Calva docs teach VS Code stuff, because ... well, that's for VS Code docs to do. (I admit there are places where we should do this anyway, but anyway. 😄 ) 

```suggestion
Some Paredit commands accept arguments. You can utilize this in keybindings and from [Joyride](https://github.com/BetterThanTomorrow/joyride).
```

PEZ:
Sorry for the delay. LGTM!

PEZ:
Regarding the note about fallback for grow/shrink selection. It reminds me that I haven't checked if the multi-cursor selection commands participate in the grow/shrink. I now checked, and they don't. Anyway, could be good to look at this, as it may be important for the general architecture of the multi-cursor editing.

PR #2502, opened 2024-04-10
Don't move on the last coor for atoms

### LINKED ISSUE #2501: Throws error when debugging atoms
The problem arises when debugging a function using an atom. In this particular case the atom is declared within a let binding.
In order to reproduce the issue:

0. Clone this [repo](https://github.com/javierrweiss/calva-debugger-issues):
1. Use "Instrument top level form for debugging" command.
2. Then either use the "Step over" or "Step out" commands.
3. Finally, in this particular reproduction case it will throw the error when evaluating the swap! function.

PEZ:
## What has changed?

For atoms (or any ref) the debugger will expand the `@...` syntax sugar to `(deref ...)` giving us an extra `coor` in debug response. Now we just go to the next `coor` value when we see that we're at a sexpr starting with `@`.

* Fixes #2501

## My Calva PR Checklist

I have:

- ~~[ ] Added to or updated docs in this branch, if appropriate~~
- [x] Tests


src/debugger/util.ts
```diff
@@ -28,9 +28,7 @@ function moveTokenCursorToBreakpoint(
   const coor = [...debugResponse.coor]; // Copy the array so we do not modify the one stored in state
 
   for (let i = 0; i < coor.length; i++) {
-    while (!tokenCursor.downListSkippingMeta()) {
-      tokenCursor.next();
-    }
+    tokenCursor.downListSkippingMeta();
```
PEZ, line 31: @bpringe I don't know why this wasn't just downList, so changed it to that, because I don't actually understand the while loop. And it seldom is a good idea to move non-structurally like it does inside there. OTOH, I am thinking there is a reason, so that's why tagging you here. 😄

src/extension-test/unit/debugger/util-test.ts
```diff
@@ -17,63 +16,119 @@ function getCoordinates(text: string): (string | number)[] {
...
+      // TODO: Figure out why this test fails.
+      // It works interactively.
+      // And the coor given to us from cider-nrepl is really [3, 1]
+      xit('deref call', () => {
+        expectBreakpointToBeFound(`;; 3, 1
+        #dbg
+        (defn boom
+          []
+          @(f)|)`);
+      });
```
PEZ, line 131: I don't understand why this test fails. Do you, @jpmonettas?

jpmonettas: Sorry, no idea, I never wrote Calva tests , not sure how that pipe thing is supposed to work

PEZ: The pipe is placing the document cursor. And these tests start the text with the `coor`. When I run this interactively with the repl message log on I can see that the cursor is exactly there when the debugger shows the ref value and that it has responded with `(3, 1)` coor.

But still a bit unfair to expect you to shed light on this, since it probably is just me having misunderstood something. 😄  Will experiment a bit more with it.

PEZ: If you're curious about the test mechanics, have a look at the rest of the tests in this file and it'll probably clear up a bit.

jpmonettas: What about the test we were doing before with `@a` instead of `@(f)` ?

PEZ: Ah, thanks. That's a typo. Now fixed. It worked because I think it is interpreted as a selection and the default is that the cursor is at the end of the selection from left to right.

src/extension-test/unit/debugger/util-test.ts
```diff
@@ -17,63 +16,119 @@ function getCoordinates(text: string): (string | number)[] {
...
+    });
+
+    describe('derefs', () => {
+      it('atom', () => {
+        expectBreakpointToBeFound(`;; 3, 1
+        #dbg
+        (defn boom
+          []
+          @!a|)`);
+      });
```
PEZ, line 110: @jpmonettas , I now realize you may be referring to this test. It passes just fine.

src/debugger/util.ts
bpringe: It's been too long for me to remember the reason that was like that 😄. However, if the existing tests pass, then we can be somewhat confident this change doesn't break anything.

PR #2511, opened 2024-04-11
Generate .circleci/config.yml from config.ys

ingydotnet:
This is the start of refactoring.

config.yml changed massively here, but should never change afterwards.

The goal is to make the input YAML/YS nice and clean and generate the correct config.yaml.

Added a simple Makefile to do the generation.
It auto-installs the `ys` binary into /tmp if needed. So you don't need to worry about that.

Run from the top level directory with:

    make -C .circleci/

or:

    bb build

or:

    make -C .circleci/ clean build

if you need to force it.


.circleci/config.ys
```diff
@@ -0,0 +1,80 @@
+# This is the main YAMLScript file that expands to the desired CircleCI
+# config.yml file when we run 'make build' or 'bb buid'.
```
PEZ:
```suggestion
# config.yml file when we run 'make build' or 'bb build'.
```

PEZ:
Thanks, @ingydotnet. 🙏  This is awesome!

PR #2540, opened 2024-04-27
Add basic REPL History search

### LINKED ISSUE #2175: Provide a way to find entries in REPL history
A very common ways to use REPL history is to use recursive search (ctrl-R). If it is hard to support, how about a simpler option of opening all history entries in a generic text file where a usual search can be used?

One issue that prevents me from keeping the entries in the calva output window is that the output my programs generate often is so huge that I want to clear the output window when it is over 1,000 or 10,000 lines long.

Slack conversation: https://clojurians.slack.com/archives/CBE668G4R/p1682562207617529.

One more thing to note is that `Show Next/Previous REPL history entry` commands when used with a shortcut often get in conflict with line moving shortcuts, so for multiline entries it is not even an easy option to scroll through all the history items. However, I understand that line moving shortcuts are probably my personal preference and maybe I can disable them in calva output.

mrkam2:
## What has changed?

This implements a very basic REPL History search.

Fixes #2175

## My Calva PR Checklist

I have:

- [ ] Tests

### DIFF
--- a/src/repl-window/repl-history.ts
+++ b/src/repl-window/repl-history.ts
@@ -142,9 +142,26 @@ function showNextReplHistoryEntry(): void {
   }
 }
 
+function searchReplHistory(): void {
+  const editor = util.getActiveTextEditor();
+  const history = getHistory(getSessionType());
+  vscode.window.showInputBox({
+    prompt: 'Enter search term for REPL history',
+    validateInput: (text) => {
+      const entry = history.filter((entry) => entry.includes(text)).pop();
+      if (entry !== undefined) {
+        showReplHistoryEntry(prependNewline(entry), editor);
+        return undefined;
+      }
+      return 'No matching entry found in REPL history';
+    },
+  });
+}
+
 export {
   addToHistory,
   addToReplHistory,
+  searchReplHistory,
   showPreviousReplHistoryEntry,
   showNextReplHistoryEntry,
   resetState,
--- a/package.json
+++ b/package.json
@@ -1979,6 +1979,12 @@
         "category": "Calva"
       },
       {
+        "command": "calva.searchReplHistory",
+        "title": "Search REPL History",
+        "enablement": "calva:connected && calva:outputWindowActive && calva:replHistoryCommandsActive",
+        "category": "Calva"
+      },
+      {
         "command": "calva.showPreviousReplHistoryEntry",
         "title": "Show Previous REPL History Entry",
         "enablement": "calva:connected && calva:outputWindowActive && calva:replHistoryCommandsActive",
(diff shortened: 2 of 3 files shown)


src/repl-window/repl-history.ts
```diff
@@ -142,9 +142,26 @@ function showNextReplHistoryEntry(): void {
   }
 }
 
+function searchReplHistory(): void {
+  const editor = util.getActiveTextEditor();
+  const history = getHistory(getSessionType());
+  vscode.window.showInputBox({
```
PEZ, line 148: It may be this dangling promise that causes the linter to complain. I think the solution is to return the promise, because otherwise systems like Joyride and `runCommands` can't wait for the command to finish, which limits its usefulness.

```suggestion
  return vscode.window.showInputBox({
```

PEZ (changes requested):
Hi! Thanks for helping improve Calva! 🙏 

Before we will consider merge, you'll need to reinstall the checklist of the PR and tick items off as you have completed them. Don't hesitate to ask us for help if there's anything that is unclear or you just want help with it for whatever reason.

mrkam2:
This implementation is very basic. It is missing a few things:
- iterating over several results having the same match,
- highlighting the match.

I'm curious how you would've implemented these. ChatGPT suggests to use a custom view. Maybe that's what I'll try next. But so far I've compiled the version of the plugin to test with and want to see how much those issues would affect me.

PEZ:
> I'm curious how you would've implemented these. ChatGPT suggests to use a custom view.

It's important that this is kept simple and low maintenance. I don't think anyone in the Calva team really uses the REPL Window much and we're not keen on building it out much more. I think that eventually we may replace it by something simpler alltogether.

I think I would try with a quickpick menu first. Building it from the full repl history. That would offload the search and match to VS Code. Joyride example:

```clojure
(vscode/window.showQuickPick
 #js [#js {:label "(when-let [requires (resolve 'clojure.main/repl-requires)]\n(clojure.core/apply clojure.core/require @requires))"
           :detail "[clj] user"}
      #js {:label "(println 42)"
           :detail "[cljs] pez.pirate-lang"}])
```

This pops open a quickpick menu with the two items, showing the start of both. When the user types, any matching text will be highlighted in the menu and the menu will shrink to only the matchin items. It has limitations. Only the start of the text is shown, and if the match is outside the visible text, it will be invisible. But the menu will still shrink to only the matching items, so I think it is good enough and worth to try.

mrkam2:
> I don't think anyone in the Calva team really uses the REPL Window much

Interesting. Do you refer to all Calva users or just some core members?

PEZ:
> > I don't think anyone in the Calva team really uses the REPL Window much
> 
> Interesting. Do you refer to all Calva users or just some core members?

I'm referring to the [maintainers of Calva](https://github.com/BetterThanTomorrow/calva?tab=readme-ov-file#current-maintainers).

PR #2663, opened 2024-11-02
Updates and refactoring of YAMLScript 0.1.82

ingydotnet:
This was pinned to 0.1.56, now 0.1.82.

We take advantage of new additions to make the inputs even cleaner.

Generated `config.yml` file remains unchanged (as required for any refactoring).

### DIFF
--- a/.circleci/workflows/build-test.yaml
+++ b/.circleci/workflows/build-test.yaml
@@ -1,44 +1,30 @@
+!yamlscript/v0/
+
+requires-checkout =::
+  requires:
+  - checkout
+
+requires-build =::
+  requires:
+  - build
+
+
 jobs:
 - checkout:
     filters:
       tags:
         ignore: /^v\d+\.\d+\.\d+-?.*/
 
-- prettier-check:
-    requires:
-    - checkout
+- prettier-check:: requires-checkout
+- test-grammar::   requires-checkout
+- build::          requires-checkout
 
-- test-grammar:
-    requires:
-    - checkout
-
-- build:
-    requires:
-    - checkout
-
-- eslint-check:
-    requires:
-    - build
-
-- test-cljslib:
-    requires:
-    - build
-
-- test-integration:
-    requires:
-    - build
-
-- test-e2e:
-    requires:
-    - build
-
-- test-e2e-sub-projects:
-    requires:
-    - build
-
-- test-ts-unit:
-    requires:
-    - build
+- eslint-check::          requires-build
+- test-cljslib::          requires-build
+- test-integration::      requires-build
+- test-e2e::              requires-build
+- test-e2e-sub-projects:: requires-build
+- test-ts-unit::          requires-build
 
 - marketplace-preview-publish:
     requires:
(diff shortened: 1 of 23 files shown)


ingydotnet:
This work uncovered a bug in 0.1.81 which has been fix, but not released. 
Merging this should not be done until 0.1.82 is released. Scheduled for 2024/11/03.

ingydotnet:
YAMLScript 0.1.82 has been released, and tested to work with this PR.

Ready for review.

.circleci/workflows/build-test.yaml
```diff
@@ -1,44 +1,30 @@
...
-
-- test-ts-unit:
-    requires:
-    - build
+- eslint-check::          requires-build
+- test-cljslib::          requires-build
+- test-integration::      requires-build
+- test-e2e::              requires-build
+- test-e2e-sub-projects:: requires-build
+- test-ts-unit::          requires-build
```
PEZ: This is sweet! Just a question: Would it be possible to make the required job an argument somehow? Not saying we should do it, just wondering.

.circleci/workflows/release-publish.yaml
```diff
@@ -1,91 +1,59 @@
...
+    ignore: /.*/
+  tags:
+    only: /^v\d+\.\d+\.\d+-?.*/
+
+# XXX This might not be required to differ from default-filters
+default-filters-2 =::
+  branches:
+    ignore: /.*/
+  tags:
+    only: /^v\d+\.\d+\.\d+$/
```
PEZ: I think the names of these filters could be `prerelease-filter` and `release-filter`.

PEZ (commented):
LGTM. I only had one small comment on the naming of the tag filters.

.circleci/workflows/build-test.yaml
ingydotnet: https://gist.github.com/ingydotnet/1f8b535cc3e157a6835ec21406c0914d#file-gistfile1-txt-L32-L34
shows 2 possible ways to do it differently (both keeping definitions to single lines).

You could predefine a `requires` fn that takes an arg,
or you could predefine nothing and just use YAML flow syntax.

Might be best to just stick with YAML syntax here as nothing very interesting is happening that really requires a function.

ingydotnet: See 2082ad149

.circleci/workflows/release-publish.yaml
ingydotnet: See f694be9e2
I made comments on the prerelease regex.
I won't change the regex for this PR, since it's a refactoring and should cause no normative changes.

PEZ: Yeah, I think I may have just wanted to hint at the intention. Seems right to make it actually stop malformed release tags like you suggest. But later, of course.

The naming could probably be clearer still. Like `include-prerelease-tags` and `only-release-tags`, or something. But we can do that later as well.

PEZ:
Thanks for doing this! 🙏 ❤️

PR #2685, opened 2024-12-29
a more robust paredit deleteBackward

### LINKED ISSUE #2611: Paredit garbles while backspacing rapidly
When I backspace quickly, a document that begins valid-with-respect-to-Paredit can become invalid.

Rapidity (e.g., ordinary keyboard auto-repeat) is the key. I have not seen the problem when backspacing more slowly, such as once per second.  Furthermore, the problem seems more severe with relatively big documents on relatively feeble computers.  The more powerful the computer or the smaller the clj file, the less likely the problem.

To reproduce the problem: Open a clj or cljs file that is well-formed with respect to Paredit. Position the cursor at a point from which you can backspace a lot. Hold down the backspace key and allow it to auto-repeat. Soon, parens will start turning red or double-quotes will be mismatched.

To reproduce the problem on a pretty good machine, I started with Calva's src/cljs-lib/test/pez_rewrite_clj/paredit_test.cljs (simply because it was already large-ish), Selected-all, and Pasted numerous times until the buffer's length was about 20,000 lines. Then I positioned the cursor near the middle of the file, after a top-level form, and held down the Backspace key. On a more modest virtual machine with 8g memory and 4 cpus, a quite average clj file of a few hundred lines is enough.

pbwolf:
This is for review and comment to fix 2611, the backspace garble.

## What has changed? 

- A pattern is established for **non-async command handlers** that run faster and expose themselves less to the risk of race conditions. paredit.backspace is the guinea pig. The non-async pattern need not "take over the world" but will be helpful for commands that might be issued rapidly. deleteForward could be another candidate.
- A pattern is roughed out for **slightly-delayed, debounced reformat-as-you-type**, which is technically necessitated by the non-async command handlers. The reformatter needs a post-edit document and model, which do not exist until after the command has exited and processChanges has been apprised of the changes. Anyway, decoupling reformatting from the commands helps keep the commands fast enough to enjoy on autorepeat.

- paredit.backspace is the guinea pig for the aforementioned patterns.
- paredit.backspace is registered with **registerTextEditorCommand** (so it receives a TextEditorEdit from VS Code), and **does not await anything**. In the pareditCommands array, it has "handlerNow" instead of "handler", distinguishing the registration technique. (A suffix helps keep twins identifiable, and the particular suffix 'Now' is shorter and more pointed than 'Sync'.)
- paredit.backspace uses a new **editNow**, vs the async "edit", in EditableDocument. Such is implemented in doc-mirror too. *To avoid duplication, as a rule the predominant logic has moved to the "...Now" function and the original async function delegates to it with async dressings before or after.*
- paredit.backspace uses a new **getConfigNow** (instead of async "getConfig"). getConfig already stuffed the haul from LSP in a variable; getConfigNow uses that variable. A new TTL of a few seconds is observed to avoid flagrant staleness. 
- paredit.backspace's **ModelEdits are contrived to result in the appropriate final caret location**. For example, the subroutine backspaceOnWhitespace is changed to use delete followed by insert, instead of a single changeRange, because insert positions the cursor and replace does not. The reason for the convention is that - if both an edit and a selection were attempted - the selection apparently isn't well coordinated with the edit and can result in garble when backspace is on autorepeat. (Not every command could be implemented this way, but not every command needs to be non-async.)
- Compensating for paredit.backspace not explicitly specifying a selection to refine each edit, StringDocument infers a selection change from the ModelEdits, in mimicry of VS Code's text editor.   
- wrapPareditCommandNow, the "...Now" synchronous-handler analog of wrapPareditCommand, makes a handler with a **document-version safety guard** that compares with a version number kept in the model by **processChanges**. In case of mismatch, the handler discards the command. This is "work shedding". Without this guard, it is not uncommon for autorepeat to invoke backspace a second or third time before processChanges is apprised of the first backspace... another race condition! Consequently the 2nd and 3rd invocations would use an outdated model and produce a garbled document. (It turns out not to be productive to put a similar guard in async handlers' wrapPareditCommand, since autorepeat may invoke a second paredit command while an in-progress async paredit command, which has not yet modified the document, is parked awaiting something.) 
- paredit.backspace invokes a new **scheduleFormatAsType** (but doesn't await it!) in one case: upon removal of an empty pair of brackets.

Risks

- format.ts - the original async paths tangential to format-as-you-type were touched. Is any feature disrupted?
- paredit-test.ts - the backspace tests, when checking the cursor location after an edit, are in fact checking the doc.model's new simulation of VS Code's TextEditor behavior, since paredit.backspace can no longer override post-edit selection.
 
Fixes #2611

## My Calva PR Checklist

I have:

    - [ ] **No** Updated the `[Unreleased]` entry in `CHANGELOG.md`, linking the issue(s) that the PR is addressing.
- [x] **I think not** Figured if **anything** about the fix warrants tests on Mac/Linux/Windows/Remote/Whatever, and either tested it there if so, or mentioned it in the PR.
- [x] N/A ~Added to or updated docs in this branch, if appropriate~
- [x] Tests
  - [x] Tested the particular change (backspace)
  - [ ] **Don't know** Figured if the change might have some side effects and tested those as well.


pbwolf:
Superseding

PR #2692, opened 2025-01-04
For #2691, stabilize deleteForward by reducing asyncness

### LINKED ISSUE #2691: Paredit garbles when deleteForward is repeated rapidly
When I rapidly repeat deleteForward, a Clojure editor that begins valid-with-respect-to-Paredit can become invalid.

Rapidity is the key, and ordinary keyboard auto-repeat is sufficient. I have not seen the problem when using deleteForward sedately, such as once per second. Furthermore, the problem seems more severe with relatively big documents on relatively feeble computers. The more powerful the computer or the smaller the clj file, the less likely the problem.

To reproduce the problem: Open clojure_core.clj included with VS Code's test data. Go to the beginning of line 1725, defmacro defmulti. Hold down the delete key for a while. Soon, parens will start turning red or double-quotes will be mismatched. For example, at one point I got the following:

```
({'(
  :added "1.0"}
```
where evidently a closing paren is missing; the final closing brace is highlighted red.

The behavior is non-deterministic.

pbwolf:
This is a small addition to the backspace PR, doing the same for deleteForward.

## What has changed?

- The deleteBackward PR is included in this PR
- deleteForward is registered with registerTextEditorCommand and does not await anything
- deleteForward skips reformatting
- 'async' and 'await' are removed from tests of deleteForward
- EditableDocument.delete, a disused subroutine of deleteForward, is removed

The reason to skipFormat is: formatting moves the selection past commas, which in combination with deleteForward makes commas resist deletion! This interplay is newly exposed by the non-async deleteForward. The (prior) async deleteForward async'ly moved the selection, which "undid" formatting's selection if the timing worked out right.

Fixes #2691 and #2611

## My Calva PR Checklist

I have:

- [x] N/A ~Added to or updated docs in this branch, if appropriate~
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Changes to Calva.
 
 ## [Unreleased]
 
+- Fix: [Paredit garbles while backspacing rapidly](https://github.com/BetterThanTomorrow/calva/issues/2611)
+
 ## [2.0.482] - 2024-12-03
 
 - Fix: [Added 'replace-refer-all-with-alias' & 'replace-refer-all-with-refer' actions to calva.](https://github.com/BetterThanTomorrow/calva/issues/2667) 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Changes to Calva.
 ## [Unreleased]
 
 - Fix: [Paredit garbles while backspacing rapidly](https://github.com/BetterThanTomorrow/calva/issues/2611)
+- Fix: [Paredit garbles when deleteForward is repeated rapidly](https://github.com/BetterThanTomorrow/calva/issues/2691)
 
 ## [2.0.482] - 2024-12-03
 
--- a/src/calva-fmt/src/format.ts
+++ b/src/calva-fmt/src/format.ts
@@ -49,10 +49,10 @@ export async function indentPosition(position: vscode.Position, document: vscode
   }
 }
 
-export async function formatRangeEdits(
+export function formatRangeEdits(
   document: vscode.TextDocument,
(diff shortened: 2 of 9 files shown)


PEZ:
Thanks so much for contributing this work, @pbwolf! And thanks for researching throughly and detailing the rationale for the changes. It all makes sense, I think. The only thing I wonder about is why we call it **Now** and **non-async**, instead of just **sync**, as is the pattern in JS land. But we can change that later.

It looks ready to merge, but the PRs are both marked **Draft**. Is there something you still want to change, @pbwolf?

pbwolf:
I restored the reformat-as-you-type in deleteForward for symmetry with deleteBackward: upon removing an empty list, it reformats. This can cause a strange cursor move, but I am not sure whether the answer is for deleteForward not to reformat or for reformatting not to move the cursor. There is a logic branch in the reformatting devoted specifically to moving the cursor if reformatting did not change the document. For example: `|(),,,()`, del, del, and suddenly the cursor scoots over the commas: `,,,|()`

skylize:
~~Maybe I'm alone here? I think this should be viewed as a workaround, rather than a fix. We should hope at some point someone might be able to come back later and make things work both in-order and  async.

Please don't let that come across as any sort of disagreement with going through on the patch. A successful workaround is still successful.~~

Thanks for the work.

Edit:

> A pattern is established for non-async command handlers that run faster and expose themselves less to the risk of race conditions. This will be helpful for commands that might be issued rapidly, such as backspace

I might take that back? Seems I was talking without enough knowledge on the topic, so can't say for sure. In whatever, case thanks again for the hard work.

PR #2709, opened 2025-01-25
2708 cursor context ns

### LINKED ISSUE #2708: cursor-context calva:ns is less accurate than the *ns* used by calva.evaluateSelection
Cursor context `calva:ns`, for key/command bindings, is described as "A string with the current namespace" at https://calva.io/when-clauses/. 

In fact, `calva:ns` is the _last_ outermost-level `(ns...)` or `(in-ns...)` form in the document, regardless of `#_`; and "user" if the document does not contain a top-level `ns` or `in-ns`. 

This is unexpected because it is neither 
- "current namespace", the namespace of the current document: 999% of .clj files have one `ns`, at the top, by classloader convention; nor 
- "current namespace", what you get by evaluating `*ns*` at each point in the document

```
#_(ns a)
(ns b)
(ns c)
#_(ns d)
(comment
 (in-ns 'e) 
 (do
  (ns f)
 )
)
```

Additional consideration: Computing `calva:ns` is in a hot loop, unlike eval, which is on-demand and has the leisure to spend 20ms figuring out what namespace to use.

pbwolf:
Search-backward for ns by starting from the cursor position (not the end of the document) for consistency with evaluateSelection

## What has changed?

- In when-contexts.ts, find current ns using the same technique as in evaluate.ts; i.e., instead of calling getDocumentNamespace, which is a wrapper around getNamespace that compels it to search from the end of the document, call getNamespace directly and start the search at the cursor position.

Folks should perceive a change only if they have customized key bindings to make use of calva:ns in a "when" clause.
No change is necessary to the documentation.

I do not know how to test key maps.

There remain some uses of getDocumentNamespace, the function this PR bypassed instead of changing. Do they really want the document's last declared namespace, as opposed to the first one? In any case, they are outside the scope of #2708.

- src/api/document.ts (getNamespace)
- src/evaluate.ts (requireREPLUtilitiesCommand)
- src/fiddle-files.ts (evaluateFiddleForSourceFile)

Fixes #2708

## My Calva PR Checklist

I have:

- [-] Tests
  - [-] Tested the particular change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Changes to Calva.
 ## [Unreleased]
 
 - Fix: [Have test runner show diff between "actual" and "expected"](https://github.com/BetterThanTomorrow/calva/issues/1007)
+- Fix: [cursor-context calva:ns is less accurate than the *ns* used by calva.evaluateSelection](https://github.com/BetterThanTomorrow/calva/issues/2708)
 
 ## [2.0.483] - 2025-01-08
 
--- a/src/when-contexts.ts
+++ b/src/when-contexts.ts
@@ -21,7 +21,7 @@ export function setCursorContextIfChanged(editor: vscode.TextEditor) {
   }
   const contexts = determineCursorContexts(editor.document, editor.selections[0].active);
   setCursorContexts(contexts);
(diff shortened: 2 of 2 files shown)


PEZ:
Thanks! 🙏

PR #2727, opened 2025-02-16
2726 rewrap imbalance

### LINKED ISSUE #2726: Rewrapping to or from a Set introduces imbalance
Starting with
```
a #{b c|} d
```
Calva Paredit: Rewrap []

Expected:
```
a [b c|] d
```
Actual:
```
a [b ]|} d
```

And also going the other way, starting with
```
[b c|] d
```
Calva Paredit: Rewrap #{}

Expected:
```
#{b c|} d
```
Actual:
```
#{b c|]}d
```

observed in Calva 2.0.485, VS Code 1.97.0

pbwolf:
## What has changed?

- rewrap has been corrected, to position the brackets accurately even when the opening bracket changes in size. This was done by _removing_ code that compensated for the size change when composing a ModelEdit plan. VS Code apparently wants all positions in a transaction of edits to be specified with reference to the unchanged document; as though it sorted edits by position and executed them from end-to-start of the document. To get the same result from the tests' TextEditorEdit emulator in LineInputModel, rewrap sorts its edits back-to-front.* 

- rewrap has been improved, not to override VS Code's default adjustment to (multi-) selections following the edit transaction. There was a race condition when rewrap calculated selections before arranging the edit transaction and asynchronously reasserted them later. A generalization of the selection-adjustment-inference code has been moved to LineInputModel, which has a role in emulating VS Code's TextEditorEdit for unit tests.  

- rewrap still triggers reformat-as-you-type. It seems to actually reformat at only 1 cursor, but that issue is beyond the scope of rewrap.

- No changes to docs or tests. The tests were already well-specified, - they passed erroneously due to complementary quirks of LineInputModel and rewrap.
 
*Perhaps, instead of rewrap sorting its edits back-to-front so LineInputModel will come up with the same result as TextEditorEdit, the sorting should be done by LineInputModel, for all commands (even though, to date, only rewrap is known to have run into trouble)?

Fixes #2726 

## My Calva PR Checklist

I have:

- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Changes to Calva.
 ## [Unreleased]
 
 - Fix: [Fix description of calva.autoOpenResultOutputDestination setting](https://github.com/BetterThanTomorrow/calva/pull/2721)
+- Fix: [Rewrapping to or from a Set introduces imbalance](https://github.com/BetterThanTomorrow/calva/issues/2726)
 
 ## [2.0.485] - 2025-01-27
 
--- a/src/cursor-doc/model.ts
+++ b/src/cursor-doc/model.ts
@@ -587,10 +587,12 @@ export class LineInputModel implements EditableModel {
   edit(edits: ModelEdit<ModelEditFunction>[], options: ModelEditOptions): Thenable<boolean> {
     return new Promise((resolve, reject) => {
       this.editTextNow(edits, options);
-      if (this.document && options.selections) {
-        this.document.selections = options.selections;
-      } else {
-        this.document.selections = selectionsAfterEdits(edits, this.document.selections);
+      if (this.document) {
+        if (options.selections) {
+          this.document.selections = options.selections;
+        } else {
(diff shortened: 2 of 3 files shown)


PEZ:
Sweet! 

> * rewrap skips format-as-you-type, because rewrapping is formatting-neutral.

If you rewrap something like:

```clojure
[1
 2]
```

With `#{}`, then you'd expect it to reindent, right?

> *Perhaps, instead of rewrap sorting its edits back-to-front so LineInputModel will come up with the same result as TextEditorEdit, the sorting should be done by LineInputModel, for all commands (even though, to date, only rewrap is known to have run into trouble)?

I think this makes great sense. Though we should probably merge this PR first. (I don't think you suggested making the bigger change in this PR, just noting.)

pbwolf:
> If you rewrap something like:
> 
>     [1
>      2]
> 
> With #{}, then you'd expect it to reindent, right?

All right. Removed skip-format. 
The reformatting is apparently incomplete after a _multi-cursor_ edit, but that matter could be addressed more globally, not in rewrap specifically.

Issue #2737, opened 2025-03-01
splice-kill-forward or splice-kill-backward, then undo, undoes too much


pbwolf:
1. Initial
```([fo| bar])```
2. Type letter o
```([foo| bar])```
3. Splice Kill Forward
```(foo|)```
4. Undo once
```([fo| bar])``` - reverted to No.1; expected No.2

- - - - -

1. Initial
```([fo| bar])```
2. Type letter o
```([foo| bar])```
3. Splice Kill Backward
```(| bar)```
4. Undo once
```([fo| bar])``` - reverted to No.1; expected No.2

PR #2741, opened 2025-03-02
Multicursor reformatting

### LINKED ISSUE #2738: Reformat all edited locations after multicursor structural edit
After a multicursor structural edit, the automatic reformat touches only one of the cursors, and loses the other cursors.

Initial
```
[(foo
   |bar)
 (foo
   |bar)]
```
Rewrap {}
Actual
```
[{foo
  |bar}
 {foo
   bar}]
```
Expected:
```
[{foo
  |bar}
 {foo
  |bar}]
```

pbwolf:
## What has changed?

- After a multicursor command, all cursors are preserved and all affected places are reformatted.
- Consequently, Calva's user may issue one multicursor command after another, incrementally affecting the same bunch of cursors.

All commands that manipulate the document via DocumentModel are potentially affected, even commands that are not multicursor-capable. The single-cursor reformatting code path is just the base case of the multi-cursor code path, so it should be regarded as all-new despite the same underlying clj-format.

Extending auto-reformatting to multiple cursors changed the calculation of the region(s) to reformat. It may be that the region that gets reformatted after some commands will be benignly different from before, e.g., more extensive.

I did not add reformatting to the multicursor section of the paredit documentation, since it is not controlled by a setting.

Details:

- Formatting preserves multiple cursors by inserting and removing individual runs of whitespace, instead of replacing a monolithic Clojure code block with a formatted version of the same. VS Code automatically adjusts each cursor as spaces are inserted or removed above it or to its left.
- The user's cursors no longer relate directly to reformatting. 
  1) One structural edit might be indicated by multiple cursors, perhaps arrayed at various points in the same list.
  2) Structural edits' blast radius ("formatDepth") may depend on the surroundings and one structural edit may differ from another in the same simultaneous bunch.
- Regions to reformat are computed from the array of ModelEdits instead of from one cursor position and a formatDepth option. Specifically: DocumentModel.edit finds the bounds of the list enclosing each ModelEdit and translates those bounds to offsets in the expected post-edit document. (The heuristics are those used by Calva's unit test suite, which also tries to predict the net result of a bunch of ModelEdits.) After the edit transaction, DocumentModel.edit formats whatever forms are at those positions. "formatDepth" did not scale to multiple cursors and has been removed.
- Reformatting is skipped if the document version number is greater than expected, a clue that VS Code interleaved two async commands.

Fixes #2738, #2737

## My Calva PR Checklist

I have:

- [x] Tests


src/calva-fmt/src/format.ts
```diff
@@ -203,66 +238,160 @@ function _calculateFormatRange(
...
+  start: number;
+  end: number;
+  text: string;
+};
+
+/** whitespace and substance. Pre-format and re-formatted text can be expressed
+ * as a series of SpacedUnit. The substance parts can be aligned and then
+ * changes in size can be translated to TextEdits.
+ */
+type SpacedUnit = [spaces: string, stuff: string];
```
PEZ: This is pretty awesome stuff! Looks lik some of it could be extracted into a module that doesn't need vscode and thus we could unit have unit tests?

PEZ:
Not sure if there's something else but the conflicts to fix here? I have been running with this branch for a few days and it is awesome. Don't think I've noticed anything that doesn't work.

pbwolf:
> This is pretty awesome stuff! Looks like some of it could be extracted into a module that doesn't need vscode and thus we could unit have unit tests?

> Not sure if there's something else but the conflicts to fix here? I have been running with this branch for a few days and it is awesome. Don't think I've noticed anything that doesn't work.

Thank you for the notes.  I aim to merge dev's recent advances and extract some unit-testable parts

pbwolf:
Procedures that calculate whitespace adjustments are extracted to a new module and demonstrated with tests that check the text and cursor positions in the results. The cursor-position tests motivated some changes to the model of VS Code's implicit cursor adjustment after edits above or to the left of the cursor. The implicit behavior is described in passing on the VS Code Extension API reference page (https://code.visualstudio.com/api/references/vscode-api#TextEditorEdit) as "Although the equivalent text [insertion] can be made with replace, insert will produce a different resulting selection (it will get moved)."

PEZ:
Thanks, @pbwolf! This is such a QOL change for me. You don't mention it in the PR, but also the format-current-form command is now multi-cursor.

Issue #2789, opened 2025-04-18
Format on save + cljfmt fails on 2.0.499 while working on 2.0.498


artem-barmin:
After update to recent version of Calva file stopped formatting on save. Enabled "verbose" level tracing for clojure-lsp intraction I found that command "textDocument/didSave" returns unformatted document back(maybe this will help, but looks like problem with lsp server configuration). 

On version 2.0.498 everything works as expected and "textDocument/didSave" return formatted file back

```
Calva version used: v2.0.499
clojure-lsp version used: 2025.03.27-20.21.36
clj-kondo version used: 2025.02.21-SNAPSHOT
Clojure-lsp server info:
{
  "log-path": "/tmp/clojure-lsp.4456348992111000772.out",
  "project-settings": {
    "clean": {
      "ns-inner-blocks-indentation": "next-line",
      "sort": {
        "ns": true,
        "require": true,
        "import": true,
        "refer": true
      }
    },
    "keep-parens-when-threading?": true
  },
  "classpath": [],
  "analysis-summary": {
    "internal": {
      "keyword-usages": 4315,
      "instance-invocations": 63,
      "protocol-impls": 25,
      "local-usages": 3106,
      "namespace-alias": 107,
      "namespace-usages": 162,
      "symbols": 2,
      "var-usages": 5224,
      "namespace-definitions": 41,
      "java-class-usages": 7,
      "locals": 1893,
      "var-definitions": 455
    },
    "external": {
      "java-class-definitions": 8778,
      "keyword-definitions": 918,
      "instance-invocations": 11,
      "protocol-impls": 6251,
      "local-usages": 187,
      "namespace-alias": 6841,
      "namespace-usages": 8643,
      "java-member-definitions": 84914,
      "var-usages": 442,
      "namespace-definitions": 2491,
      "java-class-usages": 13,
      "locals": 108,
      "var-definitions": 33675
    }
  },
  "client-settings": {
    "dependency-scheme": "jar",
    "auto-add-ns-to-new-files?": true,
    "document-formatting?": false,
    "document-range-formatting?": false,
    "text-document-sync-kind": null,
    "source-paths": null,
    "source-aliases": null,
    "cljfmt-config-path": ".cljfmt.edn"
  },
  "clj-kondo-version": "2025.02.21-SNAPSHOT",
  "server-version": "2025.03.27-20.21.36",
  "port": "NREPL only available on :debug profile (`bb debug-cli`)",
  "final-settings": {
[...]

PR #2794, opened 2025-04-24
multicursor format kluge for sort-ns-references

pbwolf:
## What has changed?

cljfmt sorting of namespace references is restored, but the cursor may wind up in an unexpected place if it does so.

The issue title was "Format on save + cljfmt fails..." but the particular problem was that multicursor-capable formatting relied on cljfmt to change only whitespace.

The multicursor-capable formatter used "console.error" when that condition was violated, but the messages may not have been noticed.

This PR makes the multicursor-capable formatter fall back to replacing the whole range if it cannot calculate micro-edits of whitespace. Whole-range replacement will be unkind to cursors that were within the range.

Edits 2025-04-27: The PR is revised to refine the kluge slightly: using the same technique as for cursor-contexts, it locates a `(ns...)` form, and if there is one, then it formats that form separately from the remainder of the document. In that way, each of the two parts produces efficient spaces-only edits that allow cursors to float, or degenerates to making a whole replacement that may shove cursors to the end; but an `(ns...)` form that gets sorted does not prevent cursors after that form from floating.

A test case is added for the "respacer" module for the case in which it stuffs the remainder of the substantively-differing document into a single edit.
 
The CHANGELOG wishfully renames the issue for clarity, saying, (Fixes) "Reformat may do nothing if cljfmt's :sort-ns-references? is true".

Addresses #2789

## My Calva PR Checklist

I have:

- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Changes to Calva.
 
 ## [Unreleased]
 
+- Fix: [Reformat may do nothing if cljfmt's :sort-ns-references? is true](https://github.com/BetterThanTomorrow/calva/issues/2789)
+
 ## [2.0.502] - 2025-04-26
 
 - Fix: [formatCode throws an error when output output is printed in the output window](https://github.com/BetterThanTomorrow/calva/issues/2791)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,6 @@ Changes to Calva.
 ## [2.0.503] - 2025-04-27
 
 - Fix: [Release 2.0.502 repl connection fails, breaking Jack-In](https://github.com/BetterThanTomorrow/calva/issues/2799) (Hopefully)
->>>>>>> dev
 
 ## [2.0.502] - 2025-04-26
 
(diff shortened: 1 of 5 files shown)


PEZ:
> Perhaps, when Calva invokes cljfmt, it should trample `:sort-ns-references?` to false?

It should be considered. This is a tricky one. Depends on if we think we can fix the underlying problem. (Which seems hard?). If the situation is temporary displacing cursors may be an annoyance we can live with. If it is permanent, we may be at a point where we need to decide not to sopport sorting from the formatter?

cc: @bpringe

CHANGELOG.md
```diff
@@ -4,9 +4,12 @@ Changes to Calva.
 
 ## [Unreleased]
 
+- Fix: [Reformat may do nothing if cljfmt's :sort-ns-references? is true](https://github.com/BetterThanTomorrow/calva/issues/2789)
+
 ## [2.0.503] - 2025-04-27
 
 - Fix: [Release 2.0.502 repl connection fails, breaking Jack-In](https://github.com/BetterThanTomorrow/calva/issues/2799) (Hopefully)
+>>>>>>> dev
```
```suggestion
```

bpringe:
What's the state of this now? Do we only not support ns form sorting in certain situations, or do we never support it?

My opinion is that we should try to support all cljfmt options in all situations, even if that causes some other (at least tolerable) issue. If we decide not to support sorting in certain situations, then I think maybe informing the user in docs would be good and also maybe informing them with an editor popup when the non-supporting case occurs during formatting that allows them to select to not show it again.

pbwolf:
> What's the state of this now? Do we only not support ns form sorting in certain situations, or ...?

Calva supports cljfmt's namespace sorting.

PR #2868, opened 2025-06-22
Reload content of output view when VS Code is reloaded

### LINKED ISSUE #2827: Reload contents of output view when VS Code window is reloaded
Use a webview panel serializer: https://code.visualstudio.com/api/extension-guides/webview#serialization

Seems we need to use setState to store state in order for the webview panel serializer to work. But keep using `retainContextWhenHidden` because this may be the only way to send messages to a hidden webview.

bpringe:
## What has changed?

- The contents of the output view are reloaded when VS Code is reloaded. (Note: this is different than closing and reopening the output view without closing / reloading VS Code. Contents are not yet preserved in that case.)
- The scroll position is preserved between VS Code reloads
- The `getState` and `setState` API is used, which will also provides a foundation for persisting contents between closing and reopening the output view without reloading VS Code. This work will be done in another PR.

Fixes #2827

## Todo

- [x] Changed theme should be remembered between closing and reopening
- [x] Scroll to previous scroll position after VS Code is reloaded 
  - Save scrollLeft and scrollHeight on scroll event - maybe throttle
- [x] Make sure copy button is working after a reload of the window + output view - I saw it not changing to "copied" when clicked one time
   - We need to call highlightElement or the function from highlightjs that highlights all elements after the reload of the output view. The highlightjs-copy plugin provides an `"after:highlightElement"` function that highlightjs will call after highlighting an element.
   - **It seems we need to remove the copy buttons before re-highlighting. Just re-highlighting is not working.** This worked when I manually deleted the div with the class `hljs-copy-container` and then ran highlightAll.
- [x] Look into error seen in console after reloading VS Code and the output view reloading
   - `Uncaught TypeError: Cannot read properties of undefined (reading '__vscode_post_message__')`
   - Not sure if this is something we're doing wrong. I don't see anything obvious and things work as they should. I'll leave it for now.
- [x] Make sure existing tests are passing
- [x] Add new tests

## My Calva PR Checklist

I have:

- [x] Tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,8 @@
     "workspaceContains:**/project.clj",
     "workspaceContains:**/shadow-cljs.edn",
     "workspaceContains:**/deps.edn",
-    "onDebugResolve:type"
+    "onDebugResolve:type",
+    "onWebviewPanel:calva.output-view"
   ],
   "main": "./out/extension",
   "capabilities": {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Changes to Calva.
 
 ## [Unreleased]
 
+- [Reload contents of output view when VS Code window is reloaded](https://github.com/BetterThanTomorrow/calva/issues/2827)
+
 ## [2.0.519] - 2025-06-09
 
 - [Support `before` and `after` functions for refresh namespaces commands](https://github.com/BetterThanTomorrow/calva/issues/2862)
--- a/shadow-cljs.edn
+++ b/shadow-cljs.edn
@@ -28,7 +28,8 @@
                              :showReplOutputWebviewPanel calva.repl.webview.core/show-repl-output-webview-panel
                              :appendToReplOutputWebview calva.repl.webview.core/append
                              :appendStackTraceToReplOutputWebview calva.repl.webview.core/append-stacktrace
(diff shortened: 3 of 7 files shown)


package.json
```diff
@@ -61,7 +61,8 @@
     "workspaceContains:**/project.clj",
     "workspaceContains:**/shadow-cljs.edn",
     "workspaceContains:**/deps.edn",
-    "onDebugResolve:type"
+    "onDebugResolve:type",
+    "onWebviewPanel:calva.output-view"
```
PEZ, line 65: Why do we need this? Not saying we don't, just wondering why =)

bpringe: See this info: https://code.visualstudio.com/api/extension-guides/webview#serialization

My understanding is that it ensures that Calva is activated if the output-view webview panel is restored upon launching VS Code, which I think is what we want. We wouldn't want the webview panel to be restored but Calva not be activated.

PEZ: Cool. TIL!

bpringe: I actually don't know that we _need_ it. I don't know of a case in which the output view webview panel would be reloaded on VS Code startup but Calva not activated via some other reason. It seems the webview panel is only reloaded if the same folder is opened (or something like that), in which case Calva will activate because it's a Clojure / ClojureScript project.

bpringe: But... I think it doesn't hurt to leave it.

PR #2960, opened 2025-10-24
feat: uses latest version of depencies for jack-in

### LINKED ISSUE #2959: Use latest version of libs to inject in jack-in if no specific version was set by the user
## Problem
As discussed in the Calva channel at Clojurians ([thread](https://clojurians.slack.com/archives/CBE668G4R/p1760987101382249)), would be good to Calva use latest version of libraries so it does not need to rely on:
 - Bump from time to time the [dependencies default version](https://github.com/BetterThanTomorrow/calva/blob/7974d596a538a803cabdb6715d7b717a11fefca7/package.json#L493-L495)
 - Users need to set a custom version while wait for the bump

## Solution
1. Create a Calva command that fetches the latest versions and stores them in extension global storage.
  1.1  `clojure -X:deps find-versions`, but using node to shell out (there is prior art)
  1.2 falls back on something hardcoded in Calva
2. We also run this version fetching as part of Calva start
3. The current settings are retrofitted (same names) to override whatever is in extension global storage
4. Jack-in uses the settings first, otherwise global storage, otherwise hardcoded versions

> include the current versions in the changelog so that it is easy for someone who’s not ready to use latest to go back to what we have today

afucher:
## What has changed?

- Jack-in now auto-fetches the latest `nREPL`, `cider-nrepl`, and `piggieback` versions on first startup (via clojure/deps; falls back to deps.clj), caches them in global state, and falls back to built-in defaults if fetching fails.
  - The fall back to deps.clj is because Calva already have deps.clj.jar, but can be removed. I added because I thought that was a nice quickwin here.
- User overrides remain the highest priority; effective resolution is: overrides → cached latest → built-in defaults.
- New module `src/nrepl/jack-in-dependency-versions.ts` implements fetching, EDN parsing, persistence, and a single-flight refresh to avoid duplicate requests.
- Added `jackInVersionFor` helper in `project-types` and removed direct dependency on raw config for version lookup.
- Activation now triggers a background refresh of missing versions after `deps.clj` is downloaded; this is non-blocking for startup.
- Built-in default versions (used only if needed): nrepl 1.3.1, cider-nrepl 0.55.4, piggieback 0.6.0. (can be defined to be the latest before merge this PR)
- Backwards compatible: existing pinned versions and workflows continue to work unchanged.

Fixes #2959 

## My Calva PR Checklist

I have:

- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,12 @@ Changes to Calva.
 
 ## [Unreleased]
 
-- Bump default `nREPL`, `cider-nrepl`, and `cider/piggieback` Jack-in dependency to `1.5.1`, `0.58.0`, and `0.6.1`
-- [Get latest version of libs to inject in jack-in if no specific version was set by the user](https://github.com/BetterThanTomorrow/calva/issues/2959)
+- [Use latest version of libs for Jack-in, if no specific version was set by the user](https://github.com/BetterThanTomorrow/calva/issues/2959)
+
+  Previous default versions:
+  - `nrepl`: 1.3.1
+  - `cider-nrepl`: 0.55.4
+  - `cider/piggieback`: 0.6.0
 
 ## [2.0.538] - 2025-10-19
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ Changes to Calva.
 
 ## [Unreleased]
 
+- Bump default `nREPL`, `cider-nrepl`, and `cider/piggieback` Jack-in dependency to `1.5.1`, `0.58.0`, and `0.6.1`
+- [Get latest version of libs to inject in jack-in if no specific version was set by the user](https://github.com/BetterThanTomorrow/calva/issues/2959)
+
 ## [2.0.538] - 2025-10-19
 
 - Fix: [Semantic tokens fail on some input with ignored forms](https://github.com/BetterThanTomorrow/calva/issues/2956)
--- a/package.json
+++ b/package.json
@@ -490,10 +490,10 @@
                 "description": "Piggieback is used to create nREPL sessions in ClojureScript projects (not with shadow-cljs though, which provides its own middleware for this."
               }
             },
-            "default": {
-              "nrepl": "1.3.1",
-              "cider-nrepl": "0.55.4",
-              "cider/piggieback": "0.6.0"
+            "default": {  
+              "nrepl": "1.5.1",
+              "cider-nrepl": "0.58.0",
+              "cider/piggieback": "0.6.1"
             }
           },
           "calva.clojureLspVersion": {
--- a/package.json
+++ b/package.json
@@ -475,7 +475,7 @@
           },
           "calva.jackInDependencyVersions": {
             "type": "object",
-            "description": "Versions of the dependencies injected by Calva Jack-in",
+            "description": "Versions of the dependencies injected by Calva Jack-in. Calva caches the latest releases on first startup and uses them unless you override them here. If fetching fails, the built-in defaults below are used.",
             "properties": {
               "nrepl": {
                 "type": "string",
--- a/package.json
+++ b/package.json
@@ -475,7 +475,7 @@
           },
           "calva.jackInDependencyVersions": {
             "type": "object",
-            "description": "Versions of the dependencies injected by Calva Jack-in. Calva caches the latest releases on first startup and uses them unless you override them here. If fetching fails, the built-in defaults below are used.",
+            "description": "By default Calva Jack-in will inject the latest `nrepl`, etcetera versions when starting the repl. Override this default behaviour by declaring which versions to use here.",
             "properties": {
               "nrepl": {
                 "type": "string",
--- a/package.json
+++ b/package.json
@@ -490,7 +490,7 @@
                 "description": "Piggieback is used to create nREPL sessions in ClojureScript projects (not with shadow-cljs though, which provides its own middleware for this."
               }
             },
-            "default": {  
+            "default": {
               "nrepl": "1.5.1",
               "cider-nrepl": "0.58.0",
               "cider/piggieback": "0.6.1"
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -43,6 +43,7 @@ import * as converters from './converters';
 import * as joyride from './joyride';
 import * as api from './api/index';
 import * as depsClj from './nrepl/deps-clj';
+import { refreshJackInDependencyVersions } from './nrepl/jack-in-dependency-versions';
 import * as clojureDocs from './clojuredocs';
 import { capitalize } from './utilities';
 import * as overrides from './overrides';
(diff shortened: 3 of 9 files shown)


src/extension.ts
```diff
@@ -186,7 +187,11 @@ async function activate(context: vscode.ExtensionContext) {
     );
   }
 
-  void depsClj.downloadDepsClj(context.extensionPath);
+  void depsClj
+    .downloadDepsClj(context.extensionPath)
+    .finally(() => {
+      void refreshJackInDependencyVersions();
```
afucher: Added here the code to get latest versions, because of the fallback to use deps.clj

src/nrepl/project-types.ts
```diff
@@ -236,9 +240,11 @@ export enum JackInDependency {
   'cider/piggieback' = 'cider/piggieback',
 }
 
-const NREPL_VERSION = () => getConfig().jackInDependencyVersions['nrepl'],
-  CIDER_NREPL_VERSION = () => getConfig().jackInDependencyVersions['cider-nrepl'],
-  PIGGIEBACK_VERSION = () => getConfig().jackInDependencyVersions['cider/piggieback'];
+const jackInVersionFor = (key: JackInDependencyKey) => getEffectiveJackInDependencyVersions()[key];
```
afucher, line 243: avoid get all config, and use directly the function that return the versions

src/nrepl/jack-in-dependency-versions.ts
```diff
@@ -0,0 +1,225 @@
...
+      errors.push(`deps.clj output missing version for ${library}`);
+    } catch (error) {
+      errors.push(`deps.clj failed: ${(error as Error).message}`);
+    }
+  } else {
+    errors.push('deps.clj.jar not available');
+  }
+
+  throw new Error(errors.join(' | '));
+}
```
afucher, line 105: Since Calva already download deps.clj jar, added this fallback to try to get the versions if the first approach fails.
I don't know if is needed to be honest, adds more complexity and I can't argue if when it fails by any reason using clojure cli, if this will be a good fallback.

src/nrepl/jack-in-dependency-versions.ts
```diff
@@ -0,0 +1,225 @@
...
+};
+
+const JACK_IN_DEPENDENCY_KEYS = Object.keys(
+  JACK_IN_DEPENDENCY_LIBRARIES
+) as JackInDependencyKey[];
+
+const JACK_IN_DEPENDENCY_DEFAULTS: Record<JackInDependencyKey, string> = {
+  'nrepl': '1.3.1',
+  'cider-nrepl': '0.55.4',
+  'cider/piggieback': '0.6.0',
```
afucher: These versions can be bumped to the latest (or any other specific version), before we merge the PR

PEZ: Which are the latest now? I think we should use latest. And in the changelog we can put the versions used in Calva currently, so that someone who wants to configure those can easily find them.

PEZ:
Thanks! This is so wonderful. And so wonderfully summarized. I will look closer tomorrow, late here now. =)

PEZ:
The build is failing because of formatting btw. You can install the Prettier extension and it will run as a watcher if you are using the Build task. Then code will always be formatted on save.

src/nrepl/jack-in-dependency-versions.ts
PEZ: I think this is great. So if someone is using Leiningen and do not have `clojure` they will still be running with latest jack-in dependencies.

package.json
```diff
@@ -475,7 +475,7 @@
           },
           "calva.jackInDependencyVersions": {
             "type": "object",
-            "description": "Versions of the dependencies injected by Calva Jack-in",
+            "description": "Versions of the dependencies injected by Calva Jack-in. Calva caches the latest releases on first startup and uses them unless you override them here. If fetching fails, the built-in defaults below are used.",
```
PEZ:
```suggestion
            "description": "By default Calva Jack-in will inject the latest `nrepl`, etcetera versions when starting the repl. Override this default behaviour by declaring which versions to use here.",
```

PEZ: I'm guessing the user will need to specify all, if specifying one? If so we should probably say that.

src/nrepl/jack-in-dependency-versions.ts
PEZ: Do we need the defaults here? Maybe we can keep specifying them in the manifest (package.json). I'm not sure we can, but it may be possible to distinguish between defaults and user provided settings.

package.json
afucher: It works if the user just specify one.

src/nrepl/jack-in-dependency-versions.ts
afucher: @PEZ the latest are:
```
cider-nrepl = '0.58.0'
cider/piggieback = '0.6.1'
nrepl = '1.5.1'
```

> in the changelog we can put the versions used in Calva currently  

So, do you want to bump the default versions to the latest and add to the changelog which versions? The ones that were before the bump? or the latest?

afucher: Moved to the package.json

package.json
afucher: Also, updated the docs: https://github.com/BetterThanTomorrow/calva/pull/2960/files#diff-6130791168367a489ce43eaa7c6a6c269300ff5bd20b85045f433aae9abc25b4

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ Changes to Calva.
 
 ## [Unreleased]
 
+- Bump default `nREPL`, `cider-nrepl`, and `cider/piggieback` Jack-in dependency to `1.5.1`, `0.58.0`, and `0.6.1`
+- [Get latest version of libs to inject in jack-in if no specific version was set by the user](https://github.com/BetterThanTomorrow/calva/issues/2959)
```
```suggestion
- [Use latest version of libs for Jack-in, if no specific version was set by the user](https://github.com/BetterThanTomorrow/calva/issues/2959)

  Previous default versions:
  - `nrepl`: 1.3.1
  - `cider-nrepl`: 0.55.4
  - `cider/piggieback`: 0.6.0
```

PEZ: Since the new versions now only work as fallbacks, I am thinking it is more important to mention the versions replaced.

PEZ:
Thanks. Looks fabulous!

I've suggested an update to the changelog entry.

Only other thing I think we want is that we log the checking to the console, also when things go well. Which versions we find and when global storage is updated. Good when developing on Calva and also for users to check when they report any problems.

We could also log latest versions to Calva says, together with the effective version, and from where the effective versions are decided (like “latest known by Calva”, “user configured override”, "Calva defaults fallback”).

I love the integration tests! ❤️

afucher:
@PEZ added the logs and messages to Calva says
Also, fixed the integration test because it was flaky

PEZ:
Sweet!

PR #2970, opened 2025-12-07
Handle spaces in path when jacking-in

### LINKED ISSUE #2549: [Windows] Jack-in sometimes fails from quoting issues
Sometimes Jack-in fails on Windows, especially for deps.edn projects. It is something with the quoting of the command line, or rather something with the environment where the command line is executed that is not friendly to the quoting Calva does.

The quoting is made for the cmd.exe shell, so it may be that sometimes it is some other shell that gets to execute the the command line. I'm a bit clueless on what the fix is because most of the times I can't reproduce the problem.

There's a workaround, though:

1. Instead of Jack-in, use the command for copying the jack-in command line
2. Open a cmd.exe shell (in an integrated Terminal or outside VS Code, use whatever makes most sense to you)
3. Paste and execute the command line
4. When the nREPL server is started, use the Calva command for connecting to a running nREPL server.

TheRealMintd:
Linked issue is specific to Windows, but I'm sure that the same bug can happen on other platforms as well.

## What has changed?

- Quote paths to the bundled deps.clj.jar

Fixes #2549 

## My Calva PR Checklist

I have:

- [x] Tests


PEZ:
Thanks!

We'll need to test this on non-Linux. You can add a test project in `test-data/projects` that would expose the problem without the extra quoting.

TheRealMintd:
This isn't something that can be tested with a test project. The bug occurs when the path to the `deps.clj.jar` file contains spaces.

Tested using WSL and cloning Calva to a dir with spaces also reproduces this issue on Linux.

PEZ:
I agree the quoting added by this PR needs to happen, but we need to test it a lot in different environments. Quoting, especially on Windows, has proven very, very tricky, with all different shells and settings and all. We risk making something stop working that used to work. I'm at loss for how to conduct the tests, though...

I have another PR with high risk that I am going to merge first. Then, once that is out we can try to get people to use a build of this PR (merged on the new `dev`) and see how people fare. My experience with getting help with tests like this is that people generally are not interested to participate, but worth a try.

Issue #2085, opened 2023-02-16
Slurp forward to wrap a form introduces extra space


wevre:
With cursor right before a form, if I type a delimiter and then slurp forward, there is a space between the opening delimiter and the newly wrapped form.
```
|:b
;; Want to wrap :b, type '['
[|]:b
;; slurp forward
[| :b]
;;^ don't like that space, wish instead it were
[|:b]
```
This outcome is the same regardless of delimiter.

Request that Calva not include this extra preceding space.

PEZ:
Thanks. This has to do with how we format (or not) the forms after Paredit operations. It was an effect of starting to use the formatter in on-type mode. In on-type mode our formatter relaxes the formatting and does not trim spaces.

* https://github.com/BetterThanTomorrow/calva/issues/1924

I don't know what the real solution is here, but I think it makes sense with on-type by default and maybe looking over which Paredit operations we want to either do some other post-formatting on, or if they should be changed in the actual paredit operation. Like with the example, the slurp replaces the bracket with a space, maybe it could delete the bracket instead, making on-type post formatting sufficient...

Something you would want to take a look at, @SillyCoon?

I think it is worth having a look at all paredit operations https://calva.io/paredit/ and see how they behave. It could give us an idea of what we should do about it, generally. Then maybe use TDD to expose the cases we want to change, and fix them.

SillyCoon:
I'd be happy to take a look, thank you for the suggestion @PEZ!

SillyCoon:
Hi @PEZ. After a little investigation, I guess the issue here is that we don't check if the 'slurping' form is empty or not and it just inserts the `whitespace` to split the args like:

`[a|]  b` => `(a b)` is correct but 
`[|] b` => `( b)` behaves the same

I don't really know if should we include the formatting after `slurp forward` or after the `slurp backward`, but in general it could `fix` a lot of bugs without actually fixing them.

PEZ:
Thanks for checking this out, @SillyCoon!

> I don't really know if should we include the formatting after `slurp forward` or after the `slurp backward`, but in general it could `fix` a lot of bugs without actually fixing them.

Iirc we are doing formatting, but we do a relaxed variant, which does not trim whitespace inside brackets. We used to do the trimming too, but that had other effects, #1924 might have some clues about it. I think Slack might hold some discussion that led to #1924. 

[...]

PR #3006, opened 2026-01-30
Fix: slurp backward empty form

### LINKED ISSUE #1440: Paredit interaction with strings
When I do “slurp sexp forward” at:

`"|"somestuff`

where the | is my cursor, I get ” somestuff”.

I would expect the string to not start with a space.
I've tried a few other paredit commands inside of strings and they seem Ok, but I'm to a paredit expert.

According to @PEZ :

> Not really intended, It’s a side effect of that strings are treated as lists and we haven’t made a special case for it. But we can consider it.

jramosg:
## What has changed?

Fixed Slurp Backward command adding unwanted trailing spaces when slurping into empty or whitespace-only forms.

### Empty Forms
**Before:** `foo "|"` would become `"foo |"` with a trailing space
**After:** `foo "|"` becomes `"foo|"` without trailing space

**Before:** `foo (|)` would become `(foo |)` with a trailing space
**After:** `foo (|)` becomes `(foo|)` without trailing space

**Before:** `foo (|   )` (whitespace-only list) would become `(foo |)` with trailing space
**After:** `foo (|   )` becomes `(foo|)` without trailing space

### Forms with Content
Forms with content still get the space as expected: `"a" b` becomes `"a b"` and `(bar) foo` becomes `(bar foo)`.

The fix detects when the target form contains no actual content (only whitespace between open and close brackets, or empty strings) and omits the trailing space in that case. This brings consistency with the forward slurp fix.

Fixes #1440 and  #2085

## My Calva PR Checklist

I have:

- [x] Directed this pull request at the `dev` branch.
- [x] Made sure I have changed the PR base branch, so that it is not `published`.
- [x] Made sure there is an issue registered with a clear problem statement that this PR addresses.
    - [x] Updated the `[Unreleased]` entry in `CHANGELOG.md`.
- [x] Figured if **anything** about the fix warrants tests on Mac/Linux/Windows/Remote/Whatever~ (paredit logic, no platform-specific behavior)
- [x] Added to or updated docs in this branch~ (bug fix, no user-facing doc changes needed)
- [x] Tests
  - [x] Added tests for slurping into empty string, list, whitespace-only list, and vector
  - [x] Added regression tests for non-empty forms (should still add space)
- [x] Formatted all JavaScript and TypeScript code that was changed.
- [x] Confirmed that there are no linter warnings or errors.


PEZ (approved):
Thanks! ❤️

PR #3013, opened 2026-02-01
2981 load file promise completion

### LINKED ISSUE #2981: The 'calva.loadFile' command returns a Promise that is only completed on the first execution.
I am sure this counts as a minor issue as VS Code itself does not care about this for normal operations.

But when called from Joyride (or from another extension, for which I only have a very experimental personal use case for now) with `await`, you get stuck. **Update:** it is probably relevant to mention that this happens when I have all `calva.outputDestinations` set to `"terminal"`.

Reproducing Joyride user script (JavaScript):
```
const vscode = require("vscode");

async function run() {
  const output = vscode.window.createOutputChannel("My Joyride script");
  output.appendLine("Executing calva.loadFile");
  output.show(true);
  await vscode.commands.executeCommand("calva.loadFile");
  output.appendLine("Done executing calva.loadFile");
  output.show(true);

  // Do something else after calva.loadFile is finished.
}

run();
```

When run with an active REPL with a regular Clojure file open, this will output:
```
Executing calva.loadFile
Done executing calva.loadFile
```
initially.
But when run again (and again, etc.), it will only output:
```
Executing calva.loadFile
```
and not get beyond the `await`.

---

This seems to be caused by `loadFileCommand` in 'evaluate.ts' returning a Promise that invokes `output.replWindowAppendPrompt(resolve);` that does not really expect a Promise (certainly does not always call `resolve`, which is called `onAppended` inside `replWindowAppendPrompt`, which communicates another expectation.)

Unfortunately, I am not at all sure how this mismatch should best be taken care of.

jurjanpaul:
… new load-file-test.ts integration test

+ improved integration test run isolation by removing `.vscode-test` at the start of each run.

## What has changed?

- The `appendPrompt` function in repl-window-doc.ts now resolves the given promise (`onAppended`) even when no prompt needs to be appended. 

As alternative solution I considered turning `appendPrompt` into an `async` function itself. I seemed to get quite far with this refactoring that affected more code than I identified initially. Unfortunately, I must have overlooked some consequences, because even if I got the new integration test to pass this way, other integration tests started to fail with timeouts. :-) Now realizing that I understood even less of the expected async/sync behavior and the interaction between those integration tests I decided to put more emphasis on having a better isolated integrationtest first.

I have also considered (and tried) renaming the (now rather misleadingly named) `onAppended` parameter to `onAppendDone` (+ renaming the specific type accordingly), but that also required me to make rather extensive changes to the code base that I could not yet fully oversee and therefore chickened out of. I do still think such a rename might make sense and would not necessarily break things. So, if desirable for this PR, I can certainly give that another go now that I am experiencing more control over the integration tests.

(Another possible direction that I have not tried in the end, might be to wrap the call to `appendPrompt` from `loadFileCommand` in evaluate.ts in another `if` statement over `replWindow.isReplWindowDoc(doc)`, but that would also require some code duplication/refactoring to require `doc` again.)

- I introduced a new integration test `load-file-test.ts` to prove the problem and verify the solution. 

I can imagine there is some overlap with jack-in-and-connect-test.ts, but having a dedicated test file was of course easier to work with. In a way this kind of test may also belong to the sphere of 'end-to-end' tests. So, please let me know if it would be preferable to merge or rearrange things after all.

- Modified src/extension-test/integration/runTests.ts to always start by removing any existing '.vscode-test' directory, because I found during development of my integration test that I was really bitten again and again by the configuration not being returned to its expected original state. By now, I believe that the integration test itself does a good job of resetting the configuration, so that this explicit removal may no longer be strictly necessary, but I still think it is useful to guarantee in this manner that the integration tests always start from a clean configuration state.

Fixes #2981 'The 'calva.loadFile' command returns a Promise that is only completed on the first execution.'

## My Calva PR Checklist

I have:

- [~] Figured if **anything** about the fix warrants tests on Mac/Linux/Windows/Remote/Whatever, and either tested it there if so, or mentioned it in the PR.
- [~] Added to or updated docs in this branch, if appropriate
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Changes to Calva.
 
 ## [Unreleased]
 
+- Fix: [The 'calva.loadFile' command returns a Promise that is only completed on the first execution.](https://github.com/BetterThanTomorrow/calva/issues/2981)
+
 ## [2.0.546] - 2026-01-30
 
 - [Consider form pairs in `:let` bindings when growing the selection and when dragging forms](https://github.com/BetterThanTomorrow/calva/issues/2988)
--- a/src/extension-test/integration/runTests.ts
+++ b/src/extension-test/integration/runTests.ts
@@ -1,4 +1,5 @@
 import * as path from 'path';
+import * as fs from 'fs';
 import { glob } from 'glob';
 
 import { runTests } from '@vscode/test-electron';
@@ -8,6 +9,8 @@ async function main() {
     // The folder containing the Extension Manifest package.json
     // Passed to `--extensionDevelopmentPath`
     const extensionDevelopmentPath = path.resolve(__dirname, ...['..', '..', '..']);
+    const vscodeTestPath = path.resolve(extensionDevelopmentPath, '.vscode-test');
+    fs.rmSync(vscodeTestPath, { recursive: true, force: true });
 
     // The path to the extension test runner script
     // Passed to --extensionTestsPath
--- a/src/extension-test/integration/suite/load-file-test.ts
+++ b/src/extension-test/integration/suite/load-file-test.ts
@@ -80,6 +80,7 @@ suite('Load File Command Test', () => {
       });
       await jackInHarness.waitForNextClient();
       testUtil.log(suite, 'Waiting for jack-in to complete (terminal output mode)...');
+      // Unfortunately, awaiting jack-in does not work with terminal output mode.
       await testUtil.sleep(5000);
     } else {
       await jackInHarness.jackInWithConnectSequence(testFilePath, connectSequence);
--- a/src/extension-test/integration/suite/load-file-test.ts
+++ b/src/extension-test/integration/suite/load-file-test.ts
@@ -1,0 +1,121 @@
+import * as vscode from 'vscode';
+import * as assert from 'assert';
+import * as path from 'path';
+import { before, after, beforeEach, afterEach } from 'mocha';
+import { getDocument } from '../../../doc-mirror';
+import * as jackIn from '../../../nrepl/jack-in';
+import {
+  ReplConnectSequence,
+  ProjectTypes,
+  CljsTypes,
+} from '../../../nrepl/connect-sequence-types';
+import * as outputWindow from '../../../repl-window/repl-window-doc';
+import * as testUtil from './util';
+
+suite('Load File Command Test', () => {
+  const suite = 'Load File Command Test';
+  const jackInHarness = new testUtil.JackInHarness(suite);
+  let originalDestinations: any;
+
+  const testFile = 'test.clj';
+  const testFilePath = path.join(testUtil.testDataDir, testFile);
+  const connectSequence: ReplConnectSequence = {
+    projectType: ProjectTypes['deps.edn'],
+    name: 'deps.edn',
+    cljsType: CljsTypes.none,
+    projectRootPath: [path.join(testUtil.testDataDir, '..')],
+    menuSelections: { cljAliases: [] },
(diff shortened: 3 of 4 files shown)


PEZ:
Thanks for this work! 🙏 ❤️ 

Some of the async in Calva is very messed up, as you have noticed. So if you see ways to clear things up, that is very appreciated. 

The rename to `onAppendDone` makes sense. I choose to merge so as not to let things depend on this, but you can include it in follow-up PRs. 

I think a separate integration test for load-file makes perfect sense. The other layer of e2e tests could possibly be a good fit, but to my experience we have better control with the electron-test based things. (To the point where we should consider retiring the e2e-stuff because it is a bit questionable if it adds any extra protection. But that is for another day.)

> the interaction between those integration tests

This is unfortunate. We shouldn't have tests that depend/interact on each other. Is that what you saw, or do you mean that changes in `appendPrompt` caused other tests to fail? That is also very unfortunate. Appending the prompt to the REPL Window should not be allowed to interfere with loading a file, at least not in an ideal world, which Calva is not, but anyway. If you see some way to disentangle things there, that would be lovely.

jurjanpaul:
Thank you @PEZ for accepting the PR, as well as releasing it so quickly, and for the encouraging words!!

Integration test isolation (or lack thereof) only became a problem because I set the output destination(s) in the new test to ‘terminal’ without (originally) resetting afterwards. Understandably this caused other tests to fail. What confused me more - especially with days between my coding sessions - was the fact that the tests kept failing on new runs after I commented out my new tests, but this was of course the result of the persistent nature of the config. Hence my addition of explicitly deleting the ‘.vscode-test’ directory at the start of each new integration test run. I just clearly underestimated the effort I needed to put in to familiarize myself with the various aspects of the environment in which the tests run. Now that I’ve come this far I hope additional changes will be a bit more straightforward... But yes, I noticed that the async stuff is hard to reason about and not particularly consistent, especially with all the state and side effects going on. But I totally appreciate the history and your continued efforts to make Calva the best VS Code Clojure extension that it can be.

For a renaming/refactoring-only PR, do you prefer to have a separate GitHub issue or would it be sufficient to reference this one?

PEZ:
Calva is a stateful monster. We are often slowly improving that and then quickly making it worse again.

No need for an issue for renaming things internally, at least not now that we have communicated about such needs here. We want a Changelog entry if thee changes touches Calva itself. (Reason being that if we introduce regression the Changelog is often where users can start checking for recent changes related to the issue they see.) Just link the PR.

jurjanpaul:
@PEZ Now that I have the integration tests running properly I managed to finish my original idea of making more of the functions async (specifically `replWindowAppendPrompt` in output.ts and `appendPrompt` in repl-window-doc.ts), only creating a void Promise callback further down the stack. This way there is no need to invoke `onAppend` when nothing needs to be appended. Therefore it does not need the suggested rename from `onAppended` to `onAppendedDone` after all.
Here is the diff: [async-appendPrompt branch](https://github.com/BetterThanTomorrow/calva/compare/published...jurjanpaul:calva:2981-async-appendPrompt) .

Alternatively, I have another branch with the rename, also renaming the type(s) involved: [onAppendDone branch](https://github.com/BetterThanTomorrow/calva/compare/published...jurjanpaul:calva:2981-onAppendDone).

So, I myself much prefer the direction of the first change over the second. I only can imagine that it can be taken a bit further (doing the same for `forceAppendPrompt` seems low hanging fruit; going further would mean a lot more impact obviously).

So, which would you prefer that I turn into a PR?
Thanks!

PEZ:
Agree about the first direction.

PR #3098, opened 2026-02-21
Format after toggle comment

### LINKED ISSUE #3095: Toggle comments structurally does not re-format
When toggling comments using the new default ignore/discard, the code is not formatted. Ideally, paredit commands should result in code that is always formatted (i.e. manual formatting with the `tab` key should seldom do anything).

jramosg:
## What has changed?

- Added `reformatFormAtPosition`, it is called after `toggleIgnoreForm`. Finds the enclosing form and formats.

Fixes #3095

## My Calva PR Checklist

I have:

- [x] Tests

### DIFF
--- a/src/edit.ts
+++ b/src/edit.ts
@@ -178,7 +178,10 @@ async function applyStructuralCommentsToSingleSelectionLines(
           new vscode.Position(lineNum, insertionColumn)
         );
 
-        const wouldBreakWhere = paredit._semiColonWouldBreakStructureWhere(mirrorDoc, insertionOffset);
+        const wouldBreakWhere = paredit._semiColonWouldBreakStructureWhere(
+          mirrorDoc,
+          insertionOffset
+        );
 
         editBuilder.insert(new vscode.Position(lineNum, insertionColumn), ';; ');
 
@@ -496,7 +499,6 @@ async function toggleCommentsThenReformatEnclosingForms(
   await updateLineComments(editor, affectedLineNumbers, shouldUncomment, candidatesMap);
   await reformatEnclosingFormsForLines(editor, affectedLineNumbers);
 }
-
 
 /**
  * Returns true if the cursor is currently positioned within a ;; line comment.
--- a/src/edit.ts
+++ b/src/edit.ts
@@ -4,7 +4,7 @@ import * as docMirror from './doc-mirror/index';
 import { EditableDocument, ModelEdit } from './cursor-doc/model';
 import * as select from './select';
 import * as printer from './printer';
-import { _semiColonWouldBreakStructureWhere } from './cursor-doc/paredit';
+import * as paredit from './cursor-doc/paredit';
 import * as format from './calva-fmt/src/format';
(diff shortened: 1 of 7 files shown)


src/edit.ts
```diff
@@ -420,6 +420,22 @@ async function reformatEnclosingFormsForLines(
...
+  const enclosing = select.getEnclosingFormSelection(document, position);
+  const range = enclosing || select.getFormSelection(document, position, false);
+  if (!range) {
+    await editor.edit(() => undefined, { undoStopBefore: false, undoStopAfter: true });
+    return;
+  }
+  const offsetRange: OffsetRange = [document.offsetAt(range.start), document.offsetAt(range.end)];
+  await reformatRanges(editor, [offsetRange]);
+}
+
```
PEZ: Is there a reason we do not implement `toggleLineCommentCommand` as a paredit command? Then reformatting would be built-in.

jramosg: It's outside paredit because initially was formatting only ;; commented lines and not the form itself. 
We can move it to paredit
But I don't understand that formatting would be built-in. I don't see where that is done for every paredit command (if it is done). Shouldn't I need a custom format fn (that calls formatRanges of course) even if it is a paredit command?

PEZ: It happens in doc-mirror. I asked Copilot to describe it for me:

---
## The Flow: Data → Edit → Format

Paredit commands in [paredit.ts](src/cursor-doc/paredit.ts) are **pure computation** — they produce `ModelEdit[]` arrays (insert/change/delete primitives) plus a `ModelEditOptions` bag that includes the key `skipFormat` flag. They never format anything themselves.

The formatting happens in the **bridge layer** at [index.ts](src/doc-mirror/index.ts), specifically in `DocumentModel.edit()`. The flow:

1. **Paredit computes** structural edits + `{ skipFormat: true/false }`
2. **`DocumentModel.edit()`** pre-computes reformat offsets, then applies text edits to VS Code
3. **If `!skipFormat`**, it calls `postEditReformat()` which uses the Calva formatter (`formatDocIndexesInfo()`) to compute whitespace changes and applies them in a **second grouped edit**
4. Both phases share a **single undo unit** — Ctrl+Z reverts structure + formatting together

### When `skipFormat` varies

Most commands format (`skipFormat: false`), but some opt out:

- **`killLeft`** skips formatting when only killing whitespace (don't re-indent what was explicitly cut)
- **`insertSemicolon`** skips (comments shouldn't trigger reformat)
- **`addRichComment`** uses it conditionally — false for actual insertions, true for cursor-only moves

The design is elegant: paredit stays as a pure function from document state → edit descriptors, and the imperative formatting/VS Code integration lives entirely in the doc-mirror bridge. Data in, data out — the Tao of structural editing.

---

Checks out with my understanding. 😄

jramosg: oh, thanks!!

jramosg:
restarted in a fresh branch https://github.com/BetterThanTomorrow/calva/pull/3110 
I focused only in `#_` toggles because formatting `;;` was becoming a mess, I will give another  try to `;;` but for now it's a good step to have `#_` as paredit command

PR #3104, opened 2026-02-24
Fix code navigation fail after starting repl

### LINKED ISSUE #2804: Code navigation fail after starting repl
We can see where a function is called from by cmd(ctrl)+left-clicking the function name at its definition, and we can also jump to its definition by cmd(ctrl)+left-clicking on the function name where it's called. However, after starting repl, the latter still works while the former doesn't anymore. Exiting repl restores it.

jramosg:
## What has changed?

- Updated the provideClojureDefinition function to check if the resolved URI matches the current document's URI. If they match, it delegates to the lspDefinition function for further processing.
- Updated the provideClojureDefinition and provideDefinition methods to include a cancellation token parameter.

Fixes #2804

## My Calva PR Checklist

I have:

- [x] Tests


src/providers/definition.ts
```diff
@@ -28,7 +31,11 @@ async function provideClojureDefinition(
       if (info.file && info.file.length > 0) {
         const pos = new vscode.Position(info.line - 1, info.column || 0);
         try {
-          return new vscode.Location(vscode.Uri.parse(info.file, true), pos);
+          const uri = vscode.Uri.parse(info.file, true);
+          if (uri.toString() === document.uri.toString()) {
+            return lspDefinition(clientProvider, document, position, token);
```
PEZ: It seems wrong to provide lsp definition in the function that is mapped to providing repl definitions?

If this is a fallback, then that fallback should probably happen in provideDefinitions?

jramosg: you're right, ClojureDefinitionProvider already handles the fallback
```ts
 async provideDefinition(document, position: vscode.Position, token: vscode.CancellationToken) {
    const providers = config.getConfig().definitionProviderPriority;
    for (const provider of providers) {
      const providerFunction = definitionFunctions[provider];
      if (providerFunction) {
...
```

PEZ:
This looks good to me now. Is it ready to merge, from your perspective, @jramosg ?

jramosg:
I've been using this build yesterday afternoon and today morning I didn't see anything strange and code navegation was fixed, so yes!

Issue #2655, opened 2024-10-26
Keep support for evaluating top level form in ignored forms when at top level


PEZ:
Today I was made aware, by @rafd, that **evaluate top-level form** “works” when the ignored form is at the top level. This makes sense, but it would also make sense that this should include the ignore marker in the evaluation, and then nothing (or something similar to nothing) should happen. We have actually intended for the latter, but because reasons, this is quite tricky business and we have failed with the intent.

It's likely that some users has started to rely on the current behaviour (at least one has), so I think we should make the current behaviour explicit and intended in the code. It gets important when  dealing with issues like these:

* https://github.com/BetterThanTomorrow/calva/issues/2387
* https://github.com/BetterThanTomorrow/calva/issues/2164

In `rangeForDefun` when we traverse the forms upwards to find the top level we check for if we are in a `comment` form and treat that as the top level context.
 
https://github.com/BetterThanTomorrow/calva/blob/published/src/cursor-doc/token-cursor.ts#L768

I think this probably is the function to deal with this, since it would allow us to figure out the general handling of ignored forms without risking that evaluating top level ignored forms stop working as intended.

PR #3114, opened 2026-03-01
Current form takes `#_` commented forms with cursor in all positions

### LINKED ISSUE #3115: When we are in a commented form in any cursor position  both the marker and the ignored form must be considered as Current Form
https://github.com/BetterThanTomorrow/calva/pull/3111#issuecomment-3979704874

Update the current form for all these:

```clojure
|#_:a
#|_:a
#_|:a
#_:|a
#_:a|
```

To be:

```clojure
|#_:a|
```

jramosg:
## What has changed?

- Current form includes #_ at all cursor positions
- nsFromCursorDoc no longer treats `#_` discarded ns forms as valid 
    - All three code paths in `nsRangeFromCursorDoc` now check if the   
  form range starts with an ignore token and skip it. This prevents `#_` discarded forms
   from being treated as valid ns declarations.

Fixes #3115

## My Calva PR Checklist

I have:

- [x] Tests

### DIFF
--- a/src/cursor-doc/token-cursor.ts
+++ b/src/cursor-doc/token-cursor.ts
@@ -657,6 +657,11 @@ export class LispTokenCursor extends TokenCursor {
       !this.tokenBeginsMetadata()
     ) {
       afterCurrentFormOffset = this.offsetEnd;
+      // If preceded by #_, include the ignore marker in the range
+      const ignoreCursor = this.clone();
+      if (ignoreCursor.backwardThroughAnyIgnore()) {
+        return [ignoreCursor.offsetStart, afterCurrentFormOffset];
+      }
     }
     // console.log(0, afterCurrentFormOffset);
 
@@ -669,9 +674,16 @@ export class LispTokenCursor extends TokenCursor {
         cursor.getToken().type !== 'reader' &&
         !cursor.tokenBeginsMetadata() &&
         cursor.getPrevToken().type !== 'reader' &&
+        cursor.getPrevToken().type !== 'ignore' &&
         !cursor.prevTokenBeginsMetadata()
       ) {
         if (cursor.backwardSexp() && !cursor.tokenBeginsMetadata()) {
+          // If the cursor is at end-of-line and the form is immediately preceded
+          // by #_ (no whitespace between), include the ignore marker in the range.
+          if (this.getToken().type === 'eol' && cursor.getPrevToken().type === 'ignore') {
+            cursor.previous();
+            return [cursor.offsetStart, offset];
+          }
           afterCurrentFormOffset = offset;
         }
       }
@@ -718,6 +730,12 @@ export class LispTokenCursor extends TokenCursor {
           // only handles 'reader' tokens, not 'ignore' tokens.
           if (tokenTypeAtFormStart === 'ignore') {
             return [formStart, afterCurrentFormOffset];
+          }
+          // When the form at formStart is preceded by #_, include the ignore
+          // marker in the range (e.g. cursor between #_ and the form).
+          const ignoreCursor = this.doc.getTokenCursor(formStart);
+          if (ignoreCursor.backwardThroughAnyIgnore()) {
+            return [ignoreCursor.offsetStart, afterCurrentFormOffset];
           }
         }
       }
--- a/src/cursor-doc/token-cursor.ts
+++ b/src/cursor-doc/token-cursor.ts
@@ -832,7 +832,10 @@ export class LispTokenCursor extends TokenCursor {
           return lastCandidateRange;
         }
       } else {
-        lastCandidateRange = cursor.rangeForCurrentForm(cursor.offsetStart);
+        const end = cursor.offsetStart;
+        const sc = cursor.clone();
+        sc.backwardSexp(true, true);
+        lastCandidateRange = [sc.offsetStart, end];
       }
     }
     return lastCandidateRange;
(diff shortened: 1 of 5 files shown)


jramosg:
this test fails
```ts

    // TODO: Figure what to do with ignored forms
    //       For now, this is what they do (nothing)
    it('Finds ns in top level ignored form', function () {
      expect(nsFormUtil.nsFromCursorDoc(docFromTextNotation('#_ (ns a-b.c-d)|'))).toStrictEqual([
        'a-b.c-d',
        '(ns a-b.c-d)',
      ]);
    });
```

it's receiving `["a-b.c-d", "#_ (ns a-b.c-d)"]`
I think ns for `#_ (ns a-b.c-d)` should be `null` what do you think @PEZ ?

PEZ:
Yeah, `null` is fine. Is that what you get without fixing anything?

jramosg:
I had that "error" when I did the change about treating #_ form as current form at all cursor positions.
Now I changed the `nsRangeFromCursorDoc` so it does not tre`at #_(ns x) ` as valid ns

PEZ:
Sweet. I've tested it and It feels pretty natural. There's some inconsistency now between selecting the current form and moving forward and backward a form at a time, as well as with dragging pairs. But it is not like we did not have inconsistencies before. I think we are in a much better place now. Thanks! 🙏

src/extension-test/unit/cursor-doc/token-cursor-test.ts
```diff
@@ -773,6 +773,36 @@ describe('Token Cursor', () => {
...
+      const expected = textAndSelection(b)[1];
+      for (const notation of ['|#_(foo bar)', '#|_(foo bar)', '#_|(foo bar)']) {
+        const a = docFromTextNotation(notation);
+        const cursor: LispTokenCursor = a.getTokenCursor(a.selections[0].anchor);
+        expect(cursor.rangeForCurrentForm(a.selections[0].anchor)).toEqual(expected);
+      }
+    });
+    it('selects only the keyword when cursor is in whitespace after #_:a in #_:a :b', () => {
+      const a = docFromTextNotation('#_:a| :b');
+      const b = docFromTextNotation('#_|:a| :b');
```
jramosg: what do you think about this @PEZ ?
if I change it to take `|#_:a|` this test should be changed too, and I think we don't want that
```ts
   // https://github.com/BetterThanTomorrow/calva/issues/2655
      it('Does not include ignore marker', () => {
        const a = docFromTextNotation('aaa (comment #_ [bbb ccc|]  ddd)');
        const b = docFromTextNotation('aaa (comment #_ |[bbb ccc]|  ddd)');
        const cursor: LispTokenCursor = a.getTokenCursor(a.selections[0].active);
        expect(cursor.rangeForDefun(a.selections[0].active)).toEqual(textAndSelection(b)[1]);
      });
```

But maybe this is inconsistent with 
`#_:a|` -> `|#_:a|` ? And it should be `#_|:a|` ?

PEZ: I think we should include the ignore marker in both cases. The whitespace shouldn't make a difference.

jramosg: yes, whitespace does not make difference, I've whitespaces tested
```ts
   it('selects ignore form including #_ when cursor is at end of #_     :a (spaces between)', () => {
      const a = docFromTextNotation('#_     :a|');
      const b = docFromTextNotation('|#_     :a|');
      const cursor: LispTokenCursor = a.getTokenCursor(a.selections[0].anchor);
      expect(cursor.rangeForCurrentForm(a.selections[0].anchor)).toEqual(textAndSelection(b)[1]);
    });
```

It it when we have a form after.  I thought it was intentional after reading  https://github.com/BetterThanTomorrow/calva/issues/2655
So I will change it 👌

PEZ: Good catch! It's important that we keep supporting #2655. This is what I referred to when I said I couldn't quite grasp the implications of this type of change. Let me think a bit... brb!

jramosg: aah okey, so I will try to support that in `rangeForDefun` while selections works as `#_:a| :b -> |#_:a| :b`

PEZ: I think we will need some sort of special case. Similar, in spirit, to how we have a special idea of top-level when inside Rich comments.

Let's see where you arrive with your idea of handling it in `rangeForDefun`. Sounds like the right approach to me.

src/cursor-doc/token-cursor.ts
```diff
@@ -815,7 +832,10 @@ export class LispTokenCursor extends TokenCursor {
           return lastCandidateRange;
         }
       } else {
-        lastCandidateRange = cursor.rangeForCurrentForm(cursor.offsetStart);
+        const end = cursor.offsetStart;
```
jramosg, line 835: @PEZ this is the `rangeForDefun` case, it cn be managed  with `backwardSexp` and skipping comments
I also added a test to check when there is not whitespace between the  #_ and the form

jramosg:
@PEZ shall I merge this? Looks good to me now

PEZ:
I lost track a bit. But as long as #2655 is safe, I am good with merging.

PR #3118, opened 2026-03-02
[ISSUE 3117] allow custom comment forms

### LINKED ISSUE #3117: allow other forms to create a new toplevel context like `comment` does
There currently is a feature which allows treating the inside of `comment` forms like their own toplevels, e.g. in the context of `Evaluate Top Level Form`. It would be nice if this behavior was configurable for other forms as well.

EvenMoreIrrelevance:
## What has changed?

- added the `customCommentForms` config key, with the value being a string array
- added `isCommentFormHead` to `utilities`, which compares with `comment` or else checks if the symbol or the resolution (per `aliasMap`) thereof is in `customCommentForms`
- replaced all checks that compared a form head with `comment` to `isCommentFormHead` calls

Fixes #3117

## My Calva PR Checklist

I have:

- [ ] Tests

### DIFF
--- a/src/utilities.ts
+++ b/src/utilities.ts
@@ -12,6 +12,8 @@ import * as url from 'url';
 import { isUndefined } from 'lodash';
 import * as fiddleFiles from './fiddle-files';
 import * as output from './results-output/output';
+import { getConfig } from './config';
+import { resolveAliasedSymbol } from './cursor-doc/paredit-config';
 
 const specialWords = ['-', '+', '/', '*']; //TODO: Add more here
 const syntaxQuoteSymbol = '`';
@@ -645,6 +647,19 @@ function pathExists(path: string): boolean {
   return fs.existsSync(path);
 }
 
+function isCommentFormHead(symbol: string): boolean {
+  if (symbol === 'comment') {
+    return true;
+  }
+
+  const { customCommentForms, aliasMap } = getConfig();
+  return (
+    symbol === 'comment' ||
+    customCommentForms.includes(symbol) ||
+    customCommentForms.includes(resolveAliasedSymbol(symbol, aliasMap))
+  );
+}
+
 export function lastLineIsEmpty(
   doc: vscode.TextDocument = vscode.window.activeTextEditor.document
 ): boolean {
@@ -654,6 +669,7 @@ export function lastLineIsEmpty(
 }
 
 export {
+  isCommentFormHead,
   distinct,
   getWordAtPosition,
   tryToGetDocument,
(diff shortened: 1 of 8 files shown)


PEZ:
Thanks for contributing! 🙏 ❤️ 

Looks good, generally. There are some things missing still, though:

* Configuration needs to be specified in the extension manifest
* We need to mention this in the documentation
* Both unit and integration tests are failing
  * For the unit tests, the changes here bring in vscode into a module that is unit tested, which won't work
  * For integration tests I don't know what's up. One of the fails looked like it couldn't have to do with these changes, but still. Run integration tests locally and see how you fare.
* Not sure, but there may be that we want unit tests that should cover this change. WDYT, @jramosg?

src/utilities.ts
```diff
@@ -645,6 +647,19 @@ function pathExists(path: string): boolean {
...
 }
 
+function isCommentFormHead(symbol: string): boolean {
+  if (symbol === 'comment') {
+    return true;
+  }
+
+  const { customCommentForms, aliasMap } = getConfig();
+  return (
+    symbol === 'comment' ||
```
jramosg, line 657: you already have this check before

jramosg:
yes, I think `isCommentFormHead` should be unit tested. It could receive an optional config arg and as a fallback do `getConfig`. The same for the functions that have been modified and already have tests, like `atTopLevel`. They also could receive an optional config map, which is then passed to `isCommentFormHead`. This way we can add unit tests

This two tests are failing in integration tests


```txt
1) serialization
       input equals output:
     TypeError: Cannot read properties of undefined (reading 'indexOf')
        at resolveAliasedSymbol (out/cursor-doc/paredit-config.js:89:31)
        at isCommentFormHead (out/utilities.js:581:79)
        at /home/circleci/calva/out/NotebookProvider.js:65:47
        at Array.flatMap (<anonymous>)
        at parseClojure (out/NotebookProvider.js:61:30)
        at NotebookProvider.deserializeNotebook (out/NotebookProvider.js:29:29)
        at /home/circleci/calva/out/extension-test/integration/suite/serialization-test.js:52:43
        at Generator.next (<anonymous>)
        at fulfilled (out/extension-test/integration/suite/serialization-test.js:5:58)

  2) Format suite
       should remove spaces and commas from an empty list:

      AssertionError [ERR_ASSERTION]: '(•|•,)' == '(|)'
      + expected - actual

      -(•|•,)
      +(|)
      
        at /home/circleci/calva/out/extension-test/integration/suite/format-test.js:90:16
        at Generator.next (<anonymous>)
        at fulfilled (out/extension-test/integration/suite/format-test.js:5:58)
```

jramosg:
Unit tests are failing because `getConfig` and therefore `import * as vscode from 'vscode';` is being called from some unit tests. So this is another reason to put config as an arg in `isCommentFormHead`.

PR #3124, opened 2026-03-12
Fix nightly clojure-lsp downloads to resolve actual version and avoid redundant re-downloads

### LINKED ISSUE #3125: Nightly clojure-lsp downloads re-download on every startup even when already up to date
When `calva.clojureLspVersion` is set to `"nightly"`, should the downloader resolve the actual nightly release tag and skip the download if the user already has it, rather than re-downloading the full ~41 MB binary on every editor startup, silently falling back to an older version on failure without surfacing the error reason, and writing the literal string `"nightly"` to the version file so there's no way to tell which build is actually installed?

marco-carvalho:
## What has changed?

When `calva.clojureLspVersion` is set to `"nightly"`, the downloader now resolves the actual latest nightly tag from the [clojure-lsp-dev-builds](https://github.com/clojure-lsp/clojure-lsp-dev-builds) GitHub API before downloading. This fixes three issues:

- **Redundant downloads on every startup**: Previously, the literal string `"nightly"` was written to the version file, so the `downloadVersion === 'nightly'` condition was always true and triggered a full ~41 MB download on every editor startup — even if the user already had the latest nightly. Now the resolved tag (e.g. `2026.03.12-18.21.43-nightly`) is stored in the version file, and the download is skipped when the version matches.

- **Silent download failures**: When a nightly download failed, the error was swallowed and the user was silently rolled back to a previous binary with only a generic warning toast. The error reason (e.g. "Download timed out") is now included in the warning message and logged via `console.error`.

- **No way to identify installed nightly version**: The version file previously contained the literal `"nightly"` string, making it impossible to tell which build was actually installed. It now contains the actual release tag.

### Changes

- **`src/lsp/client/downloader.ts`**:
  - Added `getLatestNightlyVersion()` — queries the GitHub `/releases/latest` endpoint for `clojure-lsp-dev-builds` to resolve the actual nightly tag.
  - `ensureServerDownloaded` now resolves `"nightly"` to the real tag before comparing versions.
  - `downloadClojureLsp` detects nightly versions via `.endsWith('-nightly')` and routes to the correct GitHub repo.
  - Removed the `downloadVersion === 'nightly'` always-download condition — the standard version comparison handles it.
  - Warning message now includes the specific error reason on download failure.

- **`src/lsp/client/downloader-utils.ts`**:
  - `downloadWithBackupRecovery` now returns the `error` message in its result so callers can surface it.
  - Changed `console.log` to `console.error` for download failure logging.

Fixes #3125

## My Calva PR Checklist

I have:

- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Changes to Calva.
 
 ## [Unreleased]
 
+- Fix: [Nightly clojure-lsp downloads re-download on every startup even when already up to date](https://github.com/BetterThanTomorrow/calva/issues/3125)
+
 ## [2.0.563] - 2026-02-26
 
 - Fix: [REPL menu defaults to "Start your project" even when REPL is already running](https://github.com/BetterThanTomorrow/calva/issues/3106)
--- a/package.json
+++ b/package.json
@@ -496,6 +496,7 @@
           "calva.clojureLspVersion": {
             "type": "string",
             "default": "latest",
+            "examples": ["latest", "nightly"],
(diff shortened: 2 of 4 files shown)


PEZ:
Ha, so that explains some weirdness I have seen. Thanks! 🙏

PR #3128, opened 2026-03-14
docs: add macOS Local Network access quirk

### LINKED ISSUE #2801: VSCode requires Local Network access on MacOS
After having spent way too much time trying to get Calva working again after my upgrade to MacOS 15.4 (from 15.3) I found out that I had been to restrictive with the permissions for VSCode. If it's not granted "Local Network" access, Calva fails to connect to any REPL, without any clue as to why.

It feels like this could be a pretty common gotcha. Would it make sense to add this to the [Quirks section](https://calva.io/quirks/)? (I'm aware this is no fault of Calva's, but I didn't find any other place that made more sense.)

kovan:
## Summary
Adds a note to the Quirks page about macOS 15+ requiring Local Network permission for VS Code to connect to a REPL. Without it, Calva fails silently with no clear error message.

Fixes #2801

### DIFF
--- a/docs/site/quirks.md
+++ b/docs/site/quirks.md
@@ -21,7 +21,13 @@ See [Using with Parinfer](parinfer.md)
 
 See [Using Calva with the VIM Extension](vim.md).
 
-## ”Command not found” Errors on Jack-in
+## macOS: VS Code Needs Local Network Access
+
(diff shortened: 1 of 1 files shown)


PEZ:
🙏

PR #3130, opened 2026-03-14
docs: note JVM 20+ interrupt evaluation limitation

### LINKED ISSUE #2324: Inform about that Interrupt Evaluation does not awlays work on JVM 20+
With JVM 20, `Thread.stop()` was removed, making interrupting evaluations sometimes fail, because that's what nREPL falls back to, if `Thread.interrupt()` fails. See:

* https://github.com/nrepl/nrepl/issues/296

Calva could probably detect this failure somehow, and inform the user about it. Possibly linking to the nREPL issue, to get more people's brains on the problem.

kovan:
## Summary
Adds a note to the Evaluation docs about JVM 20+ removing `Thread.stop()`, which means nREPL's interrupt fallback no longer works. Documents the workaround:
- Use nREPL 1.3.0+ (configurable via `calva.jackInDependencyVersions`)
- Add `-Djdk.attach.allowAttachSelf` to JVM options

Fixes #2324


PEZ:
Thanks!

PR #3134, opened 2026-03-14
docs: add cljsType to REBL connect sequence examples

### LINKED ISSUE #1234: User Guide: Calva + REBL + deps.edn
I've only just started with Clojure so this may just be a case of my ignorance, but I had a problem following the instructions for setting up Calva + REBL + deps.edn. The user guide says add the following to `settings.json`

```
{
        {
            "name": "Rebl Connect",
            "projectType": "deps.edn",
            "menuSelections": {
                "cljAliases": [
                    "rebl",
                    "rebl-12",
                    "nrebl"
                ]
            }
        }
    ]
}
```
But this didn't work because it's missing an entry for `cljsType`. (I fixed this by adding `"cljsType": "none"`)

Should the snippet in the user guide include a `cljsType` entry or is it deliberately missing?

kovan:
## Summary
Adds `"cljsType": "none"` to both the deps.edn and Leiningen REBL connect sequence examples. Without this, Calva prompts for a ClojureScript REPL type during connection, which is confusing since REBL is Clojure-only.

Fixes #1234

### DIFF
--- a/docs/site/rebl.md
+++ b/docs/site/rebl.md
@@ -46,6 +46,7 @@ Create a Calva custom connect sequence for your VSCode editor. (Read [Custom REP
         {
             "name": "Rebl Connect",
             "projectType": "deps.edn",
+            "cljsType": "none",
             "menuSelections": {
                 "cljAliases": [
                     "rebl",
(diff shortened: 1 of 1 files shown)


PEZ:
Thanks!

I think this may be a regression. Even if it is not, I think cljsType missing from the sequence should mean `none`.

PR #3135, opened 2026-03-14
Fix clojure-lsp setting descriptions: restart → window reload

### LINKED ISSUE #2557: clojureLspPath only read at opening of the window
The description for the calva.clojureLspPath setting says:

> Will take effect after a restart of clojure-lsp. 

But this is not true. It takes effect after a reload of the VS Code window.

kovan:
## Summary
Both `calva.clojureLspVersion` and `calva.clojureLspPath` setting descriptions said they take effect after "a restart of clojure-lsp" but they actually require a reload of the VS Code window.

Fixes #2557

### DIFF
--- a/package.json
+++ b/package.json
@@ -500,13 +500,13 @@
               "latest",
               "nightly"
             ],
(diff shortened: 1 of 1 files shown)


PEZ:
Thanks!

I changed the merge base to `dev`, because this goes into the distributed Calva artifact.

PEZ:
I didn't know how to resolve the conflict (the web ui for this is a bit lacking, you could say) so that's why this stays unmerged, @kovan. If you fix the conflict, I'll merge.

kovan:
Rebased on dev - conflict resolved. The upstream had reformatted the examples array to multiline; I kept that formatting with our text fix.

PR #3140, opened 2026-03-14
Fix "Run current test" matching tests with similar names

### LINKED ISSUE #2168: Running a single test runs all tests containing the single test's name
Mentioned in the Calva channel here: https://clojurians.slack.com/archives/CBE668G4R/p1682118654199809

If I have the following tests defined and I run the "Run current test..." command with my cursor on `a-test`, all three tests run.

```clojure
(deftest a-test
  (println "running a-test")
  (is (= 1 1)))

(deftest b-a-test
  (println "running b-a-test")
  (is (= 1 1)))

(deftest a-test-b
  (println "running a-test-b")
  (is (= 1 1)))
```

We need to pass a regex [here](https://github.com/BetterThanTomorrow/calva/blob/13cfa8dc16db1760db1837990c6c0aa536be55f7/src/nrepl/index.ts#L619) that matches the exact test name, I think. I tried doing this for a bit but couldn't get it to work as desired. `search` is a string, and I'm not sure exactly what is expected there by cider-nrepl/orchard. The docs for `var-query` are [here](https://cljdoc.org/d/cider/orchard/0.11.0/api/orchard.query?q=var-query#vars). (The cider-nrepl docs point there.)

kovan:
## What has changed?

"Run current test" was running all tests whose names contained the target name as a substring. For example, running `a-test` would also run `b-a-test` and `a-test-b`.

The search regex used `\b` (word boundaries) to match test names, but hyphens are word boundary characters in regex. So `\ba-test\b` matches `a-test` inside `b-a-test` because `-` is a boundary.

Replaced `\b...\b` with `/...$` — anchoring on the namespace separator (`/`) and end-of-string (`$`). This works because cider-nrepl's orchard uses `re-find` against namespace-qualified var names like `my.ns/a-test`. The namespace is already filtered by `ns-query`, so we only need to match the name portion exactly.

Previous attempts to use `^...$` failed because `escapeStringRegexp` escapes `^` and `$`. The fix puts the anchors outside the escape call.

Extracted the pattern into a testable `testNameSearchPattern()` function with 5 unit tests.

Fixes #2168

## My Calva PR Checklist

I have:

- ~[ ] Added to or updated docs in this branch, if appropriate~ (no docs changes needed)
- [x] Tests


PEZ:
Thanks again! ❤️

PR #3210, opened 2026-05-07
Reuse a local clojure-lsp binary when the latest version cannot be downloaded

### LINKED ISSUE #3211: clojure-lsp fails to start when the latest version cannot be downloaded, even when older copies are already on the user's machine
When a Clojure project is opened right after a Calva extension update and the request to fetch the latest clojure-lsp version cannot complete (slow network, GitHub outage, corporate proxy, redirect parsing returning empty, etc.), `ensureServerDownloaded` throws and the user sees the "Failed to download clojure-lsp server. Could not fetch latest version for clojure-lsp" error toast immediately on project open. The Calva extension itself activates (the throw is caught in `provisionClient`), but all clojure-lsp features (rename, find-references, code actions, etc.) are unavailable for the session. Previously downloaded clojure-lsp binaries often already exist on the same computer (either inside Calva's own backup folder or inside older sibling Calva install folders), so the expected behavior is to reuse a working local copy as a fallback and surface a non-blocking notice instead of leaving the LSP unavailable. Reproduction: trigger the download failure (offline, blocked network, or a flaky connection at the moment of startup) on a freshly upgraded Calva install with no main binary in the new extension folder; the toast appears and clojure-lsp features stay off even when an older Calva install on the same machine still has a usable binary.

marco-carvalho:
## What has changed?

When `clojure-lsp` cannot be downloaded because the GitHub `releases/latest` redirect cannot be resolved (slow network, GitHub outage, corporate proxy, redirect parsing returning empty, etc.) AND the main binary is missing in the current extension folder (e.g. right after Cursor or VS Code creates a fresh `betterthantomorrow.calva-X.Y.Z-universal/` folder during an extension upgrade), `ensureServerDownloaded` previously threw `Could not fetch latest version for clojure-lsp`. The throw is caught in `provisionClient`, so the Calva extension itself still activates, but the user sees the "Failed to download clojure-lsp server" error toast immediately on project open and all clojure-lsp features (rename, find-references, code actions, etc.) are unavailable for the session. It now tries two local fallbacks before throwing, so a usable binary that already exists somewhere on the user's machine is reused.

- **Local backup folder fallback**: First, attempt to restore the binary from the extension folder's `backup/clojure-lsp` file (or `backup/clojure-lsp.exe` / `backup/clojure-lsp-standalone.jar` per platform) if present. Covers the case where a previous in-flight download left a recovered backup behind.
- **Sibling Calva install fallback**: If no backup is found, scan sibling extension folders matching the same publisher.name prefix derived from the current folder name (e.g. `betterthantomorrow.calva-`), pick the highest recorded clojure-lsp version using lexicographic comparison (works for clojure-lsp's date-based tags such as `2026.05.05-12.58.58-nightly`), and copy the binary plus the `clojure-lsp-version` file into the current extension folder. The binary is `chmod 0o775` on Unix-like platforms.
- **User feedback**: When either fallback succeeds, a non-blocking warning toast names the reused version and clojure-lsp comes up using the reused binary so the user keeps all LSP features. The original throw is preserved for the truly hopeless case where neither fallback finds anything.

### Changes

- **`src/lsp/client/downloader.ts`**:
  - In `ensureServerDownloaded`, the `(downloadVersion === '' && !exists)` branch now tries `restoreFromBackup` and then `findHighestLocalClojureLsp` + `adoptLocalClojureLsp` before throwing.
  - Surface a non-blocking warning toast naming the reused version on either fallback.

- **`src/lsp/client/downloader-utils.ts`**:
  - Added `restoreFromBackup(filePath)` that recovers a binary from the sibling `backup/` directory next to `filePath` when present.
  - Added `findHighestLocalClojureLsp(currentExtensionPath, binaryName, versionFileName)` that scans sibling extension folders by publisher.name prefix and returns the highest-version candidate, ignoring the current folder, unrelated extensions, and folders without a binary. Siblings without a `clojure-lsp-version` file are kept as low-priority candidates (their empty version string sorts to the bottom).
  - Added `adoptLocalClojureLsp(candidate, targetBinaryPath, targetVersionFilePath)` that copies the chosen binary into the target path, chmods it on Unix-like platforms, and writes the matching version file.

- **`src/extension-test/unit/lsp/downloader-test.ts`**:
  - Added unit tests covering: empty backup directory, restored backup, no sibling folders, siblings without binary, current folder skipped, unrelated extensions ignored, the real reporter scenario (5 sibling folders including one with no binary, picking `betterthantomorrow.calva-2.0.582-universal/clojure-lsp` at `2026.05.05-12.58.58-nightly`), siblings without a version file, and the copy-and-write behavior of `adoptLocalClojureLsp`. Full unit suite (1585 tests) passes.

Fixes #3211

## My Calva PR Checklist

I have:

- ~Added to or updated docs in this branch, if appropriate~
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Changes to Calva.
 
 ## [Unreleased]
 
+- Fix: [Keep Calva starting reliably when the latest clojure-lsp cannot be downloaded, by falling back to a previously downloaded binary in the local backup folder or in a sibling Calva install folder](https://github.com/BetterThanTomorrow/calva/issues/3211)
+
 ## [2.0.584] - 2026-05-07
 
 - [enable new clojure-lsp code actions: inline function, if<->cond, extract selection to function, move to :let](https://github.com/BetterThanTomorrow/calva/issues/3204)
--- a/src/extension-test/unit/lsp/downloader-test.ts
+++ b/src/extension-test/unit/lsp/downloader-test.ts
@@ -49,4 +49,224 @@ describe('downloader', () => {
     expectLib.expect(fs.existsSync(binaryPath)).toBe(true);
     expectLib.expect(fs.readFileSync(binaryPath, 'utf8')).toBe(binaryContent);
   });
+
+  describe('restoreFromBackup', () => {
+    it('returns false when the backup directory exists but is empty', async () => {
+      fs.rmSync(binaryPath);
+      fs.mkdirSync(path.join(tmpDir, 'backup'), { recursive: true });
+
+      const restored = await downloaderUtils.restoreFromBackup(binaryPath);
+
+      expectLib.expect(restored).toBe(false);
+      expectLib.expect(fs.existsSync(binaryPath)).toBe(false);
+    });
+
+    it('restores backup binary to original path when main binary is missing', async () => {
+      const backupDir = path.join(tmpDir, 'backup');
+      fs.mkdirSync(backupDir, { recursive: true });
+      const backupPath = path.join(backupDir, 'clojure-lsp');
+      fs.writeFileSync(backupPath, 'backup-version');
+      fs.rmSync(binaryPath);
+
+      const restored = await downloaderUtils.restoreFromBackup(binaryPath);
+
+      expectLib.expect(restored).toBe(true);
+      expectLib.expect(fs.existsSync(binaryPath)).toBe(true);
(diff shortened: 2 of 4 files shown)


PEZ:
Ah, I missed that you had provided a PR. Sorry about that! I have also fixed this, but I took a slightly different path:

* #3213

I think separating the clojure-lsp version from the Calva version makes the most sense so will go with my PR, I think. Let me know what you think, if I am missing some nuance here.

